PR #47, opened 2014-09-17
Converters

dietzc:
### REVIEW NEEDED ###

@dscho I tried to implement Normalizers and Converters in a clean way, especially such that we can optimize each and every op. Can you please review, comment, etc on the design. I'm aware of the fact that the commit messages aren't perfect, but I really wanted to see progress while I'm here and we really have to discuss design issues. (I will clean up the history later of course :-))

You can start reviewing the design by following the examples provided in the ConvertIterableTest I guess. 

Anyway: We also need to figure out a way to auto-generated e.g. fast variants of GenericScaleIterableRT (ArrayImg/PlanarImg etc, for all the type combinations). We will see this pattern occur more frequently: Given one or two ArrayImgs of same or different type apply a pixel-wise function  on them. Maybe we can have a more generic way for creating such classes? any idea?

to make this clear: this is not intended to be merged. I will redo the commit history after we agreed if the design is OK. I'm really nitpicking on the design here because I have no idea if this is good or not. Please simply start from the provided test-case and have a look.


dietzc:
needs more polishing!

PR #48, opened 2014-09-17
Port kernels

bnorthan:
I've been working on porting kernel generation.   At this point still looking for feedback on the design.  

So far I've ported the Gaussian kernel and the Log kernel from pre-existing sources (knime and trackmate).  Both of these follow the same pattern.  

a) the number of dimensions is defined.
b) the radius (or sigma) is defined as a double 
c) the size of each dimensions in pixels is determined by the radius

I've tried for a class hierchy that will handle these implementations but is flexible as to future variations.


bnorthan:
@dietzc I added the methods you suggested.  Anything else??

dietzc:
not from my side. @ctrueden.

I think with merging these changes (especially the createImg stuff), the example in the README.md will work again. @bnorthan did you check that?

bnorthan:
@dietzc, @ctrueden 

The script did not work even with my changes because I only added 'AbstractCreateDefaultImg'.  If I add a new create class that derives from 'AbstractCreateDefaultImg' it works. 

I called it DefaultCreateDefaultImg.  I can change the name if you have a suggestion.  It uses FloatType and PlanarFactory for defaults if nothing is specified.   I can change this too.  Any opinion on what should the defaults be?? 

This class is here.  

https://github.com/imagej/imagej-ops/blob/port_kernels/src/main/java/net/imagej/ops/create/DefaultCreateDefaultImg.java

dietzc:
Thanks

@bnorthan I would (always) prefer to use `ArrayImgFactory` over `PlanaImgFactory`. `PlanarImg`s are just helpers for easier integration with ImageJ1 but is slower in the general case.

one thing which I just discovered: https://github.com/imagej/imagej-ops/blob/port_kernels/src/main/java/net/imagej/ops/create/AbstractCreateImg.java be unifided with `AbstractCreateDefaultImg` right? Because `AbstractCreateDefaultImg` doesn't implement any interface. Then you don't get in the naming hell, because `DefaultCreateImg` would then use `FloatType` and `ArrayImgFactory`. Do I miss something here?

@ctrueden any suggestions?

dscho:
@dietzc you will have to consider the case where `width x height x ? > 2^31`. (Hint: this is something SCIFIO had to implement itself because ImgLib2 does not offer that convenience functionality.)

bnorthan:
@dietzc good idea, I merged the classes as you suggested. 

I did not use ArrayImgFactory as the default though.  There could be problems with very large images.  Of course planar will have problems to if the images are very large and you would need a cell image.  So it seems that eventually something is needed like "SmartImgFactory" that creates the right type based on size and available memory... but I don't see anything like that currently.

ctrueden:
My suggestion for default would be to use a heuristic. If the requested image size is small enough, use `ArrayImg`; if not, use a dynamic cell image like SCIFIO does. (This would be a great thing to work on with @hinerm at the upcoming hackathon!)

ctrueden:
We had to switch the `DefaultCreateImg` type from float to double, but it is finally merged!

dietzc:
:+1:

Issue #67, opened 2014-09-23
Add descriptors/features framework


ctrueden:
This includes Haralick, Zernike, FirstOrderStatistics, Moments, Shape...

Current work can be seen on the increasingly ill-named [feature-service branch](https://github.com/imagej/imagej-ops/compare/feature-service).

dietzc:
we may not forget to offer ops which respect the calibration of the image. In 2D these are especially the geometric features (area for example). In 3D, these are geometric and also (potentially) intensity based features. The calibration of an image can be passed as double[] array. Basically there are two ways to integrate respect calibration in a feature: 
* we always provide a calibration as an optional input with default values `1.0`. 
* we offer two different implementations of the feature, one with calibration one without.

dietzc:
TODO after reviewing with @ctrueden 

* Renaming Def to Default
* get rid of <? extends RealType<?>> and RealType<?> completely. Introduce generic types if needed.
* Feature<T>?
** Can have some helper methods
* DefaultFeatureService make inner classes private!
* Do I really have to restrict myself on features or can have a "OpCompileService"?!
* OpInfo should be in net.imagej.ops (issue #57). OpArgs ().

dietzc:
see https://github.com/imagej/imagej-ops/tree/outputop-service

ctrueden:
@angrauma A little bird told me that Zernike 3D is also working! Anything else that can be checked off the list above?

dietzc:
@ctrueden haralick 3D, zernike 3D not yet :-)

ctrueden:
@dietzc Oh yeah, right! And it is checked, so I am happy.

LeeKamentsky:
3D zernike? How are you planning to generalize to multiple dimensions? I found this paper here:
http://cg.cs.uni-bonn.de/aigaion2root/attachments/novotni-2003-3d.pdf
which creates a 3D Zernike-like basis set on top of spherical harmonics (think of electron orbitals if you want to get an idea).
So inquiring minds want to know *exactly* what you guys might have planned.

angrauma:
@LeeKamentsky Yes, this is exactly the paper I would like to be oriented towards when implementing 3D Zernike moments. I also want to take a closer look at this paper: http://www.emis.de/journals/HOA/MPE/Volume2012/353406.pdf. 
[...]

PR #74, opened 2014-10-06
FeatureService and Implementations

dietzc:
#### THIS PR IS FOR REVIEW ####

related issue: https://github.com/imagej/imagej-ops/issues/67

@ctrueden @dscho @hinerm @DanielSeebacher @angrauma @tinevez 

Ideas behind FeatureService:
* Provide Service to 'compile' a set of features
* Features are numerical descriptors for e.g. images or ROIs or ... (e.g. mean intensity, haralick texture, shape, ...)

Compiling means:
* Resolve dependencies between features automatically (i.e. Mean requires Sum and Area)
* Calculate Features in same FeatureSet only once (i.e. if FeatureSet contains Sum and Mean, only calculate the Sum once and reuse it).

What to review
* Interface design (features vs. regular ops). For example FirstOrderStatOps vs. FirstOrderStatFeatures
* Visible vs. Invisible operations. Visible features are features which are exposed to the user, invisible ops are dependencies which can't be automatically resolved. Lots of prior knowledge required by the developer. And naming somehow not the best?
* What ever you like?

Good starting point: 
* As usual, the tests. FeatureSetTest (which is not really a test yet, more a use-case) or FirstOrderFeaturesTest

Open Points:
* Smarter & nicer implementation of DefaultFeatureService: Currently it is a hack, but I think we can do way more now, as we have the tests ready and the interfaces more or less defined.
* In the end it's an optimization problem, which implementations of Features one should use in combination. But this can be done later. For now the simple heuristic in DefaultFeatureService is fine I think. We just need to make sure, that in the future we still can add more complex FeatureServices (i.e. smarter ones).
* JoinedFeatureSet (which combines several predefined feature-sets to one big one (easy I think)
* Multi-Threading?
* Discussion: How  to add "PixelFeatures" to this framework (for pixel-classification). Should we at all?

Next Steps:
* @DanielSeebacher will implement geometric (size, circularity, convexity, perimeter etc pp), shape (fourier descriptors), tamura (texture), haralick 3d, zernike moments, hue moments etc. features and FeatureSets.
* @DanielSeebacher will create templates for testing of 'FeatureSet's

 Comments? Suggestions? Wishes?


dietzc:
see https://github.com/imagej/imagej-ops/tree/outputop-service

ctrueden:
FTR, this work moved to #110.

PR #81, opened 2014-11-14
Polygon ops

DanielSeebacher:
The default java.awt.polygon only stores points in an int array, but it is possible to get coordinates which are denoted with double values. Therefore I created a simple polygon class. Currently it only consists of a list to store RealPoints but it can be extended later.

I added the Moore Contours algorithm which can extract polygons from BitType images. I also included a BoundingBox as well as a ConvexHull Op. 

Additionaly I created a JUnit Test for each Op.

### DIFF
--- a/src/main/java/net/imagej/ops/geometric/interval/DefBoundingBox.java
+++ b/src/main/java/net/imagej/ops/geometric/interval/DefBoundingBox.java
@@ -1,0 +1,93 @@
+/*
+ * #%L
+ * ImageJ software for multidimensional image processing and analysis.
+ * %%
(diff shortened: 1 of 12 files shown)


dietzc:
@tpietzsch @ctrueden @DanielSeebacher lets discuss this at the beginning on the hackathon. maybe we can reuse/integrate parts of @tpietzsch imglib2-roi implementations instead of reinventing the wheel with a `Polygon` class here.

dietzc:
invalid. merged manually.

ctrueden:
FTR, it merged to the `feature-service` branch.

PR #82, opened 2014-11-20
Added first classes to Complex Binary API

apal4:
This package provides a general framework for ops that perform binary operations.
The code was migrated from net.imglib2.ops.operations.complex.binary, and adapted
to the ImageJ OPS paradigm.


hinerm:
@ctrueden @apal4 looks like this branch can be deleted because it was replaced by [`binary_new`](https://github.com/imagej/imagej-ops/tree/binary_new)... right?

PR #83, opened 2014-11-20
 Added first classes to Complex Binary API

apal4:
This package provides a general framework for ops that perform binary operations.
The code was migrated from net.imglib2.ops.operations.complex.binary, and adapted
to the ImageJ OPS paradigm.


hinerm:
@ctrueden what would you like to do with this PR? Assign it, delete it, or leave as is?

ctrueden:
@LeonYang5114 Can you please look through this code, and cherry pick any ideas you think should be implemented? We won't merge this as is, but there might be some good things in here (e.g., maybe `pow`?).

LeonYang5114:
I will look into that.

Issue #69, opened 2014-09-23
Improve prime OPS example script


ctrueden:
The OPS library is focused on image processing, and as such it would be best if the front-facing example script did more than gimmicky operations. Rather, it should perform a really common and useful—but simple—image analysis workflow. We want something succinct that elegantly performs some powerful image processing.

We need to update both the `README.md` as well as the [using-ops tutorial](https://github.com/imagej/imagej-tutorials/tree/master/using-ops).

bnorthan:
Hi 

Quick update on this.  Once I get these pull requests done it should be possible to concisely write a spot detector using ops.  

https://github.com/imagej/imagej-ops/pull/94
https://github.com/imagej/imagej-ops/pull/76
https://github.com/imagej/imagej-ops/pull/48

This code also uses IJ1 to generate a result table so it also demonstrates interoperability. 

``` python
# @DatasetService data
# @DisplayService display
# @OpService ops
# @net.imagej.Dataset inputData

from net.imglib2.meta import ImgPlus
from net.imglib2.img.display.imagej import ImageJFunctions

from jarray import array
from ij import IJ

# create a log kernel
logKernel=ops.logKernel(2, 1.0);

# convolve with log kernel
logFiltered=ops.convolve(inputData, logKernel);

# display log filter result
display.createDisplay("log", ImgPlus(logFiltered));

# otsu threshold and display
thresholded = ops.run("otsu", logFiltered)
display.createDisplay("thresholded", ImgPlus(thresholded));

# convert to imagej1 imageplus so we can run analyze particles
impThresholded=ImageJFunctions.wrap(thresholded, "wrapped")

# convert to mask and analyze particles
IJ.run(impThresholded, "Convert to Mask", "")
IJ.run(impThresholded, "Analyze Particles...", "display add");
IJ.run("Close"); '''
````

ctrueden:
@bnorthan That is beautiful!

bnorthan:
@ctrueden, @dscho, @dietzc 

Looks like you guys are getting lots done.  Thanks for merging those pull requests.  I tested this script this morning using the new 'master' and using the hela cells-image from 'sample images'.  

1.  It seems to be working for 2D images.  A good test image is the red channel of hela-cells.  That image has lots of spots.  

2.  For nd images...

One small change is needed 
logKernel=ops.logKernel(2, 1.0) --> logKernel=ops.logKernel(inputData.numDimensions(), 1.0);

[...]

Issue #54, opened 2014-09-23
Implement needed SciJava Converters


ctrueden:
Need to think about exactly which are needed, but here is a non-exhaustive list:
- [ ] `long[]` :arrow_right: `Dimension`
- [ ] `int` :arrow_right: `Shape`
- [x] `RandomAccessibleInterval` :arrow_right:  `Iterable` (`Interval`)
- [ ] `Iterable` :arrow_right: `Histogram`
- [x] `RandomAccessibleInterval` :arrow_right: `IterableInterval`
- [ ] `ImagePlus` :left_right_arrow: `ImageDisplay`
- [ ] `ImagePlus` :left_right_arrow: `Dataset`
- [x] `Dataset` :left_right_arrow: `ImgPlus`
- [x] `Img` :left_right_arrow: `ImgPlus`
- [ ] `ImgView` and `LabelingView`
- [ ] Any other converters using [Views](https://github.com/imglib/imglib/blob/imglib2-2.0.0-beta-26/core/src/main/java/net/imglib2/view/Views.java)
- [ ] Use `TypeConverter` for images and type elements

Then we can remove tons of `Map` operations!

See the [convert_numbers branch](https://github.com/imagej/imagej-ops/compare/convert-numbers) for current progress.

dietzc:
@DanielSeebacher I know that you worked on that a bit. Can you update the list above and implement the missing `Converter`s in the correct package? Depending on the conversion the `Converter` should live in imagej-common, scijava-common or (in very rare cases) in imagej-ops.

dietzc:
@DanielSeebacher please also add a `Converter` (SciJava converter) in imagej-ops which converts from a `ComputerOp` to a `Converter` (imglib2 converter) using the `ComputerConverter`.

Addtionally, we can add `convert(output, input, converter)` `Op`s to the `ImageNamespace`.

hinerm:
[ ] `int` :arrow_right: `Shape`

PR #88, opened 2015-01-05
Fix README's example, once and for all

dscho:
When running `UsingOps`, the user is greeted thusly:

```
[INFO] Found 199 ops
[INFO] Available operations:
	[111.0] (net.imglib2.img.array.ArrayImg<net.imglib2.type.numeric.real.DoubleType, net.imglib2.img.basictypeaccess.array.DoubleArray> image) = net.imagej.ops.arithmetic.add.parallel.AddConstantToArrayDoubleImageP(net.imglib2.img.array.ArrayImg<net.imglib2.type.numeric.real.DoubleType, net.imglib2.img.basictypeaccess.array.DoubleArray> image, byte value)
	[110.0] (net.imglib2.img.array.ArrayImg<net.imglib2.type.numeric.integer.ByteType, net.imglib2.img.basictypeaccess.array.ByteArray> image) = net.imagej.ops.arithmetic.add.parallel.AddConstantToArrayByteImageP(net.imglib2.img.array.ArrayImg<net.imglib2.type.numeric.integer.ByteType, net.imglib2.img.basictypeaccess.array.ByteArray> image, byte value)
	[101.0] (net.imglib2.img.array.ArrayImg<net.imglib2.type.numeric.real.DoubleType, net.imglib2.img.basictypeaccess.array.DoubleArray> image) = net.imagej.ops.arithmetic.add.AddConstantToArrayDoubleImage(net.imglib2.img.array.ArrayImg<net.imglib2.type.numeric.real.DoubleType, net.imglib2.img.basictypeaccess.array.DoubleArray> image, double value)
	[100.0] (net.imglib2.img.array.ArrayImg<net.imglib2.type.numeric.integer.ByteType, net.imglib2.img.basictypeaccess.array.ByteArray> image) = net.imagej.ops.arithmetic.add.AddConstantToArrayByteImage(net.imglib2.img.array.ArrayImg<net.imglib2.type.numeric.integer.ByteType, net.imglib2.img.basictypeaccess.array.ByteArray> image, byte value)
	[100.0] (java.lang.Object result) = net.imagej.ops.onthefly.ArithmeticOp$AddOp(java.lang.Object result, java.lang.Object a, java.lang.Object b)
	[0.6] (double a) = net.imagej.ops.arithmetic.add.AddConstantToDouble(double a, double b)
	[0.5] (long a) = net.imagej.ops.arithmetic.add.AddConstantToLong(long a, long b)
	[0.4] (float a) = net.imagej.ops.arithmetic.add.AddConstantToFloat(float a, float b)
	[0.3] (int a) = net.imagej.ops.arithmetic.add.AddConstantToInteger(int a, int b)
	[0.2] (short a) = net.imagej.ops.arithmetic.add.AddConstantToShort(short a, short b)
	[0.1] (byte a) = net.imagej.ops.arithmetic.add.AddConstantToByte(byte a, byte b)
	[0.0] (net.imglib2.IterableRealInterval<capture of ?> image) = net.imagej.ops.arithmetic.add.AddConstantToImageInPlace(net.imglib2.IterableRealInterval<capture of ?> image, capture of ? value)
	[0.0] (net.imglib2.IterableInterval<capture of ?> a) = net.imagej.ops.arithmetic.add.AddRandomAccessibleIntervalToIterableInterval(net.imglib2.IterableInterval<capture of ?> a, net.imglib2.RandomAccessibleInterval<capture of ?> b)
	[0.0] (net.imglib2.img.planar.PlanarImg<net.imglib2.type.numeric.real.DoubleType, net.imglib2.img.basictypeaccess.array.DoubleArray> image) = net.imagej.ops.arithmetic.add.AddConstantToPlanarDoubleImage(net.imglib2.img.planar.PlanarImg<net.imglib2.type.numeric.real.DoubleType, net.imglib2.img.basictypeaccess.array.DoubleArray> image, double value)
	[-100.0] (capture of ? out?) = net.imagej.ops.arithmetic.add.AddConstantToNumericType(capture of ? out?, capture of ? in, capture of ? value)
	[-10000.0] (net.imglib2.RandomAccessibleInterval<capture of ?> out) = net.imagej.ops.arithmetic.add.AddConstantToImageFunctional(net.imglib2.RandomAccessibleInterval<capture of ?> out, net.imglib2.IterableInterval<capture of ?> in, capture of ? value)
[INFO] What is 2 + 5? 7.0
Exception in thread "main" java.lang.IllegalArgumentException: No candidate 'createImg' ops
	at net.imagej.ops.DefaultOpMatchingService.findModule(DefaultOpMatchingService.java:93)
	at net.imagej.ops.DefaultOpService.module(DefaultOpService.java:110)
	at net.imagej.ops.DefaultOpService.run(DefaultOpService.java:77)
	at net.imagej.ops.DefaultOpService.create(DefaultOpService.java:391)
	at UsingOps.main(UsingOps.java:33)
```

Maybe this should be tested as part of a unit test, what with the `README` suggesting to use `createImg` and all...

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Try this Jython script in
 # create a new blank image
 from jarray import array
 dims = array([150, 100], 'l')
-blank = ij.op().createImg(dims)
+blank = ij.op().createimg(dims)
 
 # fill in the image with a sinusoid using a formula
 formula = "10 * (Math.cos(0.3*p[0]) + Math.sin(0.3*p[1]))"
@@ -55,10 +55,10 @@ sinusoid = ij.op().equation(blank, formula)
 ij.op().add(sinusoid, 13.0)
 
 # generate a gradient image using a formula
-gradient = ij.op().equation(ij.op().createImg(dims), "p[0]+p[1]")
+gradient = ij.op().equation(ij.op().createimg(dims), "p[0]+p[1]")
 
 # add the two images
-composite = ij.op().createImg(dims)
+composite = ij.op().createimg(dims)
 ij.op().add(composite, sinusoid, gradient)
 
 # display the images
--- a/pom.xml
+++ b/pom.xml
@@ -190,7 +190,7 @@
 			<groupId>net.imglib2</groupId>
 			<artifactId>imglib2-ij</artifactId>
 			<scope>test</scope>
-		</dependency>	
+		</dependency>
 	</dependencies>
 
 	<build>
--- a/pom.xml
+++ b/pom.xml
@@ -191,6 +191,11 @@
 			<artifactId>imglib2-ij</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.scijava</groupId>
+			<artifactId>scripting-jython</artifactId>
+			<scope>test</scope>
+		</dependency>
 	</dependencies>
 
 	<build>
--- a/src/main/java/net/imagej/ops/DefaultOpService.java
+++ b/src/main/java/net/imagej/ops/DefaultOpService.java
@@ -182,172 +182,172 @@ public class DefaultOpService extends AbstractPTService<Op> implements
 
 	@Override
 	public Object add(final Object... args) {
-		return run("add", args);
+		return run(Ops.Add.NAME, args);
 	}
 
 	@Override
 	public Object ascii(Object... args) {
-		return run("ascii", args);
+		return run(Ops.ASCII.NAME, args);
 	}
 
 	@Override
 	public Object chunker(Object... args) {
-		return run("chunker", args);
+		return run(Ops.Chunker.NAME, args);
 	}
 
 	@Override
 	public Object convert(Object... args) {
-		return run("convert", args);
+		return run(Ops.Convert.NAME, args);
 	}
 
 	@Override
 	public Object convolve(Object... args) {
-		return run("convolve", args);
+		return run(Ops.Convolve.NAME, args);
 	}
 
 	@Override
 	public Object createimg(Object... args) {
-		return run("createimg", args);
+		return run(Ops.CreateImg.NAME, args);
 	}
 
 	@Override
 	public Object crop(Object... args) {
-		return run("crop", args);
+		return run(Ops.Crop.NAME, args);
 	}
 
 	@Override
 	public Object divide(Object... args) {
-		return run("divide", args);
+		return run(Ops.Divide.NAME, args);
 	}
 
 	@Override
 	public Object equation(Object... args) {
-		return run("equation", args);
+		return run(Ops.Equation.NAME, args);
 	}
 
 	@Override
 	public Object gauss(Object... args) {
-		return run("gauss", args);
+		return run(Ops.Gauss.NAME, args);
 	}
 
 	@Override
 	public Object histogram(Object... args) {
-		return run("histogram", args);
+		return run(Ops.Histogram.NAME, args);
 	}
 
 	@Override
 	public Object identity(Object... args) {
-		return run("identity", args);
+		return run(Ops.Identity.NAME, args);
 	}
 
 	@Override
 	public Object invert(Object... args) {
-		return run("invert", args);
+		return run(Ops.Invert.NAME, args);
 	}
 
 	@Override
 	public Object join(Object... args) {
-		return run("join", args);
+		return run(Ops.Join.NAME, args);
 	}
 
 	@Override
 	public Object lookup(Object... args) {
-		return run("lookup", args);
+		return run(Ops.Lookup.NAME, args);
 	}
 
 	@Override
 	public Object loop(Object... args) {
-		return run("loop", args);
+		return run(Ops.Loop.NAME, args);
 	}
 
 	@Override
 	public Object map(Object... args) {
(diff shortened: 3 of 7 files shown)


ctrueden:
Thanks @dscho!

It looks like we already fixed the [UsingOps](https://github.com/imagej/imagej-tutorials/blob/9f7e9ce3d4eeffb3462a99806bb4cb39019d595c/using-ops/src/main/java/UsingOps.java) example (upon which the README was based) to use `create` instead.

See also #69.

dscho:
> It looks like we already fixed the UsingOps example (upon which the README was based) to use create instead.

I am afraid that `create` is deprecated...: https://github.com/imagej/imagej-ops/blob/ecea1714721bbd24e28487120e9a48ff7e312924/src/main/java/net/imagej/ops/DefaultOpService.java#L388-L392

dscho:
BTW I am not really proud of this change: https://github.com/imagej/imagej-ops/pull/88/files#diff-04c6e90faac2675aa89e2176d2eec7d8R46. It really flies in the face of the goal to make OPS easy to use if the user already has to deal with image factories and types just to create an image – the first thing they would try to do! Further, there seems to be a usability issue with the Op selection, still: at first, I fixed the example to comply with `CreateImgNative`'s demand that the dimensions be given as a `net.imglib2.Dimensions` instance, but `CreateImgDefault` was picked and complained rightfully that `FinalDimensions` could not be mapped to `long[]`... It should not have been picked, then, though...

ctrueden:
> I am afraid that `create` is deprecated

Right, I forgot, it was the other way around I guess. We must _not_ have updated the `UsingOps` example then...

> the user already has to deal with image factories and types just to create an image

I very much agree that we should avoid the need to pass such things as arguments. It's fine if there are signatures for it, but the simplest signature needs to do something sensible by default. And it should certainly not be the case that the README uses those signatures.

> complained rightfully that `FinalDimensions` could not be mapped to `long[]`

Right, we really need to complete #54.

> It should not have been picked, then

Agreed; will have to investigate that.

dscho:
@ctrueden so... what about merging this branch, fixing the example in the README in Fiji?

ctrueden:
There must be a way to fix this without importing an `ImageFactory` class. The title of this PR says it will fix the README "once and for all", but we certainly don't want to have this ugly reference in the README "once and for all"...

dscho:
The problem this PR tries to address is that the example from the README *won't work*. It does *not* try to fix the ugliness that has been introduced while breaking that example.

I am sympathetic with your goal to fix the ugliness, but I disagree that it should hold up this bug fix even further.

ctrueden:
What I don't want is to post code that will potentially be copy-paste replicated many times. If we introduce that ugliness to master, it will ripple outward, and hence take a long time for it to "fade away."

I do agree that leaving the README broken is terrible. So I will spend some time now investigating this issue and coming up with a proper fix. It will be one less thing to solve at the hackathon, then.

dietzc:
maybe, as I already told you @dscho in person, it makes sense to wait until https://github.com/imagej/imagej-ops/pull/48 has been merged? I think @bnorthan addresses some of the issues by implementing various new CreateImage Ops.

dscho:
> as I already told you @dscho in person, it makes sense to wait until #48 has been merged?

@dietzc sure. It would have been good to know about such intentional delays *before* I gave two presentations on *ImageJ OPS*, though. That way, I could have possibly held off and avoided disappointing people with yet another project that promises a lot but fulfills little (I am talking about impressions here, not facts, but in the long run impressions count more, as you well know).

ctrueden:
@dscho I'm on it; we'll have the readme working before the end of the day.

bnorthan:
@ctrueden, @dscho, @dietzc 

Yes Christian and I have been discussing the createImg issue in the port_kernels pull request https://github.com/imagej/imagej-ops/pull/48.

We have some ideas, more review and input is welcome, maybe at the Hackathon??  I can't make it but maybe I can follow some of the discussions on Skype??

The 'createImg' part of the Readme works with that pull request... however the equation part is really slow.  I'm not sure why.  It eventually finishes.

But then it fails when trying to add a constant.  The AddConstantToPlanarDoubleImage op seems to be run even though the input is a 'FloatType' image.

ctrueden:
OK, I have this sorted out. My commit to fix the README behavior is: 41ffba8757f1bdc87e2ed6579b59057cde64b7c2.

This "does the right thing" by adding a createimg signature that takes only a `long[] dims`, so that the README stays simple and keeps working as is. Then we no longer need that factory complexity.

Thanks @dscho for that awesome unit test. It indeed fails without 41ffba8757f1bdc87e2ed6579b59057cde64b7c2, and passes with it. BTW, I also fixed up the `DefaultOpService` to use the name constants, and cleaned up the discrepancy between those built-in signatures and the built-ins listed in the `net.imagej.ops.Ops` class. And downcased `createImg` everywhere, as it should have been.

ctrueden:
The `CreateImg` implementations will need further improvement in the future. In particular, we do not want to hardcode `ArrayImg` for all possible `long[]` dimensions, but rather use a heuristic like how SCIFIO does now. But that is a problem for later (hackathon or otherwise).

dscho:
Thanks.

PR #89, opened 2015-01-09
Image moments

DanielSeebacher:
Add image moment features. Our results were compared with the results from the OpenCV Library.


dietzc:
invalid. merged manually.

PR #94, opened 2015-01-15
Fft filters

bnorthan:
This is just for review right now as it depends on 

https://github.com/imagej/imagej-ops/pull/76

and 

https://github.com/imagej/imagej-ops/pull/48

This pull request introduces

1.  A class hierarchy for FFT based filters including linear filters (like convolve) and iterative filters (for example Richardson Lucy deconvolution).

2.  A new convolution op that is hopefully different enough to relicense.

### DIFF
--- a/src/main/java/net/imagej/ops/DefaultOpService.java
+++ b/src/main/java/net/imagej/ops/DefaultOpService.java
@@ -216,6 +216,11 @@ public class DefaultOpService extends AbstractPTService<Op> implements
 	}
 
 	@Override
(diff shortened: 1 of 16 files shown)


dietzc:
just two or three comments on this PR: Maybe we can introduce interfaces for Convolution in general (to also support e.g. imglib2-algorithm SymmetricSeperableConvolution) and also use the Converter Framework to reduce the number of required ops?

ctrueden:
See also #97.

Issue #100, opened 2015-01-24
Op expression parser


iarganda:
```
ops.expr("sqrt(dx*dx + dy*dy + dxy*dxy)", {"dx" => dx, "dy" => dy, "dxy" => dxy});
```
Which means:
```
ops.run("sqrt", ops.add(ops.add(ops.multiply(dx, dx), ops.multiply(dy, dy)), ops.multiply(dxy, dxy)));
```

LeeKamentsky:
Holey moley - nice, nice nice. It's starting to look Pythonic.

ctrueden:
@LeeKamentsky Heh, we have a ways to go yet. The issue is still open, after all!

ctrueden:
Closing in favor of #105.

Issue #19, opened 2014-04-06
Autogenerate methods for built-in ops


ctrueden:
The `OpService` provides a method signature for each operation name. They all take `Object... args` for the parameters. Instead, we could code generate the `OpService` class using a template, so that it has compile-safe signatures for *all* built-in ops. It would be slick if it was feasible (though I have my reservations).

ctrueden:
I agree with @dscho that instead of autogenerating, let's add a "completeness" test that fails the build if the `OpService` interface is missing any core op signatures. That will be very doable. I am pulling this into an earlier milestone, because I think having the compile-safety will be extremely helpful in many ways.

ctrueden:
I started work on this on the [built-ins](https://github.com/imagej/imagej-ops/compare/built-ins) branch, which adds a fancy unit test that verifies all `@BuiltIn`-annotated methods of `OpService` cover the set of available `Op` implementations, and vice versa.

However, I ran into trouble code generating the method signatures including generics. I [posted on SO for help](http://stackoverflow.com/q/28143029/1207769).

Fortunately, this functionality is not strictly required for the checking feature, since the actual type matching (especially between generic `Type` and `Type`, which is what the unit test is able to use in this scenario) in OPS is pretty robust. In other words: we can manually do the work of adding all 300 method signatures needed, being careful to get the generic parameters right by hand, and the unit test should at least inform us whether we are "close enough" for things to generally work.

That said, it would be ideal if the unit test could spit out the exact missing method signature which needs to be added for each uncovered `Op` implementation. But doing that would require a deeper understanding of generics than I have right now.

ctrueden:
I thought of an easier and more flexible way to do this. The type signature comparison logic I wrote (still pending on the `built-ins` branch) will still be useful, but in general we do not need to be slaves to whether everything matches. Better is for the `@BuiltIn` annotation to take an `op=Class<? extends Op>` parameter, which we can use to trivially determine which available ops do not have corresponding built-in method signatures.

The argument type matching on top of that is just gravy: it might be nice to warn if the signature could not possibly match the purported op class, but really that is an increasingly complex problem which does not really need to be solved. On the contrary, our actual requirements are:

1. Know that each concrete `Op` plugin has at least one built-in method signature in the `OpService`. A parameter on `@BuiltIn` would accomplish this nicely.
[...]

PR #105, opened 2015-01-27
Start implementing an expression parser

dscho:
Please do not merge this just yet.

A couple of things are left to do, still:
- [ ] ensure that  `"a+b"` does not modify `a` (the 2-parameter "math.add" op modifies the first parameter if it is an image; I think that is fine, but we should teach the expression parser to generate a temporary image and use the 3-parameter op)
- [x] add more operators and use the [Shunting yard algorithm](https://en.wikipedia.org/wiki/Shunting-yard_algorithm) to handle operator priorities (pointed out by @ctrueden)

A few more things need to be done, but they can be done in different Pull Requests (and by people other than myself :grinning:): optimize the temporary variable usage (i.e. reuse the temporary image in "a+b*c"), and add a way to generate new ops based on entire expression (sub-)trees using Javassist (one idea is to add a hook to the `ExpressionParser.bind(map)` method that reasons over arithmetic operations by inspecting `OpExpression`s: when the op name is e.g. "math.add" and the two parameters are either a `Variable` of type `PlanarImg` or `ArrayImg` with a primitive backing store, or a to-be-introduced `JavassistedOpExpression`, generate a new `JavassistedOpExpression` based on those parameters and execute that in `OpExpression#eval()` instead.

@tinevez the idea for this Pull Request is based heavily on your Image Expression Parser, thank you so much for that!

Even if this developer was not aware of #100, this Pull Request seems to address it!

### DIFF
--- a/src/main/java/net/imagej/oops/Constant.java
+++ b/src/main/java/net/imagej/oops/Constant.java
@@ -1,0 +1,67 @@
+/*
+ * #%L
+ * ImageJ software for multidimensional image processing and analysis.
+ * %%
+ * Copyright (C) 2014 - 2015 Board of Regents of the University of
+ * Wisconsin-Madison, University of Konstanz and Brian Northan.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * 
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE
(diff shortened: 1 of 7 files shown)


dscho:
Okay, I am done with what I wanted to do (and I'll leave the problem of "a+b" modifying the first parameter to you, @ctrueden, since you already had an idea how to fix it).

ctrueden:
Just an update on this work: we now have a much less rudimentary expression parser as [its own SciJava project](https://github.com/scijava/scijava-expression-parser). The plan is to create an `expr` op here in OPS, which leans on SJEP to do the parsing, and OPS to do the evaluation. We will still need to address the "in-place" vs. "functional" issue in the OPS framework (first bullet point in the description above). Then we'll be able to close out this issue. The conceit is then that SJEP's data structures are flexible enough for the future addition of an "OOPS" optimization layer which can do syntax tree transforms, etc.

ctrueden:
Other thoughts and considerations for the future (maybe turn into separate issues later after this one is resolved):
* Rather than our own home baked parser, we could just use a nice scripting language like Jython. This was my original idea, which is why the README gives a Jython example after all. But we need to [overload all the Jython operators](https://docs.python.org/2/reference/datamodel.html) to "do the right thing" and call the OPS framework for evaluation.
* We could also support compiling expressions somehow for performance. That is: for a given JVM, would be nice to do the discovery of the right op(s) once, then have Java bytecode that does the actual work, instead of needing to use reflection. This consideration might improve both performance and correctness. It might even make it easier to have type-safe compile-time access to all the built-in ops, too; see also #19.

ctrueden:
The [SciJava Expression Parser](https://github.com/scijava/scijava-expression-parser) is really polished now, and ready for `1.0.0` release. I am just validating the API a little more first, by using it in OPS.

I have pushed an [expression-parser branch](https://github.com/imagej/imagej-ops/compare/expression-parser) which uses it. There is still thought and work to be done to ensure only "pure functional" ops are used by the evaluator—i.e., `BOTH` parameters need to be handled specially (perhaps always filled with null in the case where they are optional, or perhaps not matched at all).

ctrueden:
This work continues in PR #113.

PR #106, opened 2015-01-27
Generate more math ops for primitive types

dscho:
This Pull Request covers all the low-hanging fruits this developer could find...

Incidentally, it also changes the generator to make sure that stale generated source files are removed. This should help in particular developers who are not involved in augmenting the code generation because it is no longer required to call `mvn clean` from the command-line when the name of a generated file is changed.


dscho:
@ctrueden just for the record, I think this is ready to be merged... ;-)

ctrueden:
@dscho Thanks, I understand. It's just that I did something very similar on the `primitives` branch, which I am in the process of rewriting to include your ideas here as well. My final version will merge in this history as well so that it is not lost. Probably will finish it tomorrow.

ctrueden:
Everything is reconciled now! And there are even unit tests with af05aff6f628155d12352a0c5fabaec0843c8d8b.

PR #110, opened 2015-01-29
For Review: OpResolverService

dietzc:
see for examples how to use the `OpResolverService` (former FeatureService): https://github.com/imagej/imagej-ops/commit/9ef46210f7bc0e34db5e3a83568b35e942e02acb.

# TODO:

* JavaDoc is missing, but I will add it asap!
* The `OpResolverService` can definitively be enhanced and tuned (for example set-up a graph containing all possibilities how to resolve a set of ops and find an optimal solution with e.g. the highest priority over the ops). 
* I'm really unhappy with all the  unsafe casting stuff I have to do (see Example). @ctrueden any ideas?


ctrueden:
@dietzc Really impressed with all your efforts!

Can you point me more specifically to where you are doing the undesirable unsafe casting? I wasn't sure what you meant by "Example"...

dietzc:
see: https://github.com/imagej/imagej-ops/blob/outputop-service/src/main/java/net/imagej/ops/features/Example.java

ctrueden:
Would it make sense to convert that Example into either: A) a unit test; or B) a tutorial; or C) both?

ctrueden:
OK, sorry for being thick, but... the [this example](https://github.com/imagej/imagej-ops/blob/outputop-service/src/main/java/net/imagej/ops/features/Example.java) does not have a single unsafe cast that I could see. (Except for `Math.random()` to `float`, which is surely unrelated to your concern??)

dietzc:
About the Unit-Tests: Are https://github.com/imagej/imagej-ops/tree/outputop-service/src/test/java/net/imagej/ops/features and https://github.com/imagej/imagej-ops/tree/outputop-service/src/test/templates/net/imagej/ops/features good enough or do you mean more explicit ones?

About the Casting: Sorry for beeing not specific enough! Let's do one question after the other: 

One thing I really don' like is, that https://github.com/imagej/imagej-ops/blob/outputop-service/src/main/java/net/imagej/ops/features/DefaultAutoResolvingFeatureSet.java#L69 can't make sure, that the provided type of the `Op` actually is an `OutputOp<T>`. I was thinking about an `OutputOpRef`, but I don't like this either.

Any ideas?

Thanks for starting the review process!

ctrueden:
Is this branch still just awaiting review? Or being actively developed still? The list of commits has gotten very long and includes variations of commits already on master, so things have gotten messy...

dietzc:
I will split this PR in smaller, easier to review branches and PRs. Hope this helps.

jaywarrick:
Just to document my experience thus far... I haven't run rigorous tests with benchmarks (e.g., against matlab versions of calculations etc) but I can now say that all the FeatureSets (except Haralick3D which I haven't tried at all) seem to work for me now (however, see https://github.com/imagej/imagej-ops/pull/155 for latest on fixes for HistogramFeatureSet). For me, the feature sets require imglib2-roi 0.3.2-SNAPSHOT for bounding box calculations on LabelRegion objects and seem to be most easily leveraged using the latest ConnectedComponents in 0.3.1-SNAPSHOT of imglib2-algorithms.

The only thing Curtis and I can't explain quite yet is some odd behavior with ConnectedComponents (imglib2-algorithms) and Img<>'s created with ImageJFunctions.wrapByte(ImagePlus) which might effect how many people might initially use these feature sets. ConnectedComponents would work with Img<> objects created with SCIFIO ImageOpener but not with Img<> objects created by wrapping ImagePlus objects. Curtis's initial foray into the matter suggests it is in the ConnectedComponents class. I'll add comments if I learn more. Here is a test project I made to debug this issue (https://github.com/jaywarrick/ConnectedComponentsTest)

Cheers and good luck with the pull request etc. Thanks!

dietzc:
closed for refactoring

PR #113, opened 2015-03-03
Expression parser!

ctrueden:
This branch adds an `eval` op that evaluates an expression using the [SciJava Expression Parser](https://github.com/scijava/scijava-expression-parser).

It also adds a `Main` class which launches an evaluator console so that you can try it out. With this console, it is possible to chain execute ops by name, as well as using standard operators such as `+`, `-`, `*`, `/`, etc.

Since it was vaguely relevant, I also added some more useful ops, such as logic operators for primitives, bitwise complement of primitives, and a `help` op that does what `OpService#help` did/does. And I renamed a couple of ops with overly verbose names, too.

Everything is not perfect, but it is well tested and IMHO good enough to merge to master now. The quality of the evaluation is of course dependent on the actual available ops and how they get matched—we will need to improve the op matcher and SJC converter in a couple of ways:

* Do not allow SJC to auto-convert lossily (e.g., `double` should not match `int`).
* Add converters from primitive numeric types to `BigInteger` and `BigDecimal`.
* Add `OpService#runFunctional` or similar that does not match populated `BOTH` parameters, since `eval` is intended to behave in a "pure functional" way, at least for operators.
* Do something about the fact that ops in a namespace cannot be referenced by their names (i.e., `math.add(1, 2)` does not work since `.` is treated as a standard operator).

But none of the above issues are roadblocks to merging, as far as I'm concerned.

@dietzc @hinerm @dscho @bnorthan Feedback welcome! I'll merge within a few days unless you see any serious flaws in the design.


ctrueden:
No one? Nothing? OK then. It was a lot of work. I hope you like it, @dscho. Also @iarganda, I hope you like it too. We are getting closer to what you wanted OPS to be...

iarganda:
Sorry I haven't had time to look into it yet, but i'm looking forward! Thanks a lot!

ctrueden:
@iarganda: No worries, I didn't ping you about it before. Just wanted you to be aware that we have a first cut `eval` op now, and it's not too bad. But it still has limitations (see above). Will work on it as time allows.

PR #114, opened 2015-04-02
Decon fixups

bnorthan:
This pull request contains several format fixes, bug fixes and minor adjustments.   I can merge it myself but I will wait a couple of days in case anyone has any concerns or suggestions.  The only "non-decon/fft-related" file I changed was Ops.list.  (I can put the format fix in a separate pull request if necessary  The individual commits should be readable.)  

- fixed formatting on several files
- added a correlation op
- changed default first guess for RichardsonLucy algorithm to be consistent with DeconvolutionLab (todo: make first guess type a parameter)
- added a 'deconvolution' namespace in Ops.list


ctrueden:
Thanks @bnorthan!

Issue #57, opened 2014-09-23
Add mechanism to "reduce" an op to another op


ctrueden:
Given a set of "op parameters" (name, type and args), the `OpMatchingService` works in multiple stages:
1. Find the op **candidate** implementations with matching name and/or type.
2. Narrow down the candidates to **exact matches**:
   - Filter down to those with the same _number_ of arguments
   - Filter down to those with compatible _types_ of arguments
   - Exclude those that implement `Contingent` and reject the specific args given

Note that "compatible" above means "able to be converted using the SJC `ConvertService`."

We will add another phase of the matching process, as follows:
- Define an `OpParams` class (tentative name) that encapsulates the (name, type, args) as a single object.
- Define some SJC `Converter` implementations that know how to "reduce" from certain `OpParams` to other `OpParams`. This is very useful in the case when an op ends up being a simple wrapper around another op, which is very common.

For example: these converters could define the following reduction path:
- `normalize(outputImgPlus, imgPlus, axisTypes)`
- :arrow_right: `normalize(outputImg, img, axisIndices)`
- :arrow_right: `slicewise(outputImg, img, op(normalize), axisIndices)`
- :arrow_right: `map(...)`

Then the `OpMatchingService` would need to—when an exact match is not found—attempt to apply a reduction then repeat the matching algorithm, repeating as needed until no more reductions can be found (excluding cycles).

See the [match-conversion branch](https://github.com/imagej/imagej-ops/compare/match-conversion) to follow progress on this issue.

ctrueden:
We will call the data structure `OpArgs`.

ctrueden:
#74 introduces the same `OpArgs` data structure.

ctrueden:
@dietzc and I changed our minds, and are going to call it `OpRef`, since it is basically a "reference" to an op that _could_ exist, should you need it.

ctrueden:
[...]

PR #115, opened 2015-04-29
Improve varargs matching

hinerm:
Add logic to handle the case where varargs may be split out into individual arguments, instead of aggregated in an array.

### DIFF
--- a/src/main/java/net/imagej/ops/DefaultOpMatchingService.java
+++ b/src/main/java/net/imagej/ops/DefaultOpMatchingService.java
@@ -32,6 +32,7 @@ package net.imagej.ops;
 
 import java.lang.reflect.Type;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 import net.imagej.ops.OpCandidate.StatusCode;
@@ -93,7 +94,12 @@ public class DefaultOpMatchingService extends
 		}
 
 		// narrow down candidates to the exact matches
-		final List<Module> matches = findMatches(candidates);
+		List<Module> matches = findMatches(candidates);
+
+		// No matches were found, so we will try to recapitulate varargs and, if
+		// successful, attempt matching again.
+		if (matches.isEmpty() && condenseArgs(ref)) matches =
+			findMatches(candidates);
 
 		if (matches.size() == 1) {
 			// a single match: return it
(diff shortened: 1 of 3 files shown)


hinerm:
@ctrueden since you love reviewing PRs so much, would you mind signing off on this?

ctrueden:
@hinerm and I talked for quite a while about how best to handle varargs in OPS. The issue with this PR's approach is that it only collapses tail elements with identical types. So it solves the `createimg` case where you give a list of `int` values and want it transformed into an `int[]` argument. But it does not solve mixed type cases, such as passing an `ArrayImg` and a `PlanarImg` and expecting it to match an op with its last parameter being `Img[]`. And solving the mixed type case is much more involved, especially when combined with the OPS framework's existing matching of optional parameters—the current approach of padding parameters to fill nulls would no longer be sufficient to handle varargs too.

As such, we agreed to close this PR, in favor of pursuing the built-ins work (#19). @hinerm is going to add all the `createimg` built-in signatures in a separate PR, which definitely solves the issue [recently discussed on imagej-devel](http://imagej.net/pipermail/imagej-devel/2015-April/002560.html). And in general, the built-ins approach solves varargs for all built-in ops.

What it does not provide is a fully dynamic (i.e., extensible) varargs solution for ops outside the core imagej-ops component. We could file a separate issue to tackle that at some point in the future, though from my perspective, it is very low priority. The vast majority of scripts will be calling into the `OpService` built-in methods—and furthermore, any third-party extensions can provide their own service class that similarly provides built-in type-safe method signatures. They could even reuse the forthcoming utility code for unit testing that all built-in methods are indeed present and type compatible.

ctrueden:
One potentially slick way of providing fully dynamic varargs would be in conjunction with op reducers (#57). It would be possible to create a general op reducer that effectively does the second "pass" this PR did, to look for matching signatures with collapsed arguments.

hinerm:
Note that the issue this PR set out to address (`createimg(int, int)` not matching) is now resolved by https://github.com/imagej/imagej-ops/commit/f96f76692577c38444e51bdf4c1b7264f2611f06

ctrueden:
OK, so I stupidly reviewed part of this again. Sorry for the noise. At least my new comments are reasonable!

PR #116, opened 2015-04-29
DO NOT MERGE

hinerm:
Anticipatory PR to allow for discussion on the scope of built-ins


hinerm:
An additional feature of built-ins is that they will be used to resolve varargs.

We can achieve this by ensuring the `OpService` signatures do not contain terminal `array[]` arguments, but use the varargs `...` instead.

Issue #98, opened 2015-01-24
Convolve only works on extended RAIs


ctrueden:
When attempting to work with `ConvolveNaive` with a 1D kernel of size 3, the op throws `ArrayIndexOutOfBoundsException` right away because it attempts to position the input image at index `-1`. I believe this is correct behavior in the sense that the convolution is being computed at the image boundary, so an `OutOfBoundsStrategy` needs to be associated with the input image.

How should we address this? Should we update the `conforms` method to fail early if the given input does not have an `OutOfBoundsStrategy`? Or should we transparently extend the input with one in the case that it doesn't already have one? What do you think, @bnorthan? @dietzc?

bnorthan:
Hi Curtis

How is this being handled for other windowed operations like Median Filter??  Maybe Median Filter, Naive Convolve, and other similar operations that perform a calculation based on a windowed region should derive from a common base class that takes care of this logic?

dietzc:
@bnorthan I agree. We should have one naive convolution class which takes another `Function` or `Op`. I want to investigate if we can reuse the `RectangleShape`s in this context and optimize them (Basically, I need to find out if we can assume `FlatIterationOrder` in this case). . Then tons of algorithms will benefit from this optimization. @tpietzsch already pointed me to some ideas how to make this faster.

bnorthan:
Hi @dietzc any more thoughts on this??  

I think we need a base class for spatial neighbourhood operations.  This class could have a border size (in pixels) and an out of bounds factory.   This class would then set up the RAIs for the low level algorithm.  It also needs to be set aware.  For example xyz would be different then xy-time.  In the second case we need to iterate and apply the operation to each 2D algorithm.

dietzc:
Hi @bnorthan,

as mentoined above, we already have a very general version using `Map` in combination with `Shape` or to be more precise `Neighborhood`s. However, this very general solution is too slow. Therefore I think, as you've already mentoined, we should have a base class for the general case which is optimized (e.g. only uses the OutOfBoundsStrategy when required (not for each pixel, just the border pixels). Additionally, we might want to have very special implementations for certain container types (e.g. I think we could have fast versions for `ArrayImg`s and for certain dimensionalities (we can be super fast on 2D, 3D, 4D etc images if we directly access the underlying containers).

[...]

PR #117, opened 2015-04-30
Naive convolution

bnorthan:
@ctrueden, @dietzc - This pull request fixes the NaiveConvolution boundary problem mentioned in #98, and also mentioned on the [imagej-dev listserv](http://imagej.net/pipermail/imagej-devel/2015-April/thread.html).  Long term it could be helpful to use a common base class for efficient neighbourhood operations.   Let me know if anything along those lines all ready exists, and I will update this.


ctrueden:
Fantastic!

dietzc:
@Squareys will work on efficient neighborhood operations during the next weeks.

dietzc:
One more comment: Shouldn't we operate on `RandomAccessibleInterval` or even `RandomAccessible` instead of `Img`s?

Issue #96, opened 2015-01-20
Use a heuristic for default behavior of createimg


ctrueden:
When calling createimg, whether to use an array, planar or paged cell image container should depend on the size of the dimensions. It could also be specified as a hint. Let's steal the [logic from SCIFIO](https://github.com/scifio/scifio/blob/scifio-0.17.2/src/main/java/io/scif/img/DefaultImgFactoryHeuristic.java#L46-L64).

ctrueden:
Other classes that implement `createOutput` may have the same issue; e.g.: the [AbstractFFTFilterImg](https://github.com/imagej/imagej-ops/blob/91349a409288cc561d655ba2037f532e9c549f3f/src/main/java/net/imagej/ops/fft/filter/AbstractFFTFilterImg.java#L138-L142).

ctrueden:
The SCIFIO logic, because it depends on the amount of available RAM, is not deterministic. So we should probably depend on static cutoffs for the heuristic instead.

dietzc:
it's not only about create images but also about creating new empty containers from `RandomAccessibleInterval`s (e.g. Views). We should also keep in mind to have helper/utilities for `Labeling`s

dietzc:
I'm working on tests for `CreateImg` at the moment which will serve as the basis for further CreateImg methods or some additional `Converters` [1]. However, I'm facing a problem which confuses me a bit:

In the test https://github.com/imagej/imagej-ops/blob/img-creators/src/test/java/net/imagej/ops/create/CreateImgTest.java (WIP) I receive an `IllegalArgumentException` in some logic inside the `DefaultOpMatcher`. I can't figure out why this happens, maybe @ctrueden or @hinerm you have an idea?

[1] https://github.com/imagej/imagej-ops/tree/img-creators

ctrueden:
@dietzc: It's due to an issue with how the `DefaultConverter` handles arrays and collections right now.

Debugging into your test, I see that the `ConvertService` returns `true` for `canConvert(Object arg, Type type)` when `arg` is a `net.imglib2.view.IntervalView` and `type` is a `double[]`, because `DefaultConverter#canConvert` returns `true`, because `dest` (the `type`) [is an array](https://github.com/scijava/scijava-common/blob/scijava-common-2.38.2/src/main/java/org/scijava/convert/DefaultConverter.java#L291-L292).

[...]

Issue #50, opened 2014-09-17
Discard optional "both" parameters preferentially to "input" parameters


dietzc:
We want the `out` parameter of `OutputFunction` implementations to be discarded preferentially to later optional inputs, if any. We can accomplish this in a general way by preferring optional "both" parameters (also from right to left) to pure "input" parameters, when deciding the order of optional inputs to discard.

The alternative would be to check against every possible combination of missing optional parameters, but that is exponential (e.g., twenty missing optional inputs would be 2^20 > 1 million combinations already...). So let's stick to right-to-left, but do "both" parameters first followed by "input" parameters.

ctrueden:
Fixed in 84c680057a6344bf66342807b2181abe32e7f51e. Tested in 91349a409288cc561d655ba2037f532e9c549f3f.

ctrueden:
Note that this change was reverted in 9335808ae2b3328341e371708b13f67577de7407, for reasons outlined in that commit message.

PR #123, opened 2015-05-13
Img creators

tibuch:
Redesigned of the "Create" interfaces. Introduced type and factory heuristics to simplify the image creation process i.e. images can now be created using a simple oneliner like ops.run(CreateImg.class, new FinalInterval(100, 100));. Also added converters from int and long arrays to Dimensions for further simplification. 

Also added CreateLabeling which now returns the ImgLabeling introduced in imglib2-roi. 

Tests can be found in the CreateImgTest and CreateLablingTest classes. 

@ctrueden is it somehow possible, that we can omit null for not required parameters? Currently we have to insert null so that the parameter order in the method call is exactly the same as the parameter order of the op. 
e.g.: https://github.com/imagej/imagej-ops/blob/img-creators/src/test/java/net/imagej/ops/create/CreateImgTest.java#L93-95


ctrueden:
Thanks, guys!

I see there is still a WIP commit. Is this branch considered ready for merge?

> is it somehow possible, that we can omit null for not
> required parameters? Currently we have to insert null so
> that the parameter order in the method call is exactly the
> same as the parameter order of the op. 

No. We tried, but it causes too many problems. See:
* https://github.com/imagej/imagej-ops/commit/9335808ae2b3328341e371708b13f67577de7407
* https://github.com/imagej/imagej-ops/issues/50

However, the addition of type-safe method signatures for all core ops, currently being done by @awalter17 on the [built-in-implementations branch](https://github.com/imagej/imagej-ops/compare/built-ins-implementation), will make the situation far less confusing.

DanielSeebacher:
I'm really sorry, I must have overlooked the WIP commit when I cleaned up. But this shouldn't be a problem, the JUnit Tests are already implemented, see: https://github.com/imagej/imagej-ops/blob/9d451d39b7e82c2bfc31bbd1d3cfd4e321cf1f13/src/test/java/net/imagej/ops/create/CreateImgTest.java

dietzc:
@DanielSeebacher you may get rid of my commit. Feel free to force push :) thanks!

bnorthan:
Hi Daniel

I noticed the the DefaultCreateType() op always returns a DoubleType, so does this mean that the default image type is always double?  

One of the things @dietzc and I discussed was a way to specify default values for the img type.  Which is why, originally, we had an overly complicated AbstractCreateImg.  

A lot of 'createSomething' ops will have a similar pattern.  If the user supplies a type, create the image using the type, if not, create the image using a default type.  So you have code something like this...

``` java
@Parameter(required=false)
Type theType;

public void run() {
    if (theType!=null) {
        // we have a type
        img=ops.createImg(dims, theType);
    }
    else {
       // we don't have a type, just use Float
       img=ops.createImg(dims, new FloatType() );
     }
```
It might be nice if the 'createImg' op also took a default type, so that we could make the code more concise.  

``` java
     @Parameter(required=false)
     Type theType;

public void run() {
      // create an image, use FloatType if the caller did not pass in a type
      img=ops.createImg(dims, theType, new FloatType()); 
}
```

Does that make sense??  It's not a huge deal.  Just something we had discussed before.  Good work!

dietzc:
Hi @bnorthan,

we tried to generalize this concept here by providing the default type centrally in the 'CreateType' op, as I think in most of the cases, a user wants to follow the same strategy (e.g. use `DoubleType` if no type is provided). However, `CreateImg` implementors can still specify their own default type if they want or an usser can simply override the `DefaultCreateType` with another one with higher priority. 

Does this make sense? Probably I'm missing something :-)

bnorthan:
> But I wouldn't provide the type and the default type in the same method signature. this is a bit confusing for the caller. I would to this outside if outType == null pass FloatType, else outType

@dietzc that's fine, but that is why we had AbstractCreateImg, it was an attempt to isolate that logic but it may have made things too confusing.

dietzc:
@bnorthan it would be awesome if you could have a look if `CreateImg` fulfills all your requirements now? I'm happy to discuss/change things!

bnorthan:
@dietzc

It looks good.  I am still considering what to do with `AbstractCreateKernelImg` (formally `AbstractCreateImg`) 

I changed `AbstractCreateKernelImg` so that it now uses the new `createimg` op.  I am wondering what if anything else should be done to improve it. 

The purpose of this class is to implement logic to:

1. Check if the user passed in a type
2. Check if the user passed in a factory
3. Use defaults if type and/or factory were not used
4. Override 'system' defaults if desired (for example the default createType returns DoubleType, so in turn DefaultCreateImg will use DoubleType if nothing is provided.  I wanted a mechanism to override that). 

Should this logic be implemented somewhere in the create namespace??  Or is it OK as is??  Should I consider a different design??  

https://github.com/imagej/imagej-ops/blob/img-creators/src/main/java/net/imagej/ops/convolve/kernel/create/AbstractCreateKernelImg.java

dietzc:
I added some more changes to this PR and I'd be happy if one of you @bnorthan @ctrueden @hinerm ... could review and potentially merge it. We can also close this https://github.com/imagej/imagej-ops/issues/96 issue afterwards.

So the central points of this PR:
* Add more create methods (e.g.`CreateImgLabeling` or `CreateNativeImg`).
* Implement default Ops/Heuristics to create `ImgFactory` or `NativeType` in case no type is provided from an user (see `CreateType` and `CreateImgFactory`).
* Add two converters to create `Interval`s from arrays. 

Only thing I'm not entirely sure about is if the way I added shortcuts to the `CreateOps` in the `OpsService` is the preferred way. @ctrueden any comments?


@bnorthan I suggest that we try to merge this PR and then I promise I will have a look again at `AbstractCreateKernelImg` and I'm happy to further discuss the design with you, as this sounds really like fun to me :+1: Sounds good?

hinerm:
@dietzc this doesn't currently build for me..?

dietzc:
?! builds perfectly fine for me with mvn

hinerm:
It builds with Java 8. If I do a clean build with Java 6, I get generics problems and must wear the octopus of sadness :octopus:

dietzc:
Using Java 7 here. Thanks for reporting. I will have a look.

hinerm:
Since this breaks API let's make sure we bump the minor version.

@ctrueden has set a good precedent for pre-emptively bumping versions when breaking API/SPI (e.g. this would put imagej-ops to 0.13.0-SNAPSHOT right now so there's no question of what version to release next).

Not a requirement, just a FYI and self reminder.

DanielSeebacher:
@hinerm is right, this project doesn't currently compile with ```mvn clean install``` using Java 6. I think this is caused by this [bug](http://bugs.java.com/view_bug.do?bug_id=6932571) as described in this [Stackoverflow post](http://stackoverflow.com/a/4829643/2275786). This can be fixed by first casting to Object and then casting to T. I fixed this problem in bcbe198dba2d185fb8d270edd5a82fd628376621.

dietzc:
it compiles after rebase again. Can we merge?

hinerm:
@dietzc @tibuch @DanielSeebacher thanks for this.

For future PRs, it will likely speed up the merging process if you use the ImageJ [eclipse style templates](http://imagej.net/Coding_style#Eclipse_code_style_profiles). You can set project-specific templates, so this shouldn't affect any style you may be using for other KNIME development.

A clean git history (i.e. not changing the same code multiple times within the same PR) also is helpful for review, but probably less important than code style in figuring out what the PR is actually doing.

Anyway, none of this is mandatory or anything - just wanted you to know what will give the best chance of a speedy merge. If you have any troubles/questions with style stuff, @ctrueden or I can always help.

PR #124, opened 2015-05-13
Real binary math

apal4:
Tests for RealBinaryMath now generated- ComplexBitType is still separate

### DIFF
--- a/src/main/templates/net/imagej/ops/math/RealBinaryMath.list
+++ b/src/main/templates/net/imagej/ops/math/RealBinaryMath.list
@@ -1,0 +1,26 @@
+# Generated unary and binary arithmetic ops with primitive types.
+#
+# Many of Java's binary numeric operators only work on ints and longs.
+# And still more of them purport to work on bytes and shorts, but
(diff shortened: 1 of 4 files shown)


ctrueden:
I cleaned up this branch. It is looking pretty good now, but still needs more thorough review. In my initial investigation, I noticed that it causes another preexisting test (`AddConstantToNumericTypeTest`) now fails. Will try to dig further next week.

hinerm:
@ctrueden test fails because `RealBinaryMath#RealTypeAdd` is also matching the use case in this test, and takes priority since `AddConstantToNumericType` is low priority.

hinerm:
To fix we just have to update the numeric type tests to use types that are `NumericType` and are not also `RealType`.

hinerm:
fixed in https://github.com/imagej/imagej-ops/commit/6357b651faa934426d73bb26df061a4c05d7d3a2. Squash if you want @ctrueden

ctrueden:
Thanks @hinerm!

PR #127, opened 2015-05-21
Implemented more tests for map

tibuch:
Implemented more tests for map. Especially tested map against different input/output dimensions and added contingent interface if needed.

### DIFF
--- a/src/main/java/net/imagej/ops/map/MapIterableIntervalToRAI.java
+++ b/src/main/java/net/imagej/ops/map/MapIterableIntervalToRAI.java
@@ -30,6 +30,7 @@
 
 package net.imagej.ops.map;
 
+import net.imagej.ops.Contingent;
 import net.imagej.ops.Function;
 import net.imagej.ops.Op;
 import net.imagej.ops.Ops;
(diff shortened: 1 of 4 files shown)


dietzc:
>Is there some reason MapView can't simply implement Function<I, O>? Then it would be a InputOp<I> and OutputOp<O> as well as Threadable and a Converter<I, O>. I haven't thought carefully about this though; maybe there is a reason this op shouldn't be Threadable?

It can't actually be a  `Function` as you can't pass it the output-container (as a `View` is calculated on the fly). But true, it  can be `Threadable` (but not a `Converter`).

> If you are going to change the 2 convention to To then we also need to rename gauss/GaussRAI2RAI, project/ProjectRAI2II and slicer/SlicewiseRAI2RAI. And document this convention once and for all on the Naming page of the wiki.

Working with Ops, the 2 looks more and more stupid. But you are right, this shouldn't be part of this commit and be part of a separate PR.

dietzc:
As soon as https://github.com/imagej/imagej-ops/pull/130 is merged, we will update this branch.

dietzc:
cleanuped up (thanks to https://github.com/imagej/imagej-ops/pull/130). Ready to merge? Nothing crucial happened here, just more tests.

dietzc:
@hinerm what do you think?

hinerm:
thanks!

PR #130, opened 2015-05-22
Introducing easier readable naming schema

dietzc:
While programming with ops, we figured, that the current way of method
naming causes a. confusion as it wasn't easily understandable
and b. was pretty inconsistent. Therefore we used the following naming
schema in addition to whats already written down in the WIKI:

- RandomAccessibleInterval -> Rai as the only abbreviation.
- Mappings are written with "To" instead of "2".
- OpXYInterface -> AbstractOpXY -> DefaultOpXY -> OpXYSpecialFunction

For me this somehow looks nicer and is easier to read. However, I'm happy to discuss a better naming schema with all of you. As soon as this PR is merged, I will update https://github.com/imagej/imagej-ops/wiki/Naming.

### DIFF
--- a/src/main/java/net/imagej/ops/chunker/ChunkerInterleaved.java
+++ b/src/main/java/net/imagej/ops/chunker/ChunkerInterleaved.java
@@ -28,47 +28,84 @@
  * #L%
  */
 
-package net.imagej.ops.misc;
+package net.imagej.ops.chunker;
 
-import java.util.Iterator;
+import java.util.ArrayList;
+import java.util.concurrent.Future;
 
 import net.imagej.ops.Op;
 import net.imagej.ops.Ops;
-import net.imglib2.type.numeric.RealType;
 
-import org.scijava.ItemIO;
+import org.scijava.Priority;
+import org.scijava.log.LogService;
 import org.scijava.plugin.Parameter;
 import org.scijava.plugin.Plugin;
 
 /**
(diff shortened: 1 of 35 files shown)


ctrueden:
Hmm, it may be that the case issues on OS X are fixed in newer versions of Git. At least, I was able to check out this branch without everything being hosed. But I'd still like to keep the acronym as RAI instead of Rai. As things stand, there are four classes with `RAI` in the name that this branch does not downcase:
```
$ find . -name '*RAI*'
./src/main/java/net/imagej/ops/crop/AbstractCropRAI.java
./src/main/java/net/imagej/ops/fft/filter/AbstractFFTFilterRAI.java
./src/main/java/net/imagej/ops/fft/filter/IterativeFFTFilterRAI.java
./src/main/java/net/imagej/ops/fft/filter/LinearFFTFilterRAI.java
```

ctrueden:
@dietzc I pushed a fixup that upcases all the uses of `Rai` to `RAI` instead. If you are OK with this, I will squash it into your commit, then merge. Or if not, we can discuss further—maybe take a vote!

dietzc:
sorry, had to finish my new kitchen yesterday. :) RAI is fine for me. Thanks!

hinerm:
thanks @dietzc !

ctrueden:
Squashed and merged. Thank you very much.

PR #131, opened 2015-05-26
Real binary

hinerm:
@ctrueden @apal4 is this branch still active or was it superceded by `real-binary-math`


ctrueden:
@LeonYang5114 finished a modern version of most/all of this.

PR #132, opened 2015-05-26
Ops migration

hinerm:
@ctrueden for review..


ctrueden:
* I/O functions cannot be ops because they are not deterministic (i.e., they have externalities).
* ImageCalculator is a high-level command rather than an op.
* Find Edges will make it into ops eventually, once we have the prerequisite building blocks.
* Conversion of numeric types will be taken care of as needed using `Converter`s.

PR #133, opened 2015-05-26
Binary

hinerm:
@ctrueden for review


ctrueden:
I think this was merged in a different form already.

PR #134, opened 2015-05-26
Binary boolean

hinerm:
@ctrueden for review


ctrueden:
It will be easier to redo this work on top of the latest `BinaryOp` API.

PR #140, opened 2015-06-09
Add Computer Interface

dietzc:
A `Computer` computes an output given some input. We don't have an interface yet, which reflects this reoccurring pattern. Furthermore is each `OutputFunction` a `Computer`.

Maybe the name is not the best one, but the interface is useful.


ctrueden:
Perfect!

dietzc:
yeah! easy one :)

PR #143, opened 2015-06-13
Add OpResolverService

dietzc:
Hi all,

this PR introduces the `OpResolverService`, a default implementation of this service and a few unit-tests. The `OpResolverService` can be used to resolve `Op`s which have dependencies to other `Op`s. This work is part of the work we are doing in https://github.com/imagej/imagej-ops/tree/outputop-service. However, I decided to split this branch into several smaller PRs such that its more reviewable!

Thanks for feedback in advance!

### DIFF
--- a/src/main/java/net/imagej/ops/resolver/DefaultOpResolverService.java
+++ b/src/main/java/net/imagej/ops/resolver/DefaultOpResolverService.java
@@ -1,0 +1,607 @@
+/*
+ * #%L
+ * ImageJ software for multidimensional image processing and analysis.
+ * %%
+ * Copyright (C) 2014 - 2015 Board of Regents of the University of
+ * Wisconsin-Madison, University of Konstanz and Brian Northan.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * 
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ * #L%
+ */
+
+package net.imagej.ops.resolver;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+
+import net.imagej.ops.Computer;
+import net.imagej.ops.InputOp;
+import net.imagej.ops.Op;
+import net.imagej.ops.OpCandidate;
+import net.imagej.ops.OpMatchingService;
+import net.imagej.ops.OpRef;
+import net.imagej.ops.OpService;
+import net.imagej.ops.OutputOp;
+
+import org.scijava.convert.ConversionRequest;
+import org.scijava.convert.ConvertService;
+import org.scijava.module.MethodCallException;
+import org.scijava.module.Module;
+import org.scijava.module.ModuleException;
+import org.scijava.module.ModuleInfo;
+import org.scijava.module.ModuleItem;
+import org.scijava.plugin.Parameter;
+import org.scijava.plugin.Plugin;
+import org.scijava.service.AbstractService;
+import org.scijava.service.Service;
+
+/**
+ * Default implementation of the {@link OpResolverService}. {@link Op}s are
+ * resolved using a heuristic, which simply always takes the locally best
+ * {@link Op}, which means the {@link Op} with the highest priority. NB: A
+ * smarter implementation of the {@link OpResolverService} could find globally
+ * optimal solutions.
+ * 
+ * @author Christian Dietz, University of Konstanz
+ */
+@Plugin(type = OpResolverService.class)
+public class DefaultOpResolverService extends AbstractService implements
+	OpResolverService
+{
+
+	@Parameter
+	private OpService ops;
+
+	@Parameter
+	private ConvertService cs;
+
+	@Parameter
+	private OpMatchingService matcher;
+
+	@Override
+	public <I> ResolvedOpSet<I> resolve(I input, OpRef<?>... refs) {
+		Set<OpRef<?>> pool = new HashSet<OpRef<?>>();
+		pool.addAll(Arrays.asList(refs));
+		return resolve(input, pool);
+	}
+
+	@SuppressWarnings({ "unchecked", "rawtypes" })
+	@Override
+	public <I, OP extends Op> Computer<I, OP> resolve(I input, Class<OP> type,
+		Object... args)
+	{
+		return resolve(input, new OpRef(type, args));
+	}
+
+	@SuppressWarnings("rawtypes")
+	@Override
+	public <I, O> Computer<I, O> resolve(Class<O> outType, I input,
+		OpRef<? extends OutputOp> ref)
+	{
+		final Computer<I, ? extends OutputOp> resolved = resolve(input, ref);
+		return new Computer<I, O>() {
+
+			@SuppressWarnings("unchecked")
+			@Override
+			public O compute(I input) {
+				return ((OutputOp<O>) resolved.compute(input)).getOutput();
+			}
+		};
+	}
+
+	@Override
+	public <I, OP extends Op> Computer<I, OP> resolve(final I input,
+		final OpRef<OP> opRef)
+	{
+		final HashSet<OpRef<?>> set = new HashSet<OpRef<?>>();
+		set.add(opRef);
+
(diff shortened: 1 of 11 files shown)


ctrueden:
Right now, the paradigm is to have one marker interface per org, basically. So we have `SciJavaService` for SciJava-level services, `ImageJService` for ImageJ-level services, and `SCIFIOService` for SCIFIO-level services. We could certainly introduce more layers of markers but I would be reluctant. Note that `OpService` itself cannot be used, as it is the interface for implementations of the `OpService` contract specifically. But IIUC we could have an `ImageJOpsService` marker—would be totally confusing, but we could do it... but again, I'd rather not unless we really need it. As things stand, you can write: `new Context(OpResolverService.class)` and you're fine, since it brings in its needed dependencies.

ctrueden:
Just a heads up: I talked with @hinerm on Friday and he mentioned that he has some thoughts and questions on how best to achieve the requirements this service is attempting to meet. So expect some more significant comments early next week.

hinerm:
@dietzc OK, I've looked over this PR and if I understand correctly:

* To use the `OpResolverService` a user would have to get a handle on it themselves
* The scope of the ops in the op pool is limited to that call of `resolve`.. so for a user to take advantage of your "mean, sum, average" example they would call something resembling `resolverService.resolve(image, MeanOp, SumOp, AverageOp)` - is that right?

If these assumptions are correct I'm worried about putting the responsibility for both anticipating caching needs and knowing the set of related ops on the user - as it would add a layer of complexity which would not lend itself well to trying to use ops in a lightweight context, such as in an interpreter.

I am wondering if we could instead make an Op caching framework that works something like this:

* By default, Ops cache their output, keyed on op type + input identifier in a central caching service (we can create an `OpCacheService` since we don't yet have the `LRUCacheService` and eventually migrate..)
* Ops declare their dependencies on other ops via `@Parameters`
* When populating Op `@Parameters` we use a new `OnPreprocessor` that a) ensures any non-Op `@Parameters` are populated first, and then does lookups based on op type + input identifier to see if there are cached outputs to reuse.

This removes the burden from users for a priori knowledge of what Ops they may want to cache.

Anyway if it seems like I totally missed the intent of this work please let me know...

dietzc:
> To use the OpResolverService a user would have to get a handle on it themselves 
correct. potentially we could add a shortcut through `OpService`.

> The scope of the ops in the op pool is limited to that call of resolve.. so for a user to take advantage of your "mean, sum, average" example they would call something resembling resolverService.resolve(image, MeanOp, SumOp, AverageOp) - is that right?

No. The `OpResolverService` will try to find the suitable `SumOp` and `AverageOp` given the `MeanOp` and the image as the input. Caching then takes place in the `CachedModules`. True, we could add a dependency there on a LRUCache (which doesn't exist, yet).

Basically, the rest you describe is what we do in the `OpResolverService`:-)

hinerm:
OK I'm confused.. if I make the following calls:

```
ops.op(SumOp, image1);
ops.op(AverageOp, image1);
ops.op(MeanOp, image1);
```

The call to `MeanOp` is going to recalculate `SumOp` and `AverageOp` right?

And there is nothing in `DefaultOpService` that uses the `OpResolverService`.. so if I want to reuse my `Sum` and `Average` results I need a different call:

```
OpResolverService.resolve(image1, SumOp, AverageOp, MeanOp);
```

Is that correct or am I misusing the API?

dietzc:
If you want to reuse `SumOp` and `MeanOp` then you have to pass them, right. However, we introduced `FeatureSet`s in another branch which will provide groups of features https://github.com/imagej/imagej-ops/blob/outputop-service/src/main/java/net/imagej/ops/features/FeatureSet.java and `Feature`s themselves. We will open another PR as soon as this one is merged for all the feature-calculation related classes.

Anyway, if you only want to calculate `MeanOp` the `OpResolverService` will find `SumOp` and `AreaOp` and re-use it internally, if needed for other `Op`s. The returned `ResolvedOp` will then only be able to calculate the `MeanOp` given the input.

dietzc:
> And there is nothing in DefaultOpService that uses the OpResolverService.. so if I want to reuse my > Sum and Average results I need a different call:
> OpResolverService.resolve(image1, SumOp, AverageOp, MeanOp);

it really depends what you want to do. If you want to get the results from all these `Op`s back, you have to remember them and then call `getOutput` after running them. However, if you only want the result of the `MeanOp` then you are all set if you pass the `MeanOp` to the `OpResolverService`. The `OpService` (as it is right now) has no shortcuts yet to the `OpResolverService`.

hinerm:
>However, if you only want the result of the MeanOp then you are all set if you pass the MeanOp to the OpResolverService

So my concern with the `OpResolverService` is that it feels like there are now two ways to run ops - `OpService` or `OpResolverService`. Also it's not clear to me how:
```
OpResolverService.resolve(MeanOp)
```
is different from
```
ops.op(MeanOp)
```


>OpResolverService will find SumOp and AreaOp and re-use it internally, if needed for other Ops

My impression was that the `OpResolverService` can only re-use op results that are all passed as part of the pool to a `resolve` call. Is that correct?

dietzc:
The `OpResolverService` will take care about resolving `Op`s which are required by the provided `Op`. The `OpService` will not do that. However, I agree, in an ideal world, the functionality of the `OpResolverService` would be part of the `OpService` itself. I just don't see that happening in the near future.

 > My impression was that the OpResolverService can only re-use op results that are all passed as part of the pool to a resolve call. Is that correct?

Nope. Lets assume we want to resolve two `Op`s. A `MeanOp` and a `XYZOp` which both use the `SumOp` and you only pass `MeanOp` and  `XYZOp` to the `OpResolverService`. It will discovery/resolve all `Op`s required by the passed `Op`s and re-use/cache them internally. In this case `SumOp` and `AreaOp`. However, if you want to access the results of these two automatically resolved `Op`s you would have to pass them.

hinerm:
>Lets assume we want to resolve two Ops. A MeanOp and a XYZOp which both use the SumOp and you only pass MeanOp and XYZOp to the OpResolverService

This statement only applies if you do `OpResolverService.resolve(image, MeanOp, XYZOp)` right? If I run:

```
OpResolverService.resolve(image, MeanOp);
OpResolverService.resolve(image, XYZOp);
```
is the `SumOp` reused? I think it isn't, because [the modulePool is rebuilt](https://github.com/imagej/imagej-ops/blob/opresolver/src/main/java/net/imagej/ops/resolver/DefaultOpResolverService.java#L142-143) in resolve..

dietzc:
Ah ok. It's not reused if you call it twice. It's reused if you would call `OpResolverService.resolve(image, MeanOp, XYZOp)`

hinerm:
>It's not reused if you call it twice

Exactly. In my opinion the purpose of Ops is to be a lightweight, intuitive, easy-to-use framework that allows users to rapidly do awesome image processing.

Adding a second entry point for running ops, and requiring that users know the inner workings of the ops (I have to know that `XYZOp` and `MeanOp` both use `SumOp` to get any benefit from the resolver) is adding a significant level of complexity that I think most users will not be able to take advantage of.

I want to be clear that I think the ***intent*** of the `OpResolverService` - to avoid unnecessary recalculations of ops - is fantastic and something we really want. I just think it would have much greater impact to try and work this behavior into the `OpService` itself.

Anyway I am willing to put some time into this.. I will try to put together a concrete example of what I think would work here.

dietzc:
>Exactly. In my opinion the purpose of Ops is to be a lightweight, intuitive, easy-to-use framework that allows users to rapidly do awesome image processing.

Right, this is how we defined it and thats why we added the `Feature` level in the other branch, to hide this complexity. I would rather see a call like `ops.measure(image, firstorderstatistics.class)` which internally uses the `OpsResolverService`.

> Adding a second entry point for running ops, and requiring that users know the inner workings of the ops (I have to know that XYZOp and MeanOp both use SumOp to get any benefit from the resolver) is adding a significant level of complexity that I think most users will not be able to take advantage of.

We only add this complexity on `OpResolverService` level, which then can be called from other levels (e.g. the `OpService` itself). Users should not be required to call `OpResolverService` themselves (as they will never ever use the `OpMatcher` for example, too). 

> I want to be clear that I think the intent of the OpResolverService - to avoid unnecessary recalculations of ops - is fantastic and something we really want. I just think it would have much greater impact to try and work this behavior into the OpService itself.

I think the only option we might have to achieve this general approach is to wrap ops generally into a `CachedOp` or so, which always checks if it has been already calculated and the result is still available in an `LRUCache` or so (which will actually fill up the cache really fast if you do a lot of calculations ...). I wanted to avoid this general mechanism as it puts overhead to operation e.g. calculated on pixel-level. However, if we find a nice way to avoid this overhead and additionally find ways for fast checks if an `Op` has already been calculated given some Input or not, I'm in and happy to discuss all the options.

dietzc:
by the way: if we find such a more general (centrally organized) approach, it would also potentially eliminate some of the drawbacks of the current approach, which can only resolve `Op`s for a single input. One problem might be, that a global optimization of which `Op`s we might want to use in combination might not be that easy anymore (as we have a DAG at the moment, we could do this in the current representation). However, this is not implemented yet, anyway.

dietzc:
another thing which came to my mind: we could enforce operations to be define on their own if they are `Cacheable` or not. Either they implement an interface or not and if they implement an interface they can still decide if its worth caching the output, given some input.

hinerm:
@dietzc - I was originally thinking that we would want an `InputPreprocessor` that attempts to deduce parameters for `Op @Parameters`. @ctrueden and I talked more yesterday and came to the conclusion that since `Op`s are supposed to be deterministic, caching a mapping of input parameters to `Op` results feels quite natural and should just be the default behavior of running `Op`s. Then we have one central location to control cache behavior.

As far as per-pixel complexity goes, Curtis pointed out that per-pixel operations are already terrible and we generally work to avoid mapping operations when possible. If you are running an `Op` per pixel there is already going to be a lot of overhead so adding a cache lookup seems minimal (compared to running a command). So in the name of avoiding premature optimization we would just implement the general caching solution, and then fix per-pixel performance if it becomes a problem (i.e. just ignore the cache if we're doing something per-pixel).

Also, if we just cache by default there is no need to try and construct a DAG, I think, because `Op`s just call the other `Op`s they need and we let the framework worry about if that `Op` has been called with the given set of parameters before or not.

Is there any functionality of the `OpResolverService` that would not be satisfied if we cached results by default?

@ctrueden - did I forget anything? :octopus:

dietzc:
ok. makes sense. This implies, that I would add a `if my op I depend on is null, ask the OpService to create one` or to be more precise: Each `Op` has to take care about resolving the `Op`s on which they depend on their own using the `OpService`? In the case of the `Mean` we would add null checks on the other `Op`s?

If so, I fear the the statement below is not 100% correct.

> Also, if we just cache by default there is no need to try and construct a DAG, I think, because Ops just call the other Ops they need and we let the framework worry about if that Op has been called with the given set of parameters before or not.

Of course, at the moment we don't do the global optimization. We currently rather use a heuristic based on the priorities of the `Op`s.  If we now let the `Op`s locally ask for dependend `Op`s, we will not be able to find the globally most suitable `Op`. Imagine you have several interface names of `Op`s you want to calculate and you have some inputs. In an ideal world, we would choose the set of `Op`s which is the best given the input. For example avoid unnecessary (or expensive) conversions/calculations between types. Again, more precise: from `LabelRegion<?>` to `Polygon`. If we are only interested in e.g. the `ConvexHull` of this `LabelRegion` we don't have to dervice a `Polygon` first, rather we would directly calculate the result on the `LabelRegion`. If we would have 25 other, high-priority ops we want to calculate and all of them require a `Polygon` input, then it might make sense to convert it to a `Polygon` first. Given priorities of `Op`s and the dependencies between them, we could already find a nicer heuristic than just a local call to the `OpService`.

Does this make sense? Do you see a chance that such an approach is still possible in the future?

dietzc:
@ctrueden @hinerm we could use the `CacheService` not only to cache the results, but also to cache instances of `Op` themselves given some parameters to avoid redundant `Op` matching / discovery which is actually pretty expensive.

PR #147, opened 2015-06-23
Add new Create Op to create Imgs from Imgs

dietzc:
This `Op` creates an `Img` given the information from another `Img`.

### DIFF
--- a/src/main/java/net/imagej/ops/create/CreateImgFromImg.java
+++ b/src/main/java/net/imagej/ops/create/CreateImgFromImg.java
@@ -1,0 +1,86 @@
+/*
+ * #%L
+ * ImageJ software for multidimensional image processing and analysis.
+ * %%
+ * Copyright (C) 2014 - 2015 Board of Regents of the University of
+ * Wisconsin-Madison, University of Konstanz and Brian Northan.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * 
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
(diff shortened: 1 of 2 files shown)


hinerm:
I am confused.. `Img` has a [`copy`](https://github.com/imglib/imglib2/blob/imglib2-2.3.0/src/main/java/net/imglib2/img/Img.java#L70) method... why is that not sufficient..?

dietzc:
`copy()` copies the `Img` (including the content). Creating a new empty `Img` is something else.

hinerm:
ah, I saw "given the information" and assumed it was copying data too.

OK, so the set of Ops that performed this behavior before was deleted in https://github.com/imagej/imagej-ops/commit/bf3dafe6e6010525a9ef6573c025ff93996dabbf. Was that removal a mistake?

More to the point, why do we need a 1-liner op that delegates to a different op?

dietzc:
Deleting the previous `Op` was not a mistake as this `Op` was using an entirely different infrastructure. We just didn't re-add this short-cut again. 

> More to the point, why do we need a 1-liner op that delegates to a different op?

Because it's more convenient than writing `ops.create(img, img.firstElement().createVariable(), img.factory())`.

hinerm:
>ops.create(img, img.firstElement().createVariable(), img.factory())

So without this `Op`, you could call `ops.create(img)` and right now it looks like the limitation is that it will give you a `DoubleType` `Img` back because it always forces the resulting `Img` to match the type of `CreateNativeType`.

Philosophically, my take is that optional parameters are used in [DefaultCreateImg](https://github.com/imagej/imagej-ops/blob/imagej-ops-0.14.0/src/main/java/net/imagej/ops/create/DefaultCreateImg.java#L71-L75) so that a single `Op` can encapsulate the functionality of the various `Img` creation method signatures and pathways.

So instead of adding a delegating op for this one use case, why not add the logic to `DefaultCreateImg` so that if its input is typed, that type can be reused instead of falling back to the default `NativeType`?

Sorry if you and @ctrueden already discussed guidelines for when to use optional params or conditional logic and when to create helper ops... it's not completely clear to me which option is more in line with the ops goals.

hinerm:
Another possibility would be to give the `CreateNativeType` an optional parameter.. if that param is something with a type, it returns that type, and if not returns `DoubleType`..

ctrueden:
I think that optional args vs. multiple ops must be weighed case by case. Certainly there are limitations on which situations can properly cover optional args. I would also not use optional args for mutually exclusive params—e.g., if you can say either `foo(a, c)` _or_ `foo(a, b)` but writing `foo(a, b, c)` would make no sense, then I would rather have `FooA` and `FooB` ops than a single `Foo` with `b` and `c` being optional.

In this case, I think having a single op that takes an optional output type argument is totally fair, and likely simpler than two op classes.

@hinerm One quick point about your earlier question regarding why `copy()` might not be sufficient: even if there is a single method call that does a thing to an object, we still might want to wrap it in an op so that that op is available as a function in contexts such as the op expression evaluator. As long as the `copy()` method is deterministic, it is potentially handy to wrap it as an op. We are planning to wrap lots of lower level ImgLib2 functionality in this way too (e.g., `Views`), for similar reasons.

dietzc:
> In this case, I think having a single op that takes optional output type and factory arguments is totally fair, and likely simpler than two op classes. Would you agree, @dietzc?

I don't understand what you are suggesting :-) We have an `Op` with two optional input parameters, however I still need the second `Op` which makes it easier to create `img`s from `Img`s?

One comment about case-logic and several `Op`s: I have also not really an idea what is better when and I really think we have to judge from case to case. However, case-logic is not easily extensible, especially in this case as we don't have an interface we could use (like `IsTyped`) and we would have to cover several cases (e.g. `IterableInterval`, `RandomAccessibleInterval`) and we don't know which implementations of `Dimension` will hold a `Type` in the future.

ctrueden:
> I don't understand what you are suggesting :-) We have an `Op` with two optional input parameters, however I still need the second `Op` which makes it easier to create `img`s from `Img`s?

Right, sorry, I need to read the code more carefully next time before spouting off.

So, OK, I think what this is doing is good! The only concern I have is the unsafe cast to `Img<T>`—the `CreateImg` implementation chosen might not produce anything of that type. In practice, you are implicitly relying on the fact that `DefaultCreateImg` will be chosen; maybe it would be better to just invoke that implementation directly using `DefaultCreateImg.class`. Or even skipping the ops framework altogether and calling a plain old Java method instead of `ops.run`.

dietzc:
@ctrueden what about typing CreateImg as an OutputOp>?

ctrueden:
@dietzc Even then someone could implement `CreateImg` but define a second output parameter and then the `ops.run` call would give back a `List` instead of an `Img<T>`.

dietzc:
Ok. The `Op` now uses `DefaultCreateImg`.

ctrueden:
Good enough for government work! :+1:

dietzc:
:+1:

Issue #103, opened 2015-01-26
Op matching should avoid unnecessary conversion


ctrueden:
Suppose we have these ops:
* `foo(Number, Number)`
* `foo(Double, Double)`
* `foo(Long, Double)`
* `foo(Double, Long)`
* `foo(Long, Long)`

At the moment, attempting to call `foo(5L, 5L)` will cause an error because all five are matched. If we work around this limitation using prioritization, some subset of the ops will become always inaccessible—e.g., either `foo(Long, Double)` or `foo(Double, Long)` no matter what priority order is used.

In terms of non-primitive objects, There are really two types of "conversion":
* Casting, which is lossless
* Conversion via the `ConvertService`, which might be lossy

It would be nice of the op matcher preferred matches in the following order:
1. Matches that require only object casting.
2. Matches that require only lossless conversion.
3. Matches that require lossy conversion.

The ops framework could be improved right now to separate case 1 from 2+3 above. Separating cases 2 and 3 from other would require improvements to the `ConvertService` so that converters can report whether the conversion is lossy.

Regarding primitives, it should "just work" since all primitives are boxed, and the type hierarchy for the boxed types behaves as we desire (e.g., `Integer` objects cannot be cast to `Double`, even though at the primitive level Java lets you do it without writing an explicit cast because it is lossless).

ctrueden:
Big progress with the merge of #389. Need to validate that the above scenario now works as desired before closing this issue.

ctrueden:
Validation is overrated. Will reopen if any problems crop up.

PR #148, opened 2015-06-23
Copy Ops 

dietzc:
DO NOT MERGE ... but explain me why the copy tests fail?

* `ImgLabelingTest`: I don't understand why having two `Op`s with the same priorities but different input parameters are a problem.

* I don't understand the problem with `CopyTypeTest` at all.

Can you maybe help @hinerm @ctrueden?

### DIFF
--- a/src/main/java/net/imagej/ops/AbstractOpEnvironment.java
+++ b/src/main/java/net/imagej/ops/AbstractOpEnvironment.java
@@ -37,7 +37,15 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 
+import org.scijava.AbstractContextual;
+import org.scijava.command.CommandInfo;
+import org.scijava.command.CommandService;
+import org.scijava.module.Module;
+import org.scijava.module.ModuleItem;
+import org.scijava.plugin.Parameter;
+
 import net.imagej.ops.convert.ConvertPix;
+import net.imagej.ops.copy.CopyNamespace;
 import net.imagej.ops.create.CreateNamespace;
 import net.imagej.ops.deconvolve.DeconvolveNamespace;
 import net.imagej.ops.features.haralick.HaralickNamespace;
@@ -57,13 +65,6 @@ import net.imglib2.RandomAccessibleInterval;
 import net.imglib2.algorithm.neighborhood.Shape;
 import net.imglib2.type.Type;
 import net.imglib2.type.numeric.RealType;
-
-import org.scijava.AbstractContextual;
-import org.scijava.command.CommandInfo;
-import org.scijava.command.CommandService;
-import org.scijava.module.Module;
-import org.scijava.module.ModuleItem;
-import org.scijava.plugin.Parameter;
 
 /**
  * Abstract superclass for {@link OpEnvironment} implementations.
@@ -637,6 +638,11 @@ public abstract class AbstractOpEnvironment extends AbstractContextual
 	}
 
 	// -- Operation shortcuts - other namespaces --
+	
+	@Override
+	public CopyNamespace copy() {
+		return namespace(CopyNamespace.class);
+	}
 
 	@Override
 	public CreateNamespace create() {
--- a/src/main/java/net/imagej/ops/OpEnvironment.java
+++ b/src/main/java/net/imagej/ops/OpEnvironment.java
@@ -34,7 +34,12 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 
+import org.scijava.Contextual;
+import org.scijava.command.CommandInfo;
+import org.scijava.module.Module;
+
 import net.imagej.ops.convert.ConvertPix;
+import net.imagej.ops.copy.CopyNamespace;
 import net.imagej.ops.create.CreateNamespace;
 import net.imagej.ops.deconvolve.DeconvolveNamespace;
 import net.imagej.ops.features.haralick.HaralickNamespace;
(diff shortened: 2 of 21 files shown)


hinerm:
>Can you maybe help @hinerm @ctrueden?

@dietzc definitely! I'll look at the tests right now..

hinerm:
>ImgLabelingTest

In the `CopyImgLabeling` op, the `conforms` method just tests whether the parameters are of [equal size and class](https://github.com/imagej/imagej-ops/blob/copy-ops/src/main/java/net/imagej/ops/copy/CopyImgLabeling.java#L79-L87) - not whether they are actually `ImgLabeling`s. So it claims to be able to support copying two equal sized `ArrayImg`s

hinerm:
>I don't understand the problem with CopyTypeTest at all.

Exception is:

```
java.lang.IllegalAccessException: Class org.scijava.plugin.PluginInfo can not access a member of class net.imagej.ops.copy.CopyType with modifier
```

Look at the class.. it's `package` visibility. It needs to be `public`.

Also many of these plugin annotations don't conform to the `(type=Op.class, name=Ops.xyz.NAME)` standard. I updated `CopyType`.

dietzc:
Thanks @hinerm. I overlooked the visibility of `CopyType`. Thanks for pointing me on this. However, I still don't understan the `ImgLabelingCopy` problem. My understanding of `Contigent` is that it was not introduced to match parameter types, rather than properties of the parameters. The problem in this case is, that the generic type matching of the matcher doesn't work. @ctrueden I know there were some issues with generic - type matching. Is this one of the situations where the matcher can't perform the actual matching?

hinerm:
>However, I still don't understan the ImgLabelingCopy problem

@dietzc I dug into this more today. Looks like the actual reason both `Op` candidates are matching is that the [`DefaultConverter`](https://github.com/scijava/scijava-common/blob/scijava-common-2.43.0/src/main/java/org/scijava/convert/DefaultConverter.java#L176-L180) sees that `ImgLabeling` has a constructor that [accepts a RAI](https://github.com/imglib/imglib2-roi/blob/imglib2-roi-0.3.1/src/main/java/net/imglib2/roi/labeling/ImgLabeling.java#L85-L93) - which can therefore accept an `ArrayImg`.

So since `ArrayImgs` can be converted to `ImgLabelings` you must differentiate the `CopyImgLabeling` and `CopyArrayImg` `Op`s by priority. `CopyArrayImg` should be higher priority since it is more specific than `CopyImgLabeling`.

dietzc:
I understand. I will fix this. Thanks! However, something is still not correct and I suspect this has a lot to do with the generic types matching. `ArrayImg`s actually can't be converted to `ImgLabeling`s in general, as only `RandomAccessibleInterval` of type `IntegerType` can be consumed by the constructor.

hinerm:
>only RandomAccessibleInterval of type IntegerType can be consumed by the constructor

Well.. this test is using an `IntegerType` so it's reasonable. But actually the problem isn't with `CopyImgLabeling` matching, it's `CopyArrayImg` matching right?? So I think my suggestion is wrong...

hinerm:
>So I think my suggestion is wrong... 
Nope, just kidding. I didn't realize that `CopyImgLabeling` calls copy on the underlying `Img`s during its `safeCompute` method.

I do not think this is a problem with generic matching because if you have an `ImgLabeling` that you want to copy, then you necessarily have an `RAI<? extends IntegerType>` as the index img of the labeling. So you can always construct an `ImgLabeling` from this `RAI`, whether or not the `RAI` is also an `ArrayImg` or not (as it happens to be in this case).

The point is, you ***must*** have a higher priority `Copy Op` that can convert between `RAI`s, or the `CopyImgLabeling` will recurse infinitely.

ctrueden:
Note that issue #103 is relevant here: we would have fewer op clashes (and no clash here in particular) if the matcher preferred ops which do not require calls to the `ConvertService` to fulfill.

dietzc:
not ready for merge, yet. however, getting close.

dietzc:
@tibuch can you also take care about this? I think only tests and a few more specialized copy ops are missing in a first step.

dietzc:
I forced push several changes and `FIXME`s. which we should discuss at the hackathon (general questions). Afterwards, we (@tibuch) only need(s) to add more tests and we can merge.

* [ ] Add tests.

dietzc:
@ctrueden  @hinerm if you could do a quick review... I think we can merge then.

hinerm:
- [ ] @dietzc my understanding is that namespaces should have a [varargs](https://github.com/imagej/imagej-ops/blob/imagej-ops-0.20.0/src/main/java/net/imagej/ops/math/MathNamespace.java#L63) signature for each built-in method.. @ctrueden maybe that should be covered in the signature test..?

- [ ] Tests don't have licenses. Need a `mvn license:update-file-header` pass.

- [ ] Style and whitespace stuff. Could just use a [style cleanup](https://github.com/imagej/imagej/tree/imagej-2.0.0-rc-39/config) pass.

- [ ] Needs manual reconciliation with `master`. Just need to union the `OpService` and `DefaultOpService` changes.

dietzc:
>  @dietzc my understanding is that namespaces should have a varargs signature for each built-in method.. @ctrueden maybe that should be covered in the signature test..?

I wouldn't do a varargs for every `Copy` implementation and as the `Copy` interface is not an `Op`, we can't do it for it.

dietzc:
I forgot to push the changes I did some days ago ;-) No it should be reviewable!

dietzc:
@tibuch did you force push? My changes are gone...

dietzc:
closed in favour of https://github.com/imagej/imagej-ops/pull/235

PR #150, opened 2015-06-25
Allow use of Regions.sample(region, src)

jaywarrick:
Use of dimension(0) and dimension(1) resulted in negative width and
height when using Regions.sample(region,src) as the dimensions referred
to the src dimensions while the min referred to the region. Use of
max(0) and max(1) results in proper calculation using region only
information. Had to add +1 to make calculation accurate and pass tests.


jaywarrick:
Hope this is the right way to do things. Let me know if I'm way off. Also my commit message might not be up to par. Still learning.

PR #151, opened 2015-06-25
Adjusted plugin information for several create-ops to avoid Op clashes

dietzc:
* names of plugins fixed
* adjusted priorities to avoid op clashes


dietzc:
seems to fixed during cleanup of @ctrueden

PR #155, opened 2015-07-01
Update input, avoid extra calc, and copy type

jaywarrick:
1) if(op == null)... sets the input once and only once so that subsequent calls to run this feature set use the same input (e.g., LabelRegion) even though it is being called with .getFeatureList(Regions.sample(region, image)) where region is a new region each time. This very likely has alternative side effects I'm not aware of so please advise otherwise. Maybe a new HistogramFeatureSet op should be created for each region, which would also likely fix the issue? Potentially, if there were a simple way to just "set" the input variable for each new call instead of creating a new op instance every time, that might work but I couldn't see how to do so.

2) Remove extra call to countData(getInput()) as this is already done with the op.run() command. See HistogramCreate class line 75.

3) Copy the "type" variable when saving into "res" otherwise, all the Pair's in "res" refer to the same instance, resulting in a histogram that contains the same value for all bins.

### DIFF
--- a/src/main/java/net/imagej/ops/features/sets/HistogramFeatureSet.java
+++ b/src/main/java/net/imagej/ops/features/sets/HistogramFeatureSet.java
@@ -48,14 +48,11 @@ public class HistogramFeatureSet<T extends RealType<T>> extends
 	@Override
 	public void run() {
 
-		if (op == null) {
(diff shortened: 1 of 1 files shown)


dietzc:
I'm working on  a way nicer API for this at the moment. I will integrate your changes there. Thanks alot!

PR #156, opened 2015-07-01
Refactoring of `Gauss` implementations

dietzc:
In this first step of refactoring the default implementation `Gauss` we redefine which parameters are used. Especially, we allow the sigmas to vary over the dimensions. However, this is only a first step, as we should use convolve `Op`s as soon as `SeparableSymmetricConvolution` of ImgLib2 is integrated.


dietzc:
@ctrueden I updated the implementations and forced-push to this branch such that it can be merged to the current master.

ctrueden:
Rock on, thanks!

PR #157, opened 2015-07-01
Add CCA

dietzc:
This PR adds a wrapper for the `ConnectedComponents` implementation of imglib2-algorithms (see https://github.com/imglib/imglib2-algorithm/blob/master/src/main/java/net/imglib2/algorithm/labeling/ConnectedComponents.java).


ctrueden:
These commits are mutually interdependent. I force pushed a rewrite that fixes the issue.

ctrueden:
Thanks @dietzc!

PR #158, opened 2015-07-01
LabelingNamespace: call "run" instead of methods in the LabelingNamespace

dietzc:
I forgot about that @ctrueden. sorry! Branch was ready to go before our Namespace discussions.
However this PR makes sure that inside namespace methods only "run" is called.


ctrueden:
Oh no, I just pushed this same thing! Sorry Christian!

dietzc:
thanks for fixing this @ctrueden.  :-)

ctrueden:
See d71e11d439d1c78f9c735d28cdd283d44a5fb92b. It also swaps the order of the methods to be in "descending" order.

ctrueden:
Now I just have to fix the [failing build](http://jenkins.imagej.net/job/ImageJ-OPS/282/console) (argh Java 6...).

PR #159, opened 2015-07-02
Geometricops

DanielSeebacher:
I've added some simple two-dimensional geometric objects. All objects were implemented in such a way, so that other existing algorithms can use them. A polygon for example is a ```RealRandomAccessibleRealInterval<BoolType>``` or a ```RasterizedPolygon``` is a ```IterableRegion<BoolType>```.

There are also some simple ops, for example to get the outline of a binary image (see ```DefaultMooreOutlineOp```) or the get the convex hull of a polygon (see ```DefaultConvexHull2DOp).


dietzc:
@tpietzsch do you think some of this work should migrate into imglib2-ops?

@DanielSeebacher are you sure that a `Polytop` is define like that? Aren't you rather implementing a `Polygon` here? you also mix the naming in the class comments / class itself.

dietzc:
@DanielSeebacher thank you btw. this looks like a great starting point for further discussion. Are there also Unit-Tests? Could you add them? Then we also understand the use-cases better.   Where do you see potential performance problems or any other API problems in ImgLib2-roi?

DanielSeebacher:
Closing the Pull Request for now. I think it is better to include the geometric ROIs in imglib2-roi. If they're added there I only have to implement the ops in imagej-ops.

dietzc:
See https://github.com/imglib/imglib2-roi/pull/13

PR #162, opened 2015-07-09
Add difference of gaussian implementations

dietzc:
this PR adds a default implemenation of `DifferenceOfGaussian`s to ImageJ-Ops. There are still open (internal) 'TODO's, however the algorithm is working and the API will not change.


ctrueden:
Looks great!

I stacked on some commits with small cleanups; I hope that is OK.

Issue #149, opened 2015-06-24
Generalize OutputOp and InputOp?


dietzc:
Hi all,

I ran into a situation where I would like to split or extend `OutputOp` and `InputOp`. `OutputOp` should extend `OutputGet` and `OutputSet` as in some situations you only want to use one of these interfaces instead of `OutputOp`. Also many implementations do only need a `OutputGet`.

Example: While implementing `View` wrappers, we recognized that we could certainly implement a `OutputGet` in a `View` but not really a `OutputSet` as setting the output doesn't make sense for a `View` in general. Actually, for all `Op`s and especially for `Computer`s `OutputGet` makes sense, this means for `Op`s which output is really not `Type.Both`. 

If we agree that this makese sense and also agree on a nice naming scheme, I'm happy to file a PR.

ctrueden:
The main use case is for "read-only" `OutputOp` implementations, correct? (I.e.: there is no "write-only" use case?) You could accomplish that at the moment by making `setOutput` simply `throw new UnsupportedOperationException()`. Would that be sufficient?

> Actually, for all `Op`s and especially for `Computer`s `OutputGet` makes sense, this means for `Op`s which output is really not `Type.Both`. 

I was recently talking to @awalter17 about a related issue: in OPS, the social contract of `Function` ops is that the have a `BOTH` output. There is no `PureFunction` or similar that has an `OUTPUT` output in cases where it passing in a preallocated output would not make sense.

Yes, there is `Computer` now, which has the `O compute(I in)` method. But the design of the `Function` type hierarchy seems a little strange to me since it cannot accommodate a `PureFunction` which has an `OUTPUT` output:

| interface | output type | output optional? | setOutput | createOutput | compute(I) | compute(I, O) |
| --- | --- | --- | --- | --- | --- | --- |
| `Function` | - | - | yes | no | no | yes |
| `OutputFunction` | BOTH | yes | yes | yes | yes | yes |
| `InplaceFunction` | BOTH | no | yes | no | yes | yes |
| `AbstractStrictFunction` | BOTH | no | yes | yes | no | yes |
| `PureFunction` (proposed) | OUTPUT | N/A | no | no | yes | no |

[...]

PR #165, opened 2015-07-21
Rework the design of the function type hierarchies

ctrueden:
As discussed in issue #149, the idea of a "function" was not quite intuitive from a mathematical perspective: its `compute(I, O)` method allowed (or sometimes required) passing a preallocated output object, into which the function result would be stored. This is not pure functional programming, but instead a hybrid approach making both stateful/in-place and functional paradigms possible.

This branch establishes four core op interfaces, from which most ops derive:
* `ComputerOp`, which defines `void compute(I, O)` (formerly `Function`)
* `FunctionOp`, which defines `O compute(I)` (formerly `Computer`)
* `HybridOp`, which extends both `ComputerOp` and `FunctionOp`, and adds an `O createOutput(I)` method (formerly `OutputFunction`)
* `InplaceOp`, which extends `ComputerOp` and defines `compute(A)` (formerly `InplaceFunction`)

And it provides an abstract base class for each of these interfaces.

With these changes, it is no longer allowed to pass `null` to `compute(I, O)` for the output reference; if you need that, use `HybridOp` and call `compute(I)` instead.

The branch also renames API and variables as appropriate to reflect this change.

Lastly, the branch contains many miscellaneous changes of various sorts, performed while working on the refactoring.


ctrueden:
@dietzc @hinerm I am now super happy with this work, and will merge tomorrow, but if you want to do a quick review in the meantime, that would be welcome! (Though I'm not sure how "quick" it is possible to be when reviewing ~50 commits...)

dietzc:
First of all: Thank you very much @ctrueden. I think this refactoring will _really_ help us and others understanding the concepts and API of Ops much better. I will review as much as possible today

Question 1: 
> FunctionOp, which defines O compute(I) (formerly Computer)
> HybridOp, which extends both ComputerOp and FunctionOp, and adds an O createOutput(I) method (formerly OutputFunction)

`FunctionOp` itself can already have the `O createOutput(I)` as it has to create an output, too?

ctrueden:
> `FunctionOp` itself can already have the `O createOutput(I)` as it has to create an output, too?

Sure, that would be OK! Note that I didn't think hard about incorporating your buffer factory stuff into the core design (yet), but if you have ideas there, I'm happy to discuss. I see that as an additional change on top of this redesign, though.

ctrueden:
One reason for `FunctionOp` not to have `createOutput` is for simplicity. We have two options in order to move it upstream that way:

1. Force every `FunctionOp` implementation to implement both methods (adds complexity).
2. Put a default `createOutput` implementation in `AbstractFunctionOp`, which simply does `return compute(I);` (will make many functions in practice do extra computation work when they are only supposed to be creating an output)

So my thinking on this is that we should instead focus on polishing the buffer factory idea a bit, and somehow work that in. Composition preferred to inheritance and all that.

dietzc:
The `BufferFactory` idea is just used for `Loop`s, right? I don't think we need  this concept in general, or do youthink its really useful?

dietzc:
> Put a default createOutput implementation in AbstractFunctionOp, which simply does return compute(I); (will make many functions in practice do extra computation work when they are only supposed to be creating an output) 

I wouldn't do that. this is really something where users of the library can do really bad things ....

ctrueden:
> The `BufferFactory` idea is just used for `Loop`s, right?

Right now, yes.

> I don't think we need  this concept in general, or do youthink its really useful?

It seems generalizable to me, but I am not sure. Like I said, I didn't think about it too much yet, it's just an intuition. My current idea is actually that the `createOutput` implementations for `HybridOp`s could lean on `ops.create().foo(...)` calls instead of using a special new type hierarchy. Then we could maybe even get rid of `BufferFactory` altogether.

ctrueden:
> I wouldn't do that. this is really something where users of the library can do really bad things ....

Agreed. So if you are OK with it, I'd like to simply leave `createOutput` in `HybridOp` and not `FunctionOp`. I feel like it makes functions unnecessarily complex otherwise.

dietzc:
> My current idea is actually that the createOutput implementations for HybridOps could lean on ops.create().foo(...) calls instead of using a special new type hierarchy. Then we could maybe even get rid of BufferFactory altogether. The CreateImgFromImg op is a perfect example of this.

This should definitively be the case, yes. However, in some situation you can't simply use `CreateImgFromImg` as you need more dedicated settings (e.g. projections require images with one dimension less).

The cool thing about the `BufferFactory` is, that you can reuse objects... (e.g. while looping your don't have to create one object after the other, just circulate three objects in total).

dietzc:
anyway, I think this work is great and from my side it would be great if you would merge this branch. We can fix thing we overlooked now in the next few weeks or at the hackathon, as many thing just pop up during daily work with the library.

ctrueden:
Thanks for taking the time to look it over!

PR #166, opened 2015-07-28
CreateImg / CreateLabeling: Respect minimum/maximum

dietzc:
Hi all,

until now, we were only able to create `Img`s or `ImgLabeling`s with zero minimum. However, in many situations (`Region`s, transformed `RandomAccessibleInterval`s etc.) the minimum can be non-zero. This PR add capabilities to ImageJ-Ops to create objects with non-zero minimum or more general to create `Img`s and `ImgLabeling`s in arbitrary `Interval`s.

### DIFF
--- a/src/main/java/net/imagej/ops/create/CreateNamespace.java
+++ b/src/main/java/net/imagej/ops/create/CreateNamespace.java
@@ -37,6 +37,7 @@ import net.imagej.ops.Namespace;
 import net.imagej.ops.OpMethod;
 import net.imagej.ops.Ops;
 import net.imglib2.Dimensions;
+import net.imglib2.Interval;
 import net.imglib2.img.Img;
 import net.imglib2.img.ImgFactory;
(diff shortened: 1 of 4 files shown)


ctrueden:
Great! This is nice to have.

I guess... we have to use `Views` because `ImgFactory` always gives back an `Img` with zero-based bounds? If so, then so be it—this approach is an elegant workaround.

dietzc:
> I guess... we have to use Views because ImgFactory always gives back an Img with zero-based bounds? If so, then so be it—this approach is an elegant workaround.

As far as I understand I can't create an image with non-zero bounds directly. Therefore I used the approach with translate + `ImgView`. It's really important that we respect min/max in algorithms.

ctrueden:
OK, I'll just merge this now then.

dietzc:
https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/img/ImgFactory.java to confirm: only accepts dimensions.

PR #167, opened 2015-07-28
add /bin/ exclude in gitignore

dietzc:
see text. I don't know why, but in my setting git always wants to add all files in the /bin/ folder. It would be great if we could add /bin/ to the exclude list.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 /.project
 /.settings/
 /target/
+/bin/


ctrueden:
I am OK with adding `/bin/`, although in my experience, when this happens it is because of a weirdness (bug?) in the Eclipse behavior. E.g.: I have seen it happen when Eclipse temporarily "forgets" the project is a Maven project, and instead defaults to its usual binary folder of `bin` instead of `target`. I am not actually sure what triggers this problem, and have only seen it a couple of times myself.

My point is: if this happens to you, there is something else wrong with your config and ignoring the `/bin/` folder will just mask that the config is messed up. Are you sure that's what you want?

OTOH, if everything seems to still work for you in this scenario, and it simply annoys you for Git to tell you about `/bin/`, then we can merge it.

ctrueden:
At minimum, I would try closing the project in eclipse, and reimporting, and see if the problem with `/bin/` persists.

dietzc:
I re-added the project to eclipse and the problem seems to be gone. I will reopen this PR if I experience the same problem again. 

Thanks!

PR #172, opened 2015-08-01
min/max of realtype can be hybrid ops

dietzc:
see above.


dietzc:
merged as this is not crucial and only wasting time of @ctrueden :+1:

Issue #173, opened 2015-08-01
LazyOps


dietzc:
Hi all,

while refactoring the `Normalize` ops a bit, I figured that the concept of `LazyOp`s now nearly fits into our op-design. The latest refactoring of @ctrueden made it possible that we can nicely implement `FunctionOp`s which do not create a new output, rather return a converted or transformed view on the input which is lazily evaluated.

However, nearly always (at least I can't think of a case where it doesn't work) the result of a lazy op can be copied into a container (mostly an `Img`) and therefore we can always offer a non-lazy version of a `LazyOp` or we can provide non-lazy versions of a `LazyOp` which are optimized and therefore faster than `LazyOp` +  `Copy`. Anyway, we want to provide the user the ability to chose what she wants. 

Here is the problem: From an API point of view  (`ops.namespace().something(...)`), the calls are exactly the same. We could add a `boolean doLazy` flag to the method signatures manually, but this doesn't fit with our `NamespaceTest`s. Additionally, we potentially may want to test if the `LazyOp` produces exactly the same result as the non-lazy version and also we may want to auto-generate (code-generation, or op-reducer) the non-lazy version of a `LazyOp` (if there is no faster special implementation available).

So here is my question: Can we somehow in integrate the concept of `LazyOp`s in our current ops-design?

Thanks,

Christian 

PS: Some examples: `Converter`: https://github.com/imagej/imagej-ops/pull/174 `View`: https://github.com/imagej/imagej-ops/pull/168
PPS: potentially this is all related to https://github.com/imagej/imagej-ops/issues/57

ctrueden:
There are three relevant patterns here:

1. __Explicit.__ Do work on input object `in`, saving the result to output object `out`. Pay time cost once, up front. Pay space cost once, up front.
2. __Lazy.__ Return an output object `out` that is a thin wrapper around input `in`. When API methods are called on `out`, work is done on demand. Pay time cost only on demand but potentially multiple times. Pay space cost on demand, potentially multiple times.
3. __Lazy + cached.__ As Lazy, but caching the result after it is computed once. This is the "lazy" approach taken in various other places in SJC, although the space costs for doing it are much more trivial than they are for images etc.

My preference for ops, in all cases, is to differentiate ops by: A) op name and B) input types. The `boolean isLazy` parameter does that, although it does not differentiate between patterns 2 and 3 above. It may be better to use a "hints" object for specifying this.

[...]

PR #174, opened 2015-08-01
Refactored Normalize implementations

dietzc:
* Implement `ComputerOp` and `FunctionOp`.
* Allows lazy normalize calls

Question: Can we somehow integrate the concept of `LazyOp`s better into our ops design? See: https://github.com/imagej/imagej-ops/issues/173


ctrueden:
Thanks!

PR #178, opened 2015-08-04
Add ImageNamespaceTest

dietzc:
it was missing. on purpose?


ctrueden:
It was missing by accident!

PR #179, opened 2015-08-05
PR: Haralick Features

dietzc:
### NOT MEANT TO BE MERGED, YET ###

Missing:
* [ ] Quick review
* [ ] Question: Are "nested" namespaces supported?
* [ ] Add/Finalize test (since now only a test was added to demonstrate API usage)

@ctrueden I started implementing a much simpler version of features. To illustrate the idea I implemented the so called Haralick-Features. The main idea is: I cache the results of a `op.run(...)` within a special `FeatureOpService`. In the case of features, we don't even need a very smart cache, as the results of the computation have to fit in memory anyway. However, we can simply delete this `FeatureOpService` as soon as the `DefaultOpService` supports caching and nearly no code will change (besides the code e.g. in the `HaralickFeatureSet`, where we have to remove the `FeatureOpService`). Like this I was also able to provide a very nice API in the `HaralickNamespace` such that features can be calculated individually. Again, as soon as `OpService` supports caching, even the evaluation of these features will be smart and redundant computations are avoided implicitly. 

Does this make sense? If you like the idea/workaround I will add _all_ other features as soon as possible!

Christian

### DIFF
--- a/src/main/java/net/imagej/ops/DefaultOpService.java
+++ b/src/main/java/net/imagej/ops/DefaultOpService.java
@@ -40,6 +40,7 @@ import java.util.Map;
 import net.imagej.ops.convert.ConvertPix;
 import net.imagej.ops.create.CreateNamespace;
 import net.imagej.ops.deconvolve.DeconvolveNamespace;
+import net.imagej.ops.features.haralick.HaralickNamespace;
 import net.imagej.ops.filter.FilterNamespace;
 import net.imagej.ops.image.ImageNamespace;
 import net.imagej.ops.labeling.LabelingNamespace;
@@ -557,6 +558,11 @@ public class DefaultOpService extends AbstractPTService<Op> implements
 	@Override
 	public FilterNamespace filter() {
 		return namespace(FilterNamespace.class);
+	}
+	
+	@Override
+	public HaralickNamespace haralick() {
+		return namespace(HaralickNamespace.class);
 	}
 
 	@Override
--- a/src/main/java/net/imagej/ops/OpRef.java
+++ b/src/main/java/net/imagej/ops/OpRef.java
@@ -29,6 +29,8 @@
  */
 package net.imagej.ops;
 
+import java.util.Arrays;
+
 
 /**
  * Data structure which identifies an op by name and/or type, along with a list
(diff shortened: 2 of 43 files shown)


ctrueden:
Regarding nested namespaces: Do we really need them? Really? It complicates things even further... but they could conceivably be added.

What were you thinking? Maybe we can come up with a nice single-layer naming scheme.

ctrueden:
I didn't review everything thoroughly, but I wrote some comments. All in all, this looks awesome. It does seem that most of the code will not need to be updated at all in response to a chance in how caching works.

So: fix what you want in response to my comments above, then feel free to merge. The only thing I have a real substantial opinion about is the `OpService` override—I think a `FeatureOpService` that has an `OpService` might be better unless there is some reason it fundamentally can't work.

dietzc:
> Regarding nested namespaces: Do we really need them? Really? It complicates things even further... but they could conceivably be added. What were you thinking? Maybe we can come up with a nice single-layer naming scheme.

I was wondering if we should collect all the feature implementations in `ops.feature.xy`, as there will be many (> 8) of them.

dietzc:
I cleaned up the branch a bit!

ctrueden:
> I was wondering if we should collect all the feature implementations in `ops.feature.xy`, as there will be many (> 8) of them.

Can you give an example? You mean that `xy` would be a subnamespace, right? Not merely a feature computation op for a feature named `xy`?

dietzc:
For example,  we do have several semantically different features/measurements/descriptors:

* `ops().stats().*`
* `ops().haralick().*`
* `ops().geometric().*`
* `ops().imagemoments().*`
* `ops().tamura().*`
* `ops().localbinarypatterns().*`
* `ops().shape().*`
* `ops().zernike().*`
* `ops()....().*`

More can easily be added in the future. that's why I was asking about:


* `ops().features().stats().*`
* `ops().features().haralick().*`
* `ops().features().geometric().*`
* `ops().features().imagemoments().*`
* `ops().features().tamura().*`
* `ops().features().localbinarypatterns().*`
* `ops().features().shape().*`
* `ops().features().zernike().*`
* `ops().features()....().*`

PR #180, opened 2015-08-07
Firstorderstats

seebacherd:
add new statistic ops without dependencies on features, etc.


DanielSeebacher:
> Mean exists only with dependencies but not direct
There is a ```DefaultMean``` which uses Size and Sum but also an ```IterableMean``` class which calculates the mean directly.

> Variance the other way round.
> also add min/max

Added indirect calculation of ```Min```, ```Max``` and ```Variance```. ```Min```, ```Max``` using the ```MinMax``` Op and ```Variance``` using the power of the ```StdDev```.

As for the formatting, I don't know why the space is missing before the implements. Seems to be a problem with Eclipse 4.5.

Squareys:
@dietzc Formatting done :)

dietzc:
@ctrueden  any comments? this affects some of the existing code therefore it would be great if you could have a look before merging.

DanielSeebacher:
@ctrueden did you have time to look at this pull request and #181?

ctrueden:
Very sorry Daniel, but I will not have time before the conference.

dietzc:
we can review at the hackathon, too. Anyway, I will merge the other branches!

PR #181, opened 2015-08-07
add image moments.

seebacherd:
pretty big commit. image moments were added to Ops.list and the namespace was added to OpService and DefaultOpService.
the whole implementation is in the net.imagej.ops.imagemoments package and the tests are in the net.imagej.ops.imagemoments.ImageMomentsTest class.

@dietzc  do you mind taking a look? also could you format the imagemoments packages in src/main and src/test?

### DIFF
--- a/src/main/java/net/imagej/ops/DefaultOpService.java
+++ b/src/main/java/net/imagej/ops/DefaultOpService.java
@@ -42,6 +42,7 @@ import net.imagej.ops.create.CreateNamespace;
 import net.imagej.ops.deconvolve.DeconvolveNamespace;
 import net.imagej.ops.filter.FilterNamespace;
 import net.imagej.ops.image.ImageNamespace;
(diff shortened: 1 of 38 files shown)


dietzc:
:+1: looks very, very good. can be merged after:

* [ ] remove format only commits
* [ ] format code with Luna Eclipse and not Mars

DanielSeebacher:
format only commits are removed. also it would be very nice if either you or @Squareys  could format the classes located in src/main/net/imagej/ops/imagemoments and src/test/net/imagej/ops/imagemoments.

dietzc:
one of us will take care, thank you!

dietzc:
one further question: are these implementations n-dimensional? if not, we should add `Contigent` to check.

dietzc:
I formatted the code and simplified to commit message.

PR #182, opened 2015-08-10
Convert ops

awalter17:
Hello!

These are the convert ops for converting between some of the imglib types (ByteType, UnsignedByteType, Unsigned2BitType, ComplexFloatType, etc.). Please do not merge these commits until a new version of imglib2 is released. 

Thank you in advance for your feedback!


dietzc:
Hi,

I'm sure I don't understand something here. Could you please give me a use-case where you want to use the `Convert` ops as they are? Usually, a `Type` is bound to a container so the common convert use-case is that I want for example convert from one `IterableInterval<T>` to another `IterableInterval<V>` or you might want to convert from a `T` to another `V` for op matching. In both scenarios the current implementations (as far as I understand them) won't help. In the first scenario we have to use the type `V` which is bound to the `IterableInterval<V>` as it might just act as an proxy to the underlying data structure. For the second scenario the converters should be scijava `Converters` (or actually both, imglib2 `Converter`s and scijava `Converter`s.

Sorry if I'm missing something,

Christian

awalter17:
Hello @dietzc 

I think that the convert ops have value for both of your use cases. In the first scenario they could be used via map to accomplish this conversion. With regards to the second scenario, I do agree that there should be scijava converters for this. And in fact I have written some for primitives (scijava/scijava-common#158) and definitely there should be some for imglib types as well. We do also plan to expose the primitive converters as ops in this same namespace with the same op names, so the utility is not limited to imglib based types.

Additionally having them in ops exposes them to the ops expression parser as functions.

Lastly, the convert ops are nice to have just for sheer convenience; there are likely other scenarios in which these would be useful. 

Hopefully this helps clarify the need for the convert ops a bit.

dietzc:
Hi @awalter17 ,

thanks for reply. Unfortunately you can't use the converters as they are in the current implementations of `Map` to convert an `Img<T>` to another `Img<V>` for two reasons:

* https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/map/MapOp.java requires a `ComputerOp` as input.
* In most of the `Img` container implementations there is no 1-to-1 mapping from position in the image to `Type`, i.e. not each pixel has it's own instance of `Type`. Usually, there is only a single instance of a `Type` which has a status, often an index pointing on a position in a container. So usually a conversion rom `Img<T>` to  `Img<V>`  works as follows: Create an empty `Img<V>` and use a `ComputerOp` in a `Map` to convert from `<T>` to `<V>`. The `Map` internally always uses the type associated with `Img<T>` and `Img<V>`.

So I suggest that the `Converter`s you implemented should be `ComputerOp`s or `HybridOp`s. 

Cheers,
Christian

PS: Another option is to do it lazy with https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/converter/Converters.java and by implementing the `ComputerOp` as a https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/converter/Converter.java. Like this `Img<V>` is actually never created. There are pros and cons for lazy evaluation and we have to discuss this in detail with Curtis and Mark at the hackathon.

awalter17:
Hello @dietzc, 

I apologize for the delay! The convert ops have been switched to HybridOps, and I have created tests for using them to convert an `Img<T>` to an `Img<V>` using Map. So they're now ready for review. 

Please let me know of any changes that need to be made. 

Thanks, 
Alison

src/main/java/net/imagej/ops/convert/ConvertNamespace.java
```diff
@@ -7,13 +7,13 @@
  * %%
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
- *
```
dietzc: please remove format commit

src/main/java/net/imagej/ops/convert/ConvertNamespace.java
```diff
@@ -7,13 +7,13 @@
...
  * modification, are permitted provided that the following conditions are met:
- * 
+ *
  * 1. Redistributions of source code must retain the above copyright notice,
  *    this list of conditions and the following disclaimer.
  * 2. Redistributions in binary form must reproduce the above copyright notice,
  *    this list of conditions and the following disclaimer in the documentation
  *    and/or other materials provided with the distribution.
- * 
+ *
```
dietzc: please remove format commit

src/main/java/net/imagej/ops/convert/Types.java
```diff
@@ -0,0 +1,186 @@
...
+
+import net.imglib2.type.numeric.integer.Unsigned128BitType;
+
+
+/**
+ * Utility methods for converting between primitive types.
+ * 
+ * @author Alison Walter
+ */
+public final class Types {
```
dietzc: @ctrueden @awalter17 I'm not sure if this maybe could be a `Service`?

dietzc:
looks good to me. @ctrueden what do you think? Does it match into what you did with `Converter`s recently?

ctrueden:
> Does it match into what you did with `Converters` recently?

There is likely more unification that could be done. For example, the `Types` utility class used here should probably go in SJC, and then used by the SJC primitive `Converter` implementations, too. I will try to find time to review both branches more thoroughly and suggest improvements to @awalter17 next week.

src/main/java/net/imagej/ops/convert/Types.java
ctrueden: The main advantage of making it a `Service` would be to let people override the primitive type conversion methods. But they can already do that at the Ops level by writing their own ops—or at the SJC level by writing their own higher-priority converters. So I think here it is unnecessary complication. Do you see any other advantage to a `Service` here?

awalter17:
Thank you for the feedback @ctrueden and @dietzc !

> Is there some reason you can't just write return new BigInteger("" + value) for this method?

You can't just write `return new BigInteger("" + value)` for this method because it could return a BigInteger that is negative and this method is for uint128. That being said the validBytes method may be a bit verbose for this instance. The validBytes method will still be useful for `public static BigInteger uint128(final BigInteger ...` which I just realized I implemented wrong. This method should mask the input to be unsigned and ensure that it is 128 bits. 
> Did you check the performance difference?

I looked into this and it is my understanding that `BigInteger.valueOf(value)` has better performance than the String constructor for `long`.
> So anyway, I think we want all uint64 methods to return a BigInteger. 

Agreed; however, the get() method for UnsignedLongType returns a `long`. So in my tests I would then have to cast these `BigInteger`s back to `longs`s. Perhaps UnsignedLongType should have a method to return a `BigInteger`? On that same note, should there be an [UnsignedLongType](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/type/numeric/integer/UnsignedLongType.java) that takes in a `BigInteger`?

Please let me know if any of that was unclear, or if I am misunderstanding anything.

awalter17:
Hello!

The converters are once again ready for review. I've modified the converters a bit to work with various edge cases. This branch also currently depends on the imglib2 branch [FixUnsignedLongType](https://github.com/imglib/imglib2/pull/109), so please do not merge this branch.

Thank you in advance for your feedback!

ctrueden:
ImgLib2 2.5.0 has been released, and pinned here. So this is good to merge now.

PR #184, opened 2015-08-13
Decon updates fr

bnorthan:
This pull request includes several deconvolution updates.  Most importantly

- better edge handling
- acceleration
- implementation of Richardson Lucy Total Variation algorithm

### DIFF
--- a/src/main/java/net/imagej/ops/deconvolve/DeconvolveNamespace.java
+++ b/src/main/java/net/imagej/ops/deconvolve/DeconvolveNamespace.java
@@ -33,6 +33,7 @@ package net.imagej.ops.deconvolve;
 import net.imagej.ops.AbstractNamespace;
 import net.imagej.ops.Namespace;
 import net.imagej.ops.OpMethod;
+import net.imglib2.Dimensions;
 import net.imglib2.Interval;
 import net.imglib2.RandomAccessibleInterval;
 import net.imglib2.img.Img;
@@ -46,13 +47,15 @@ import org.scijava.plugin.Plugin;
 
 /**
  * The deconvolve namespace contains deconvolution operations.
- *
+ * 
  * @author Alison Walter
  */
 @Plugin(type = Namespace.class)
 public class DeconvolveNamespace extends AbstractNamespace {
 
-	//-- Deconvolve namespace ops --
+	// -- Deconvolve namespace ops --
+
+	// -- DeconvolveOps.RichardsonLucy
 
 	@OpMethod(op = net.imagej.ops.Ops.Deconvolve.RichardsonLucy.class)
 	public Object richardsonLucy(final Object... args) {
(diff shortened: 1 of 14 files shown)


bnorthan:
Not quite ready to merge.  I have a question (@dietzc or @ctrueden (or others)) about the last commit.  

I had to change the 'default' 'richardsonLucy' call.  Originally the op service called 'RichardsonLucy.class', however RichardsonLucy is an interface, and the code did not seem to run.  I changed it to call 'RichardsonLucy.NAME' and it works.  Is this OK??  Or should the code that called 'RichardsonLucy.class' of worked?? 

```java
 	public Object richardsonLucy(final Object... args) {
-		return ops().run(net.imagej.ops.Ops.Deconvolve.RichardsonLucy.class, args);
+		return ops().run(net.imagej.ops.Ops.Deconvolve.RichardsonLucy.NAME, args);
```

dietzc:
Actually, `RichardsonLucy.class` should have worked, if `RichardsonLucy extends (Something which extends Op)`. I personally prefer the `.class` notation, as API-Breakages are indicated by eclipse.

dietzc:
one comment about `RAI` and `Img` implementations: I hope at the hackathon we come up with a solution to avoid the explicit creation of the `Img` implementations. There are several possibilities to solve this:

* OpReducers https://github.com/imagej/imagej-ops/issues/57
* Simply just expose the API calls in the namespace with (`RAI`) and without factory class. However, I don't like that `ImgFactory` has no super-interface.
* Auto-guess the best container factory in case of a `RandomAccessibleInterval` or so (as this is really deterministic in most of the cases).
* ...

bnorthan:
Hi Christian

I've divided the deconvolution algorithms into high level and low level parts.  It doesn't have to be Img and RAI.  (Img could be replaced by ImgPlus)

The high level part performs preconditioning then calls low level algorithm.   Right now it mainly extends the images, but we could also generate kernels here, divide into hyperslices (for time series, or multi channel images). 

The low level (RAI) part has no access to meta data, it is just a number cruncher.   In the future there could be different flavors of the low level implementation (for example Native wrappers, GPU implementation). 

Maybe at the Hackathon we can review and re-factor the deconvolution code??  Maybe with these goals

1.  Address your concerns
2.  Handle ND (multi channel, multi timepoint) images
3.  Automatically generate kernels from meta data (in the future diffraction based kernels will require meta data like NA, wavelength, and spacings)

bnorthan:
>Actually, RichardsonLucy.class should have worked, if RichardsonLucy extends (Something which >extends Op) and if your Op implementations inherit the interface of course. I personally prefer the >.class notation, as API-Breakages are indicated by eclipse.

@dietzc, yes, I stupidly forgot to inherit the interface, fixed now.

bnorthan:
@dietzc, @ctrueden, @hinerm 

If no one has objections I'd like to merge this and then look at refactoring it at the hackathon (ie eliminating or replacing the 'Img' implementations)

By the way, another reason I created the 'Img' interfaces was so that I can handle very large images using 'CellImg'.  So large image processing is another factor to consider, along with handling meta data and handling ND-data.

ctrueden:
Fine with me.

dietzc:
Sorry, I'm currently in italy and don't really have Internet here. Merging is fine. Just one comment about Img. I'm not against having an API with Imgs. I just think we should auto-generate it based on low level implementations. 

Looking forward to the hackathon!

PR #185, opened 2015-08-24
Zernike features

angrauma:
Add implementation and test of zernike features.

@dietzc do you mind taking a look?

### DIFF
--- a/src/main/java/net/imagej/ops/AbstractOpEnvironment.java
+++ b/src/main/java/net/imagej/ops/AbstractOpEnvironment.java
@@ -41,6 +41,7 @@ import net.imagej.ops.convert.ConvertNamespace;
 import net.imagej.ops.create.CreateNamespace;
 import net.imagej.ops.deconvolve.DeconvolveNamespace;
 import net.imagej.ops.features.haralick.HaralickNamespace;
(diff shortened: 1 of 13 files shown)


dietzc:
@angrauma any comments on this?

dietzc:
@angrauma  can you also change them according to first order statistics, haralick and moments (see master)?

angrauma:
Modified the zernike implementation, using now the center of the given region as origin of the circle. @dietzc and @LeeKamentsky, can you please have a look at it.

src/main/java/net/imagej/ops/DefaultOpService.java
```diff
@@ -201,12 +201,14 @@ public void setPriority(final double priority) {
...
+
 	}
 
 	@Override
 	public void setInfo(final PluginInfo<?> info) {
 		this.info = info;
 	}
+	
 
 	// -- Comparable methods --
```
dietzc: Format commit. Please get rid of it ;-)

dietzc:
as soon as the format commit is gone I think its ready to merge. @LeeKamentsky do you agree? Thanks for reviewing btw!

LeeKamentsky:
Yup, this looks pretty good. I think you will like these features - they're great as inputs for a classifier for certain types of cell images.

PR #186, opened 2015-08-26
Project fix

bnorthan:
@dietzc did someone from Konstanz write DefaultProjectParallel??  I think I found an error.  It would be great if the author of DefaultProjectParallel could take a quick look at this pull request and let me know if my change makes sense.


dietzc:
nice catch! yeah, if `d>dim` we should subtract one or add `-1`.  thank you

PR #169, opened 2015-07-30
Neighborhood maps and center aware neighborhood operations

Squareys:
Hello @ctrueden !

Here are some improvements/changes/additions to neighborhood maps:

I added a `CenterAwareComputerOp` to allow implementation of neighborhood operations which require access to the center pixel value of a neighborhood (prime example, local thresholding). And a map which maps these to neighborhoods.

I moved the neighborhood related classes to a separate package in map, since there are more to come (on another day, on another PR). Since `CenterAwareComputerOp` is only relevant for neighborhood operations, it resides in that package aswell.

As soon as this is merged, I will prepare a PR for local thresholding. 

Greetings, Squareys

PS: I used the eclipse prefs/formatter/cleanup, but things seemed not quite right (no whitespace between closing `>` and `implements`, non-alphabetic import order...). I did pay careful attention to remove weird reformatting code changes, though.

PR #187, opened 2015-08-28
Linear and Non linear filters

Squareys:
Hello everybody, hello @dietzc !

Since #169 was finally merged, here is another pull request based upon that work.
This PR implements the following:
 - Max filter
 - Mean filter
 - Median filter
 - Min filter
 - Variance filter

TODO:
  - [x] Tests
  - [x] Sigma filter
  - [x] Op methods

### DIFF
--- a/src/main/java/net/imagej/ops/filter/AbstractCenterAwareNeighborhoodBasedFilter.java
+++ b/src/main/java/net/imagej/ops/filter/AbstractCenterAwareNeighborhoodBasedFilter.java
@@ -1,0 +1,82 @@
+
+package net.imagej.ops.filter;
+
+import net.imagej.ops.AbstractComputerOp;
+import net.imagej.ops.OpService;
+import net.imagej.ops.map.neighborhood.CenterAwareComputerOp;
+import net.imglib2.RandomAccessibleInterval;
+import net.imglib2.algorithm.neighborhood.Shape;
+import net.imglib2.outofbounds.OutOfBoundsFactory;
+import net.imglib2.view.Views;
+
+import org.scijava.plugin.Parameter;
+
+public abstract class AbstractCenterAwareNeighborhoodBasedFilter<I, O>
+	extends
+	AbstractComputerOp<RandomAccessibleInterval<I>, RandomAccessibleInterval<O>>
+{
+
+	@Parameter
+	protected OpService ops;
+
+	@Parameter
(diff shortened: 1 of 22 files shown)


dietzc:
I changed the title to "Linear and ..." filters as `MeanFilter` is part of this PR. Anyway, it seems like parts of https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/filter/convolve/ConvolveNaive.java can be replaced with your implementation @Squareys?

Squareys:
@dietzc Yeah, well, it never came to me that you could reverse the Mean filter.

This is ready for merge now :)

dietzc:
awesome. thanks.

LeeKamentsky:
Re the mean filter on a rectangle implemented as a kernel of 1/m/n: it's seperable (see http://blogs.mathworks.com/steve/2006/11/28/separable-convolution-part-2/ for how to detect a separable filter and reduce it to a chain of 1D convolutions). That means it runs in m+n instead of m*n. An improvement to ConvolveNaive would be to check for separability. N-D Gaussian filters are seperable and traditionally implemented as a chain of 1D filters.

dietzc:
@LeeKamentsky I didn't know that a `MeanFilter` isn't separable, but it makes sense of course. You think we should move it to non-linear category then? 

All separable filters can be applied using `SeperableSymmetricConvolution` of imglib2. However, finding out if a filter is seperable isn't easy (numerical reasons make the test inexakt).

LeeKamentsky:
Mean is linear and seperable if calculated on a rectangle (but not separable if calculated on a circle). I think the roundoff error for the singular value decomposition is 2 * eps * max(dims) where eps is the difference between the mean of the kernel in your floating point representation and the next floating point number  that is larger than it. So if you have an SVD function, it's "easy" (by which I mean possible, given a whole lot of effort).

dietzc:
> Mean is linear and seperable if calculated on a rectangle (but not separable if calculated on a circle).

Makes sense, yes.

> I think the roundoff error for the singular value decomposition is 2 * eps * max(dims) where eps is the difference between the mean of the kernel in your floating point representation and the next floating point number that is larger than it. So if you have an SVD function, it's "easy" (by which I mean possible, given a whole lot of effort).

Thank you. I always had the numerical problem here and didn't know how to solve it.

PR #189, opened 2015-08-31
Fix the order of arguments for the create op to make the test run

gab1one:
The order of arguments lead to an exception at run time.

### DIFF
--- a/src/test/java/net/imagej/ops/filter/FFTTest.java
+++ b/src/test/java/net/imagej/ops/filter/FFTTest.java
@@ -101,14 +101,14 @@ public class FFTTest extends AbstractOpBenchmark {
 
 			// create an input with a small sphere at the center
 			Img<FloatType> inOriginal =
-				(Img<FloatType>) ops.create().img(new FloatType(),
-					new ArrayImgFactory<FloatType>(), originalDimensions);
+				(Img<FloatType>) ops.create().img(originalDimensions, new FloatType(),
+					new ArrayImgFactory<FloatType>());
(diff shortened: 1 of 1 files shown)


dietzc:
I really don't understand what you are doing here? You fixed the order of arguments in tests which actually work fine on master?

gab1one:
The test does not work on master, no one noticed because it is skipped by default, as it inherits from AbstractOpBenchmark!
I ran it manually by commenting out [this](https://github.com/gab1one/imagej-ops/blob/fix-fft-test/src/test/java/net/imagej/ops/benchmark/AbstractOpBenchmark.java#L48) line, and it failed.

ctrueden:
Hmm, Jenkins _does_ pass `-Dimagej.ops.benchmark.tests`, so I'd expect Jenkins should have caught this?

gab1one:
I guess it does not, try commenting out  [this](https://github.com/gab1one/imagej-ops/blob/fix-fft-test/src/test/java/net/imagej/ops/benchmark/AbstractOpBenchmark.java#L48) and then just run the test in eclipse.
Looking at the default methods in the ops.create namespace, it is obvious, that one should put the dimensions before the type and the factory argument.
I figure this is just an leftover from the create namespace refactoring.

ctrueden:
@gab1one Can you please test the latest master and see whether the benchmarks all pass? And if not, rebase this fix so we can merge it? Thanks!

gab1one:
Actually the current FFTTestCode is already fixed. There is no need to merge this change anymore.

PR #193, opened 2015-09-08
Thinning from Knime/Knip

kephale:
Here is the thinning code we'd talked about in gitter. It seems to be working with minimal modification.

Note, the Knime copyright info remains.

### DIFF
--- a/src/main/java/net/imagej/ops/thinning/ThinningOp.java
+++ b/src/main/java/net/imagej/ops/thinning/ThinningOp.java
@@ -1,0 +1,221 @@
+/*
+* ------------------------------------------------------------------------
+*
+* Copyright (C) 2003 - 2013
+* University of Konstanz, Germany and
+* KNIME GmbH, Konstanz, Germany
+* Website: http://www.knime.org; Email: contact@knime.org
+*
+* This program is free software; you can redistribute it and/or modify
+* it under the terms of the GNU General Public License, Version 3, as
+* published by the Free Software Foundation.
+*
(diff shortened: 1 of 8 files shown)


dietzc:
@kephale sorry that this takes so long for review. I will give me best to get it done this week

kephale:
No worries. Give me a shout if you think anything else needs to be done.

ctrueden:
Looks like there is a licensing issue: the code you adapted is [GPL](http://imagej.net/GPL). But ImageJ Ops is licensed under [BSD-2](http://imagej.net/BSD-2).

Do we have permission from the KNIME authors to relicense this?

If not, it will need to go into a separate component.

kephale:
That is what I was wondering about. @dietzc said something that made it seem like this sort of code from KNIME was fair game for adding to imagej-ops.

gab1one:
@kephale a few points to improve this PR:
- [ ] Change the license in each file to BSD, this should be no problem as we want as many KNIP algorithms in Ops as possible.
- [ ] The code does use the current API as it was written before ImageJ-Ops, it needs to be adapted to become a proper op. This entails:
    - [ ] Inheriting from the correct Op / Computer /...   supertype
    - [ ] ``@Parameter`` annotations for member injection
    - etc.
- [ ] Add default methods calling the ops to a namespace (if no fitting one exists add one), making them easy to discover and call.

kephale:
Closed and moved to https://github.com/imagej/imagej-ops/pull/317

Issue #183, opened 2015-08-13
Refactor DefaultCreateImg


tibuch:
See #194. I will take care.

ctrueden:
Work continues on #194.

Issue #160, opened 2015-07-06
Generic typing of create namespace ops is broken


ctrueden:
The create namespace has several ops that use generics in unsafe and invalid ways. Whenever possible, it is best for methods which returns objects with generic parameters to accept method arguments that also specify those types. Otherwise, the compiler has a hard time guaranteeing that those generic type parameters will be fulfilled as expected.

ctrueden:
Regarding `DefaultCreateImg`: we should split it into multiple ops with different signatures, since you need to pass either a `T type` or an `ImgFactory<T>` or a `Class<T>`—_something_ to provide the `T` at runtime. And the cast from `dims` to `Img<?>` to (effectively) `Img<T>` is also bad—we should just have a separate (higher priority) op implementation that takes an `Img<T>` or `RandomAccessible<T>` or whatever typed on `T` and thus will have an associated `ImgFactory<T>` and not need to do a potentially broken cast that way. And then `DefaultCreateImg` can accept none of the `T` parameters at all and simply be typed on `DoubleType` directly, since that's ultimately what it will produce.

dietzc:
@ctrueden  https://github.com/imagej/imagej-ops/pull/194 does this help?

ctrueden:
@dietzc Yes I saw that, and it is exactly what we need to do! We should of course review and merge that PR. Then we can review the state of the create namespace as a whole again. If all the generics issues are resolved, we then close this issue—and if not, we enumerate the remaining problems here and solve them.

ctrueden:
The current plan is to use `FunctionViaComputer` in the `net.imagej.ops.chain` package to separate the computer and function portion of ops that use generics in their output type.

See also the [remove-varargs](https://github.com/imagej/imagej-ops/compare/remove-varargs) branch where this work is likely to proceed.

ctrueden:
Big progress in 267e2655191cf10da6589ba9cd3864db664808b1. But I am going to leave this issue open until I have time to verify that _all_ `create` ops have been fixed.

dietzc:
> Big progress in 267e265.

:+1:

PR #194, opened 2015-09-08
Refactor default create img

tibuch:
Hi @ctrueden and @dietzc I refactored DefaultCreateImg #183.

Now DefaultCreateImg requires some Dimensions and a Type as input parameters. The ImgFactory is an optional parameter. 

The image type has to be passed as an argument if a new image is created. I think this is the way to go because if someone creates a new image from scratch the type of the image should be set by the caller and not by Ops. 

I hope this is what you had in mind. If not I will refactor again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,8 +47,9 @@ Try this Jython script in ImageJ's
 
 # create a new blank image
 from jarray import array
+from net.imglib2.type.numeric.real import DoubleType
 dims = array([150, 100], 'l')
-blank = ij.op().create().img(dims)
+blank = ij.op().create().img(dims, DoubleType())
 
 # fill in the image with a sinusoid using a formula
 formula = "10 * (Math.cos(0.3*p[0]) + Math.sin(0.3*p[1]))"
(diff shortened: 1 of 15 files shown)


ctrueden:
I love what this branch is trying to do, but the quality of the commits needs work. I also have some competing changes on the `special-types` branch; e.g.: a60bbc87b61afe3ba4e688b27063bbcc3b8b2369 and 602512055cb89b6adb0ad0792ae6ca7c508c7147.

I will keep working on that branch, and reconcile with this work afterward, I hope.

dietzc:
can this be closed @ctrueden?

ctrueden:
Yes, it is fixed in 267e2655191cf10da6589ba9cd3864db664808b1. Mostly. At least; `create.img` is fixed (I hope). There are still a couple of `create` ops which may need work—I didn't fully finish looking through them. In particular, `create.imgLabeling` is very likely still broken. But that will stay on the radar thanks to #160, which I will leave open for now.

PR #197, opened 2015-09-10
Create HelpForNamespace op

### LINKED ISSUE #195: Make a help for namespace op
This would print out `HelpForOp` style information for all `@OpMethods` in a given `Namespace`

hinerm:
Delegates Help calls to all Ops in a given Namespace

Closes #195

### DIFF
--- a/src/main/java/net/imagej/ops/AbstractOpEnvironment.java
+++ b/src/main/java/net/imagej/ops/AbstractOpEnvironment.java
@@ -217,6 +217,14 @@ public abstract class AbstractOpEnvironment extends AbstractContextual
 	}
 
 	@Override
+	public String help(final Namespace namespace) {
+		final String result =
+			(String) run(net.imagej.ops.help.HelpForNamespace.class, namespace);
+		return result;
+	}
+
+
+	@Override
 	public String help() {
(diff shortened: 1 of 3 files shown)


hinerm:
***DO NOT MERGE***

This doesn't quite work yet.

@ctrueden said not to use `@OpMethod`. But the problem is that without it I'm not sure how to get the `Op`s associated with a workspace. `OpMethod` has runtime retention and a link back to the `Op` class so it seems almost 100% perfect.

The deficiency at the moment is that you need to pass parameters to the `OpService.op` method to get the right `Op` to look up its `CommandInfo`.

I think we can do this by using the parameters of the methods that are annotated with `OpMethod`... but that would only work for primitives/things with sensibile defaults?

But it feels like maybe there should be a nicer way to do this...

ctrueden:
The way to do this at the moment, to preserve extensibility, is to call `OpEnvironment.infos()` to get all the `ModuleInfo` objects. Then ask for their names. Keep the ones that start with the matching namespace string and a period character. Simple, works, but not the most performant. We could optimize later if needed.

hinerm:
@ctrueden thanks, works great!

PR #198, opened 2015-09-14
[NOT READY FOR MERGE] Type match fix

ctrueden:
Just filing this so we don't forget to address it. See also scijava/scijava-common#158, which is probably a prerequisite to solving this properly.


dietzc:
@ctrueden do we still need this?

ctrueden:
Someone (probably me) will need to dig in to the type matching stuff rather heavily to decide exactly where such logic should live, in terms of both elegance and performance.

For the moment, this change probably don't hurt, and likely improves performance. But I'd want to test to be certain.

If you don't have any acute type matching issues right now, let's let it sit a while...

ctrueden:
Type matching (for both op matching and converters) is being reworked by me and @hinerm very soon. No need for this temporary hack.

PR #205, opened 2015-09-16
CachedOpEnviroment

dietzc:
Adds a `CustomOpEnvironment` which cases results of `FunctionOp`s and `HybridOp`s using scijava-cache.

TODO before merge:

- [x] make scijava-common-2.47.0 available in ops
- [x] make scijava-cache available in ops
- [ ] potentially delegate `ops.op(...)` calls to specialized methods `ops.function(...)` and `ops.hybrid(...)`.

@ctrueden what do you think?

### DIFF
--- a/src/main/java/net/imagej/ops/cached/CachedOpEnvironment.java
+++ b/src/main/java/net/imagej/ops/cached/CachedOpEnvironment.java
@@ -28,7 +28,7 @@
  * #L%
  */
 
-package net.imagej.ops.featuresets;
+package net.imagej.ops.cached;
 
 import java.util.Collection;
 
--- a/src/main/java/net/imagej/ops/featuresets/CachedOpEnvironment.java
+++ b/src/main/java/net/imagej/ops/featuresets/CachedOpEnvironment.java
@@ -1,0 +1,212 @@
+/*
+ * #%L
+ * ImageJ software for multidimensional image processing and analysis.
+ * %%
+ * Copyright (C) 2014 - 2015 Board of Regents of the University of
+ * Wisconsin-Madison, University of Konstanz and Brian Northan.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * 
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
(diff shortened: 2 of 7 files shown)


ctrueden:
Looks pretty good. What do you mean by:

> potentially delegate `ops.op(...)` calls to specialized methods `ops.function(...)` and `ops.hybrid(...)`

?

Do you mean so that all `ops.op(...)` calls which result in a `FunctionOp` are actually cached? If so, why not simply override the `op(...)` methods to check `instanceof FunctionOp` and act accordingly? Well, there is one reason: it won't then actually be a `OP extends Op` class anymore, but rather a `CachedFunctionOp`, which as we discussed no longer inherits from the desired interface. So you may _need_ to limit behavior to `ops.function(...)` after all.

dietzc:
@ctrueden you are right. However, if we create a new `CachedOpEnvironment` and call e.g. `cachedOpEnvironment.stats().min(XY)`, the call will not be cached.

dietzc:
> Do you mean so that all ops.op(...) calls which result in a FunctionOp are actually cached? If so, why not simply override the op(...) methods to check instanceof FunctionOp and act accordingly? 

Right. But this can be done in the `AbstractOpEnvironment` or should I rather do it in the `CachedOpEnvironment`?

ctrueden:
So, as I also mentioned in the Gitter channel: we may need to relax the return value of `op(Class<OP>, ...)` to return simply `Op`.

And make sure there is a common method of the environment you can override that will affect all calls including `op`, `function`, `hybrid`, etc.

ctrueden:
Gotta run right now—maybe we could video chat about this tomorrow some time?

dietzc:
> So, as I also mentioned in the Gitter channel: we may need to relax the return value of op(Class<OP>, ...) to return simply Op. And make sure there is a common method of the environment you can override that will affect all calls including op, function, hybrid, etc.

@ctrueden 
Sounds great. We could file a separate issue for this? I figured this PR is independent from that discussion. So as soon as we have scijava-common 2.47.0 in pom-imagej I can clean-up the pom commit and merge?

dietzc:
I've opened https://github.com/imagej/imagej-ops/issues/214 for the relaxing of `ops.op(...)` requirements.

dietzc:
@ctrueden feel free to merge if you think its OK.

ctrueden:
:+1:

PR #210, opened 2015-09-17
implement MinMax as FunctionOp

dietzc:
see above.


dietzc:
I'd love I would be a cool kid, too. But I somehow can't.

ctrueden:
Awesome! I just posted that one nitpick about imports. Otherwise looking great!

dietzc:
I think there is still a bug in Mars that prevents the Formatter to work well.

ctrueden:
Yeah, I heard about that, so I am sticking with Luna for now.

dietzc:
I should do that, too. I will quickly format everything

dietzc:
better?

ctrueden:
:+1:

PR #211, opened 2015-09-17
refactor LocalThresholding: use initialize

dietzc:
see above


dietzc:
true, true and true.

ctrueden:
I think the typing can be nicer if you do the raw cast in initialize for each of the ops, rather than keeping raw(ish) references.

dietzc:
updated :+1: thanks!

dietzc:
updated again

ctrueden:
Thanks!

PR #215, opened 2015-09-21
Fixed bug in NormalizeRealTypeComputer

stelfrich:
I have tried to normalize an input image to the [0.0, 1.0] range with `opService.image().normalize(output, input, min, max, new DoubleType(0.0), new DoubleType(1.0));`. This wasn't very successful, since `tmp` is not correctly set, when both `sourceMin` and `sourceMax` are provided (as in my case) which in turn sets `factor` to `Double.INFINITY`.

I am not sure if that is desired behavior or not. In any case, this PR fixes the issue for me..

### DIFF
--- a/src/main/java/net/imagej/ops/image/normalize/NormalizeRealTypeComputer.java
+++ b/src/main/java/net/imagej/ops/image/normalize/NormalizeRealTypeComputer.java
@@ -55,7 +55,10 @@ class NormalizeRealTypeComputer<T extends RealType<T>> extends
 	{
 		double tmp = 0.0;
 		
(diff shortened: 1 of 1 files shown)


dietzc:
great! thanks for the fix. One thing I would add is a check if both `sourceMin` and `sourceMax` are actually non-null, because one of both still could be `null`. Additionally, I wouldn't put in the else part, rather I would do something like ```if(sourceMin != null && sourceMax != null){ ...}else{ // here the current if goes in. }```

stelfrich:
@dietzc: Makes sense. Hadn't thought about that case..

dietzc:
great! :+1:

PR #216, opened 2015-09-21
Descriptor3D and Geometric3D Features

tibuch:
I refactored the geometric3d features and from the outputop-service branch. 

The following features can be computed on 3D LabelRegions:
- [x] Volume
- [x] Surface Area
- [x] Surface Pixels
- [x] Main Elongation
- [x] Median Elongation
- [x] Compactness
- [x] Spareness
- [x] Sphericity
- [x] Convex Hull Volume
- [x] Convex Hull Surface Area
- [x] Convex Hull Surface Pixels
- [x] Rugosity
- [x] Solidity
- [x] Convexity

Volume, Surface Area, Surface Pixels, Main/Median Elongation, Compactness, Spareness and Sphericity are tested against [ImageJ 3D Suite](http://imagejdocu.tudor.lu/doku.php?id=plugin:stacks:3d_ij_suite:start).

Convex Hull Volume/Surface are tested against [qhull](qhull.org). 

Rugosity, Solidity and Convexity are all based on the other verified features.


dietzc:
awesome! I will review this at latest tomorrow. can you try to rebase master and see whats going wrong?

src/main/java/net/imagej/ops/AbstractOpEnvironment.java
```diff
@@ -697,6 +699,16 @@ public ThreadNamespace thread() {
...
+	
+	@Override
+	public Geometric3DNamespace geometric3d() {
+		return namespace(Geometric3DNamespace.class);
+	}
+	
+	@Override
+	public Descriptor3DNamespace descriptor3d() {
+		return namespace(Descriptor3DNamespace.class);
+	}
```
dietzc: alphabetic order of namespaces

src/main/java/net/imagej/ops/descriptor3d/AbstractPolygon.java
```diff
@@ -0,0 +1,136 @@
+package net.imagej.ops.descriptor3d;
```
dietzc: missing license

src/main/java/net/imagej/ops/descriptor3d/AbstractPolygon.java
```diff
@@ -0,0 +1,136 @@
...
+
+import java.util.List;
+
+/**
+ * An {@link AbstractPolygon} consists of vertices and neighbors.
+ * 
+ * @author Tim-Oliver Buchholz, University of Konstanz.
+ *
+ */
+public abstract class AbstractPolygon {
```
dietzc: can we give this class another name as there already exists a `Polygon` in imglib2-roi. Maybe `UpdateablePointSet`? Because this is not really a `Polygon`. Furthermore, I would prefer package visibility in for this class.

src/main/java/net/imagej/ops/descriptor3d/AbstractPolygon.java
```diff
@@ -0,0 +1,136 @@
+package net.imagej.ops.descriptor3d;
+
```
dietzc: missing license

src/main/java/net/imagej/ops/descriptor3d/AbstractVertexInterpolator.java
```diff
@@ -0,0 +1,35 @@
...
+import net.imagej.ops.Op;
+
+/**
+ * This is the {@link AbstractVertexInterpolator}. A vertex interpolator
+ * computes the real coordinates based on the pixel intensities. 
+ * 
+ * @author Tim-Oliver Buchholz, University of Konstanz
+ *
+ */
+public abstract class AbstractVertexInterpolator extends AbstractOp implements VertexInterpolator {
```
dietzc: how does this relate to the bresenham implementation in imglib2-roi? maybe you can reuse this?

src/main/java/net/imagej/ops/descriptor3d/DefaultBitTypeVertexInterpolator.java
```diff
@@ -0,0 +1,27 @@
...
+ * The {@link DefaultBitTypeVertexInterpolator} returns the point which is 
+ * in the middle of the two input vertices. 
+ * 
+ * @author Tim-Oliver Buchholz, University of Konstanz.
+ *
+ */
+@Plugin(type = Op.class, name = Descriptor3D.BitTypeVertexInterpolator.NAME)
+public class DefaultBitTypeVertexInterpolator extends AbstractVertexInterpolator implements BitTypeVertexInterpolator {
+
+	@Override
```
dietzc: Why BitType?

src/main/java/net/imagej/ops/descriptor3d/DefaultCentroid.java
```diff
@@ -0,0 +1,46 @@
...
+import net.imglib2.type.BooleanType;
+
+/**
+ * This {@link Op} computes the centroid of a {@link IterableRegion} (Label).
+ * 
+ * @author Tim-Oliver Buchholz, University of Konstanz.
+ *
+ * @param <B> a Boolean Type
+ */
+@Plugin(type = Op.class, name = Descriptor3D.Centroid.NAME)
```
dietzc: Do you need the `IterableRegion`? Can't you simply use the `CenterOfGravity` or `Centroid` from `Geometric2D'?
Maybe, we should introduce a namespace `geometric` comprising all n-dimensional implementations?
see also: https://github.com/imagej/imagej-ops/issues/222

src/main/java/net/imagej/ops/descriptor3d/DefaultFacets.java
```diff
@@ -0,0 +1,146 @@
...
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * This is the default implementation of {@link Facets}. 
+ * 
+ * @author Tim-Oliver Buchholz, University of Konstanz
+ *
+ */
+public class DefaultFacets implements Facets, Iterable<TriangularFacet> {
```
dietzc: Many `Facets` build a mesh? is this correct? Why not naming `Mesh` then?

src/main/java/net/imagej/ops/descriptor3d/FacetInterface.java
```diff
@@ -0,0 +1,5 @@
+package net.imagej.ops.descriptor3d;
+
+public interface FacetInterface {
```
dietzc: Why `Interface` and not just `Facet`

src/main/java/net/imagej/ops/descriptor3d/MarchingCubes.java
```diff
@@ -0,0 +1,497 @@
+package net.imagej.ops.descriptor3d;
```
dietzc: Interface + Default implementation?

src/main/java/net/imagej/ops/descriptor3d/MarchingCubes.java
```diff
@@ -0,0 +1,497 @@
...
+			0xca0, 0xda9, 0xea3, 0xfaa, 0x8a6, 0x9af, 0xaa5, 0xbac, 0x4ac,
+			0x5a5, 0x6af, 0x7a6, 0x0aa, 0x1a3, 0x2a9, 0x3a0, 0xd30, 0xc39,
+			0xf33, 0xe3a, 0x936, 0x83f, 0xb35, 0xa3c, 0x53c, 0x435, 0x73f,
+			0x636, 0x13a, 0x033, 0x339, 0x230, 0xe90, 0xf99, 0xc93, 0xd9a,
+			0xa96, 0xb9f, 0x895, 0x99c, 0x69c, 0x795, 0x49f, 0x596, 0x29a,
+			0x393, 0x099, 0x190, 0xf00, 0xe09, 0xd03, 0xc0a, 0xb06, 0xa0f,
+			0x905, 0x80c, 0x70c, 0x605, 0x50f, 0x406, 0x30a, 0x203, 0x109,
+			0x000 };
+
+	// For each of the possible vertex states listed in aiCubeEdgeFlags there is
```
dietzc: reference?

src/main/java/net/imagej/ops/descriptor3d/MarchingCubes.java
```diff
@@ -0,0 +1,497 @@
...
+	private Cursor<T> getCube(
+			ExtendedRandomAccessibleInterval<T, RandomAccessibleInterval<T>> extended,
+			int cursorX, int cursorY, int cursorZ) {
+		return Views.flatIterable(
+				Views.interval(extended, new FinalInterval(new long[] {
+						cursorX, cursorY, cursorZ }, new long[] { cursorX + 1,
+						cursorY + 1, cursorZ + 1 }))).cursor();
+	}
+
+	// For any edge, if one vertex is inside of the surface and the other is
```
dietzc: reference

src/main/java/net/imagej/ops/descriptor3d/QuickHull3D.java
```diff
@@ -0,0 +1,545 @@
...
+ * "The Quickhull Algorithm for Convex Hulls" by Barber, Dobkin and Huhdanpaa
+ * (http://dpd.cs.princeton.edu/Papers/BarberDobkinHuhdanpaa.pdf).
+ * 
+ * The computation of the initial simplex is inspired by John Lloyd's quickhull
+ * implementation (http://www.cs.ubc.ca/~lloyd/java/quickhull3d.html).
+ * 
+ * @author Tim-Oliver Buchholz, University of Konstanz
+ *
+ */
+@Plugin(type = Op.class, name = ConvexHull3D.NAME)
```
dietzc: `m_` is very KNIMEish

src/main/java/net/imagej/ops/descriptor3d/TriangularFacet.java
```diff
@@ -0,0 +1,241 @@
...
+ * 
+ * @author Tim-Oliver Buchholz, University of Konstanz
+ *
+ */
+public class TriangularFacet extends AbstractPolygon implements FacetInterface {
+
+	/**
+	 * The centroid of this facet.
+	 */
+	private Vector3D m_centroid = null;
```
dietzc: `m_`?

src/main/java/net/imagej/ops/descriptor3d/MarchingCubes.java
```diff
@@ -0,0 +1,497 @@
...
+ * Especially the lookup tables are from his implementation. 
+ * 
+ * @author Tim-Oliver Buchholz, University of Konstanz
+ *
+ * @param <T> BooleanType
+ */
+@Plugin(type = Op.class, name = Descriptor3D.Polygonize.NAME)
+public class MarchingCubes<T extends BooleanType<T>> extends
+		AbstractFunctionOp<RandomAccessibleInterval<T>, DefaultFacets>
+		implements Polygonize, FunctionOp<RandomAccessibleInterval<T>, DefaultFacets> {
```
dietzc: Where does the `Polygonize` interface stem from? Better `Interface` + `Default` as usual.

dietzc:
looks already really, really great @tibuch . Besides the comments in the code some general issues:

* [ ] rebase master into branch (and resolve conflicts)
* [ ] License headers missing
* [ ] `m_` for members is not our code style
* [ ] Set special descriptor / geometric 3d classes to package visibility only
* [ ] `Geometric3D` namespace is enough (no descriptor3D required)
* [ ] Introduce generic `Geometric` namespace for n-dimensional implementations.

src/main/java/net/imagej/ops/descriptor3d/AbstractPolygon.java
ctrueden: You can add all missing licenses by running:
```shell
mvn license:update-file-header
```

src/main/java/net/imagej/ops/descriptor3d/AbstractPolygon.java
ctrueden: My opinionated suggestions about class naming:

* If you want to use a class as the entry point into a data structure, don't prefix `Abstract`. Needing to write e.g. `List<AbstractFoo>` is ugly and unnecessary. Just name it `Foo` and write `List<Foo>` in that case.
* It is bad design to name a class `AbstractFoo` unless there is a `Foo` interface it implements. And in that case, you should type things on `Foo` and never `AbstractFoo`. The `AbstractFoo` class is an _SPI_-level class, intended for downstream _implementors_, but not downstream API _consumers_.

src/main/java/net/imagej/ops/descriptor3d/FacetInterface.java
ctrueden: > Why `Interface` and not just `Facet`

:+1: Agreed.

src/main/java/net/imagej/ops/descriptor3d/QuickHull3D.java
```diff
@@ -0,0 +1,545 @@
...
+public class QuickHull3D extends AbstractFunctionOp<HashSet<Vertex>, DefaultFacets> implements ConvexHull3D {
```
ctrueden: Never type things on `DefaultFoo`. They should be typed on `Foo` interface that backs it. The `Default` class should have _no new public API_ in it—but only implement the methods in the interface.

src/main/java/net/imagej/ops/descriptor3d/Facets.java
```diff
@@ -0,0 +1,5 @@
+package net.imagej.ops.descriptor3d;
+
+public interface Facets {
```
ctrueden: There is a convention in ImgLib2, and other places in SciJava code including ImageJ Common, to use the plural as a utility class for working with the singular. For example, `Intervals` provides handy methods for working with `Interval` objects. In ImageJ Common: `Axes` for `Axis`. I like this convention as it is succinct, and once you know it you can infer where to look for such things for additional class hierarchies.

For that reason, I'd avoid `Facet` and `Facets` in favor of something else (`FacetGroup`? Not sure on the purpose of the interface since it has no public API).

src/main/java/net/imagej/ops/descriptor3d/QuickHull3D.java
ctrueden: Yeah, we eschew [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation) really hard in the SciJava software stack. Please no `m_`.

ctrueden:
First of all, this is frickin' awesome! So fantastic to have in ops! Thank you @tibuch.

Regarding the namespaces...

@dietzc wrote:
> Geometric3D namespace is enough (no descriptor3D required)
> Introduce generic Geometric namespace for n-dimensional implementations.

Personally, I favor a single `geom` (or maybe `geometric`) namespace. A separate namespace per dimensionality (2D, 3D, etc.) gets kinda ugly fast. The 3D-specific op implementations can live in there just fine, can't they? Many of the names imply 3D already (e.g., `volume`). And even if you generalize to N-dimensional, there is no clash where two ops of the same name produce different results ever, is there?

dietzc:
if there would be a clash we still could add `2D` or `3D`. I agree. Let's put everything in `geom()` namespace after this one has been merged. Later, we may want to put the `Feature`s (everything which gives us a number) into a sub-namespace of `feature()`.

src/main/java/net/imagej/ops/descriptor3d/Facets.java
dietzc: I vote for `Mesh`

src/main/java/net/imagej/ops/descriptor3d/AbstractVertexInterpolator.java
tibuch: No, because I need the real coordinates for the marching cube algorithm. Marching cube tries to approximate the the position of the hull of a point cloud. If we have two pixels of which one is clear on the inside (value == 1) and one is clear on the outside (value == 0) we know that the hull is somewhere in between of these two pixels.

tibuch:
Thank you @ctrueden and @dietzc for reviewing the code!

I fixed these general issues:

* [x] rebase master into branch (and resolve conflicts)
* [x] License headers missing
* [x] `m_` for members is not our code style
* [ ] ~~Set special descriptor / geometric 3d classes to package visibility only~~
* [x] `Geometric3D` namespace is enough (no descriptor3D required)
* [x] ~~Introduce generic `Geometric` namespace for n-dimensional implementations.~~ I will do this after this is merged. #222 
* [x] Code cleanup (make parameters in signatures final etc)

dietzc:
@tibuch did you rename the interfaces etc?

tibuch:
> @tibuch did you rename the interfaces etc?

Yes. I renamed Facets to Mesh, FacetInterface to Facet, AbstractPolygon to UpdateablePointSet.
I did this in the code clean up commit. Maybe should have split this commit.

src/main/java/net/imagej/ops/AbstractOpEnvironment.java
```diff
@@ -703,6 +706,16 @@ public ThreadNamespace thread() {
 	public ThresholdNamespace threshold() {
 		return namespace(ThresholdNamespace.class);
 	}
+	
+	@Override
+	public Geometric3DNamespace geometric3d() {
+		return namespace(Geometric3DNamespace.class);
+	}
+	
+	@Override
```
dietzc: does descriptor3d still exist?

src/main/java/net/imagej/ops/OpEnvironment.java
```diff
@@ -39,7 +39,10 @@
 import net.imagej.ops.deconvolve.DeconvolveNamespace;
 import net.imagej.ops.features.haralick.HaralickNamespace;
 import net.imagej.ops.filter.FilterNamespace;
+
 import net.imagej.ops.geometric.Geometric2DNamespace;
+import net.imagej.ops.geometric3d.Descriptor3DNamespace;
```
dietzc: remove descriptor3d?

PR #217, opened 2015-09-22
Math templates

LeonYang5114:
Some templates for the math package are created based on the Add classes.
Meanwhile I did a merge and many files that should not have been changed were removed and re-added. I would like to modify that if I know how to.
Something interesting is that the README.md file in my branch has /r/n as EOL, while the master branch has /n, and git does not replace that for me in merge. Therefore I made a small commit to count for EOL compatibility.


ctrueden:
Moved to #219.

PR #218, opened 2015-09-22
Geometric features

angrauma:
Add implementation and test for geometric features.

@dietzc do you mind taking a look?


dietzc:
Could you use luna to Format the code? I will have a look right afterwards.

dietzc:
look great! just several `final`s are missing (didn't mark all of them) + licenses headers + formatting with luna + the rest I commented on! Then, I think, it's ready to merge.

Btw: We could add special `CenterOfGravity` and `Area` implementations based on `LabelRegion` as both values are precalculated in the `LabelRegion` object.

dietzc:
@ctrueden How to handle the 0.4.1. imglib2-roi dependency?

ctrueden:
@dietzc 
It is fine to put:
```xml
<properties>
  <imglib2-roi.version>0.4.1</imglib2-roi.version>
</properties>
```
In the project's POM temporarily, and merge that to master. In the future, we'll just have to purge that again after bumping the parent to include a newer imglib2-roi.

dietzc:
thanks for all the effort @seebacherd and @angrauma. performance related improvements can be added later.

PR #219, opened 2015-09-22
Math template amend

LeonYang5114:
Add templates for the ops.math package. The previous math-template branch is messed up so I started this new branch with cleaner and clearer commits.


ctrueden:
Since you didn't file a PR for the original `math-template` branch, there is no reason to describe how you amended things. Just describe your PR from the perspective of someone with fresh eyes looking at it, to understand what it does.

ctrueden:
This is looking extremely good. Since you are catching on quickly, I am not holding back any nitpicks. I hope you find it useful rather than discouraging. :grinning:

ctrueden:
We have a [check-branch.sh script](https://github.com/scijava/scijava-scripts/blob/master/check-branch.sh) which you might find useful for verifying that all commits of your branch compile with passing tests.

LeonYang5114:
I decide that the old commits should be squashed into one since I think that's not introducing much confusion. The EOF problem is fixed. The inner classes now have shorter names. The MathNamespace class is now only modified for the addition of the methods (not for reordering). Simple comments for the .list and .vm files have been added.

One thing is that I force the remote branch to update so that I lost all my previous commits, and your commit comments. I am not sure whether it is the right thing to do.

ctrueden:
> I decide that the old commits should be squashed into one since I think that's not introducing much confusion.

The diff is a bit huge now, but I think it's OK. There are ways you could have mitigated the diff (e.g., do preemptive indentation changes before the rename), but they are a bit artificial, and not worth doing another rewrite, since the diff as-is is still pretty readable.

My only remaining beef is your commit message for the main commit. It should be written in imperative tense—something like:
```
Autogenerate math.add classes

See: src/main/templates/net/imagej/ops/math
```
And actually, that illustrates a place where you could split the commit in half: 1) do the conversion to templates, but without mixing in sub/mul/div; then 2) add sub, mul and div to the template. The second commit will be super short, of course, and have a commit message like "Add more sub, mul and div ops".

> One thing is that I force the remote branch to update so that I lost all my previous commits, and your commit comments. I am not sure whether it is the right thing to do.

Those commits are still there if you know where to look, though they are orphaned.

It used to be the case that GitHub would keep the links here in the PR with a note like "so-and-so commented on an obsolete diff", but it seems like that isn't happening anymore?

Either way, I think it is worth the price—ultimately we need the commits to be well structured before they merge. So you did the right thing.

ctrueden:
Perfect! Thank you very much! :confetti_ball:

imagejan:
> It used to be the case that GitHub would keep the links here in the PR with a note like "so-and-so commented on an obsolete diff", but it seems like that isn't happening anymore?

@ctrueden @LeonYang5114 
I think GitHub still keeps the comments when you comment on the *diff* (see #225 for an example), but it doesn't when you commented on the *commit* itself...

Thanks for all this great work!

ctrueden:
@imagejan wrote:
> I think GitHub still keeps the comments when you comment on the *diff* (see #225 for an example), but it doesn't when you commented on the *commit* itself...

Cool, thanks for verifying that. I'll make an effort not to comment on commits then, but only on specific lines of code and on the PR thread.

PR #220, opened 2015-09-22
Add op to generate Poisson noise

imagejan:
This enables adding noise to an image as shown in this groovy snippet:

    // @OpService ops
    // @UIService ui
    // @Dataset inputData

    import net.imglib2.type.numeric.real.DoubleType

    
    // execute the filter.addPoissonNoise op on every pixel of an image
    noiseOp = ops.op("filter.addPoissonNoise", DoubleType.class, DoubleType.class)
    ops.run("map", inputData, inputData, noiseOp)
    ui.show("noisy image", inputData)

### DIFF
--- a/src/main/java/net/imagej/ops/filter/FilterNamespace.java
+++ b/src/main/java/net/imagej/ops/filter/FilterNamespace.java
@@ -83,6 +83,24 @@ public class FilterNamespace extends AbstractNamespace {
 		return result;
 	}
 
+	// -- addPoissonNoise --
+
+	@OpMethod(op = net.imagej.ops.Ops.Filter.AddPoissonNoise.class)
+	public Object addPoissonNoise(final Object... args) {
+		return ops().run(net.imagej.ops.Ops.Filter.AddPoissonNoise.class, args);
+	}
+
+	@OpMethod(op = net.imagej.ops.filter.addPoissonNoise.AddPoissonNoiseRealType.class)
+	public <I extends RealType<I>, O extends RealType<O>> O addPoissonNoise(final O out,
+		final I in, final Random rng)
+	{
+		@SuppressWarnings("unchecked")
+		final O result =
+			(O) ops().run(net.imagej.ops.filter.addPoissonNoise.AddPoissonNoiseRealType.class, out,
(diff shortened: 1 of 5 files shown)


ctrueden:
@imagejan This is so awesome!

Crazy that you found that commit by Leon and cherry picked it, before that associated PR was even filed! It seems that commit is only needed if your `core.autocrlf` setting is not `true` on Windows... But it is a good change regardless.

Thanks very much for this. It looks extremely clean, and seems you crossed the Ts and dotted the Is. I made some nitpicky comments, which can hopefully be addressed with a simple `commit --amend` followed by `push --force` of the branch. And then we can merge it!

imagejan:
@ctrueden Thanks for the comments!
> Elsewhere in ops, we use `private long seed` rather than `Random` instance. Which do you like better? My naive intuition is that the `long` is better because it is easier for other SciJava software to support—even non-Java software which might not like various Java objects.

I'm actually in favor of a `seed` as I've seen elsewhere in the code (in some tests?). I just kept `Random` because it was in the `filter.addNoise` op from where I copied. So I'll change both to `seed` if you agree.
I'll try to amend the commits by tomorrow.

LeeKamentsky:
One trick I've used is to hash the incoming data to initialize the random number generator - the upside is that you get a sequence that's different for every input - if you have a constant seed, then you'll get the same poisson noise added to images of similar size. The downside of course is that hashing the whole thing is a time-consuming operation. And an alternative might be to expose the RNG as a parameter (or the seed). And another alternative would be to use your magic to cache the image hash because it's useful in other contexts, like fingerprinting an image to make sure you have the same one when re-running an analysis.

imagejan:
@ctrueden I amended my commit to fix the javadoc and indentation.

(What still confuses me about the *git force-push* workflow is that your helpful comments on https://github.com/imagejan/imagej-ops/commit/9baf344 are now invisible from this PR)

I'll change the `Random` to `long seed` in a separate commit later.

dietzc:
ready to merge? @ctrueden? @imagejan?

ctrueden:
Yes, thank you! :+1:

PR #221, opened 2015-09-22
Non circulant fixes

bnorthan:
This pull requests implements some fixes to the non-circulant Richardson Lucy algorithm described here:

http://bigwww.epfl.ch/deconvolution/challenge/index.html?p=documentation/theory/richardsonlucy/

It also adds a couple of temporary math utilities (the deconvolution ops should be refactored once all the math functions needed are implemented as Ops). 

Also, we now use the log service to update the iteration number.

### DIFF
--- a/src/main/java/net/imagej/ops/filter/IterativeFFTFilterRAI.java
+++ b/src/main/java/net/imagej/ops/filter/IterativeFFTFilterRAI.java
@@ -141,7 +141,6 @@ public abstract class IterativeFFTFilterRAI<I extends RealType<I>, O extends Rea
 
 		performIterations();
 
-
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


ctrueden:
Thanks @bnorthan!

I did a couple other PRs this morning, so I'm definitely in the nitpick zone right now! I presume that's what you wanted, since you filed a PR rather than just merging.

bnorthan:
Thanks for the feedback @ctrueden.   More to come.  As the core math functions are ready I'll integrate them into my code.

ctrueden:
> As the core math functions are ready I'll integrate them into my code.

PR #219 just merged! This adds subtract, multiply and divide ops on par with what we already had for add.

PR #225, opened 2015-09-25
Local Phansalkar thresholding

stelfrich:
This is an Ops implementation of Phansalkar's local thresholding method as originally implemented by Gabriel Landini in the `Auto_Local_Threshold` plugin of IJ1 (http://fiji.sc/Auto_Local_Threshold#Phansalkar).

Currently, the input is not checked for correct normalization to the [0, 1] range. Is that something that should be added in the Op itself?

### DIFF
--- a/src/main/java/net/imagej/ops/threshold/localPhansalkar/LocalPhansalkar.java
+++ b/src/main/java/net/imagej/ops/threshold/localPhansalkar/LocalPhansalkar.java
@@ -49,6 +49,7 @@ public class LocalPhansalkar<T extends RealType<T>> extends LocalThresholdMethod
 	@Parameter
 	private OpService ops;
 
+	// FIXME: Faster calculation of mean and std-dev
 	private ComputerOp<Iterable<T>, DoubleType> mean;
 	private ComputerOp<Iterable<T>, DoubleType> stdDeviation;
 
--- a/src/main/java/net/imagej/ops/threshold/localPhansalkar/LocalPhansalkar.java
+++ b/src/main/java/net/imagej/ops/threshold/localPhansalkar/LocalPhansalkar.java
@@ -40,10 +40,10 @@ public class LocalPhansalkar<T extends RealType<T>> extends LocalThresholdMethod
 	implements Ops.Threshold.LocalPhansalkar
 {
 
-	@Parameter
+	@Parameter(required = false)
 	private double k = 0.25;
 
-	@Parameter
+	@Parameter(required = false)
 	private double r = 0.5;
 
 	@Parameter
--- a/src/main/java/net/imagej/ops/threshold/localPhansalkar/LocalPhansalkar.java
+++ b/src/main/java/net/imagej/ops/threshold/localPhansalkar/LocalPhansalkar.java
@@ -58,8 +58,8 @@ public class LocalPhansalkar<T extends RealType<T>> extends LocalThresholdMethod
 
 	@Override
 	public void initialize() {
-		mean = ops().computer(Mean.class, new DoubleType(), in().getB());
-		stdDeviation = ops().computer(StdDev.class, new DoubleType(), in().getB());
+		mean = ops().computer(Mean.class, DoubleType.class, in().getB());
+		stdDeviation = ops().computer(StdDev.class, DoubleType.class, in().getB());
 	}
 
 	@Override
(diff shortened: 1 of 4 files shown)


dietzc:
Can you add the method and all related signatures to the `ThresholdNamespace`?

dietzc:
Great! Thanks a lot. Just added some comments in the commits!

src/main/java/net/imagej/ops/threshold/localPhansalkar/LocalPhansalkar.java
```diff
@@ -0,0 +1,78 @@
+package net.imagej.ops.threshold.localPhansalkar;
```
dietzc: Missing license header

src/main/java/net/imagej/ops/threshold/localPhansalkar/LocalPhansalkar.java
```diff
@@ -0,0 +1,78 @@
...
+ * @author Stefan Helfrich <s.helfrich@fz-juelich.de>
+ */
+@Plugin(type = Ops.Threshold.LocalPhansalkar.class, name = Ops.Threshold.LocalPhansalkar.NAME)
+public class LocalPhansalkar<T extends RealType<T>> extends LocalThresholdMethod<T>
+	implements Ops.Threshold.LocalPhansalkar
+{
+
+	@Parameter
+	private double k = 0.25;
+
```
dietzc: required = false?

src/main/java/net/imagej/ops/threshold/localPhansalkar/LocalPhansalkar.java
```diff
@@ -0,0 +1,78 @@
...
+	@Parameter
+	private double r = 0.5;
```
dietzc: required = false?

src/main/java/net/imagej/ops/threshold/localPhansalkar/LocalPhansalkar.java
```diff
@@ -0,0 +1,78 @@
...
+
+	private ComputerOp<Iterable<T>, DoubleType> mean;
+	private ComputerOp<Iterable<T>, DoubleType> stdDeviation;
+
+	private double p = 2.0;
+	private double q = 10.0;
+
+	@Override
+	public void initialize() {
+		mean = ops().computer(Mean.class, new DoubleType(), in().getB());
```
dietzc: you should be able to use `Double.class` instead of `new DoubleType()` here.

src/main/java/net/imagej/ops/threshold/localPhansalkar/LocalPhansalkar.java
```diff
@@ -0,0 +1,78 @@
...
+		stdDeviation = ops().computer(StdDev.class, new DoubleType(), in().getB());
```
dietzc: you should be able to use Double.class instead of new DoubleType() here.

src/main/java/net/imagej/ops/threshold/localPhansalkar/LocalPhansalkar.java
```diff
@@ -0,0 +1,78 @@
...
+	private OpService ops;
+
+	private ComputerOp<Iterable<T>, DoubleType> mean;
+	private ComputerOp<Iterable<T>, DoubleType> stdDeviation;
+
+	private double p = 2.0;
+	private double q = 10.0;
+
+	@Override
+	public void initialize() {
```
dietzc: Can you add a `FIXME:  Faster calculation of mean and std-dev.`. Because we could add an `Op` which calculates both in a single run (and therefore saves runtime).

stelfrich:
Thanks for the comments @dietzc!

> Can you add the method and all related signatures to the ThresholdNamespace (in case you change Parameter to required = false)?

How would I do that properly? I have to optional parameters and the signatures if only one of them is provided would are the same..

dietzc:
> How would I do that properly? I have two optional parameters and the signatures - if only one of them is provided - are the same..  

Good call. I'm actually not sure about that. Maybe different method names? @ctrueden any comments?

ctrueden:
I commented on #233 about it. This branch should be good to go, no? As soon as it's rebased over master to resolve conflicts?

stelfrich:
I have rebased the PR on master. Sorry for the delay!

PR #226, opened 2015-09-25
Local Sauvola thresholding

stelfrich:
This is an Ops implementation of Sauvola's local thresholding method as originally implemented by Gabriel Landini in the Auto_Local_Threshold plugin of IJ1 (http://fiji.sc/Auto_Local_Threshold#Phansalkar).

Open questions:
- Currently, the input is not checked for correct normalization to the [0, 1] range. Is that something that should be added in the Op itself?
- How to properly handle two optional parameters in the (Threshold)Namespace if they cannot be discriminated by type (and hence different method signatures); see https://github.com/imagej/imagej-ops/pull/225#issuecomment-143161603 and following

### DIFF
--- a/src/main/templates/net/imagej/ops/Ops.list
+++ b/src/main/templates/net/imagej/ops/Ops.list
@@ -313,7 +313,6 @@ namespaces = ```
 		[name: "shanbhag",                    iface: "Shanbhag"],
 		[name: "triangle",                    iface: "Triangle"],
 		[name: "yen",                         iface: "Yen"],
-		[name: "sauvola",                     iface: "Sauvola"],
 	]],
 	[name: "zernike", iface: "Zernike", ops: [
 		[name: "magnitude",                iface: "Magnitude"],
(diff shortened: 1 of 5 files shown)


dietzc:
@stelfrich to make progress, I think it would be really helpful if you could open an issue describing the problem you are facing with the parameters. It's better to discuss this with @ctrueden in an issue.

ctrueden:
Well, a PR is an issue. So we don't need a second issue necessarily.
(Depends.) But I am pretty behind on email/issues/PRs/bugs right now.

stelfrich:
@ctrueden Feel free to close https://github.com/imagej/imagej-ops/issues/233 then.

dietzc:
My idea with the issue was, that the problem is pretty general and not only related to thresholding.

stelfrich:
Rebased on master. Sorry for the delay!

dietzc:
now there are some conflicts. potentially because I merged the other branch

stelfrich:
@dietzc The conflicts are fixed.

src/main/templates/net/imagej/ops/Ops.list
```diff
@@ -312,6 +313,7 @@ namespaces = ```
 		[name: "shanbhag",                    iface: "Shanbhag"],
 		[name: "triangle",                    iface: "Triangle"],
 		[name: "yen",                         iface: "Yen"],
+		[name: "sauvola",                     iface: "Sauvola"],
```
dietzc: do we want this, too? Whats the difference to `localSauvola`?

PR #230, opened 2015-09-27
Add fusion op to fuse (overlapping) RAI's

gab1one:
Provides two fusion methods:
- Fuse max (the larger pixel value is used, images are extended with zero)
- Fuse min (the smaller pixel value is used, images are extended with max value)

Images can be arranged arbitrary, the fusion type is for overlapping images and describes which pixel is used. 

Currently only works on 2d images, but could be extended to more.
Created by @EikeHeinz  and me.

### DIFF
--- a/src/main/java/net/imagej/ops/image/ImageNamespace.java
+++ b/src/main/java/net/imagej/ops/image/ImageNamespace.java
@@ -184,6 +184,27 @@ public class ImageNamespace extends AbstractNamespace {
 				net.imagej.ops.image.equation.DefaultEquation.class, out, in);
 		return result;
 	}
+	
+	// -- fusion --
+	
+	@OpMethod(op = net.imagej.ops.image.fusion.MinFusion.class)
+	public <T extends RealType<T>> RandomAccessibleInterval<T> fuseMin(final RandomAccessibleInterval<T> in,
+			final RandomAccessibleInterval<T> in2, final long... offset) {
+		@SuppressWarnings("unchecked")
+		final RandomAccessibleInterval<T> result =
+			(RandomAccessibleInterval<T>) ops().run(net.imagej.ops.image.fusion.MinFusion.class, in, in2, offset);
+		return result;
+	}
+	
+	
+	@OpMethod(op = net.imagej.ops.image.fusion.MaxFusion.class)
+	public <T extends RealType<T>> RandomAccessibleInterval<T> fuseMax(final RandomAccessibleInterval<T> in, 
+			final RandomAccessibleInterval<T> in2, final long... offset) {
+		@SuppressWarnings("unchecked")
+		final RandomAccessibleInterval<T> result =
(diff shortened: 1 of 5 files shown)


dietzc:
What about overlapping `Interval`s which are not only translated? I would suggest, that the Fusion method takes `n` arbitrary `RandomAccessibleInterval`s  which potentially may intersect. The intersection is then handled somehow (min, max, avg, median, whatever `ComputerOp` or `Converter`).

gab1one:
Right now the translation from image1 to image2 is given as an argument by the user, 
how should that work in the ``n RAI`` case?

dietzc:
The assumption is, that the input is already tranformed AS needed. There is no reason that the translation is part of the fusion, right?

gab1one:
currently the op works by creating an output image based on the translation, 
then virtually translates the second image and extends both images, so that it then can iterate over
the result image and fusing the pixels from the RAI's. 
Thus the translation is currently required to avoid out of bounds problems. It is used to locate the pixels in the source image corresponding to each pixel in the output.

One could split out the explicit fusion part from the rest though.

dietzc:
> One could split out the explicit fusion part from the rest though, that op would require all input RAIs to be of the same size though.

Why? You can always calculate the resulting  `Interval` extend by adding up the intervals of the individual `RAI`s. I just wouldn't restrict the applied transformation to be a translation.

gab1one:
You want to input eg. an ``AffineGet``? instead of just an translation?

dietzc:
possibly. But I still wonder why the fusion should know that the `RandomAccessibleInterval` is somehow transformed. This can be done outside, right?

gab1one:
Yes, but the op needs to know which pixel in the  input ``RandomAccessibleInterval``s  correspond to each pixel in the output IMG, so that it can calculate the fusion.
If it does not know the transformation, it must assume the images have the same size and directly overlay them.
We could iterate over both input's at the same time, and output the fusion to an iterator over the result image:
````java
while(outcursor.hasNext()){
    out.fwd().get() = fuse(cursor1.fwd().get(), cursor2.fwd().get());
}
````
But that does not look safe to me...

ctrueden:
@gab1one Are the next steps clear to you for this? Or do you need further guidance?

gab1one:
I think we should talk about this next week in person @ctrueden. I think the goals and parameters are not  clear enough to me.

ctrueden:
@gab1one Sorry we never got around to discussing this. Can you discuss with @dietzc when you have time?

ctrueden:
@gab1one Let's discuss in person this time. I have ideas.

ctrueden:
We hacked on something more general. As of this writing, you can see it here: ctrueden/sandbox@b551811277bf5898d2bf838abb92959bc3385de6

fusion-sexiness

Closing this PR in favor of the idea above. @tpietzsch Any interest in adding the static `fuse` method above to `Views`?

PR #235, opened 2015-10-04
Add CopyNamespace

dietzc:
This PR replaces the work of https://github.com/imagej/imagej-ops/pull/148 (as this branch was a mess). We implemented several `Copy` methods for several types. @hinerm @ctrueden would be great if you could review/comment.

### DIFF
--- a/src/main/java/net/imagej/ops/AbstractOpEnvironment.java
+++ b/src/main/java/net/imagej/ops/AbstractOpEnvironment.java
@@ -38,6 +38,7 @@ import java.util.List;
 import java.util.Map;
 
 import net.imagej.ops.convert.ConvertNamespace;
+import net.imagej.ops.copy.CopyNamespace;
 import net.imagej.ops.create.CreateNamespace;
 import net.imagej.ops.deconvolve.DeconvolveNamespace;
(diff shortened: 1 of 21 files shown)


dietzc:
@ctrueden I decided not to add separate `ComputerOp` or `FunctionOp` for `CopyImg`, as this `CopyImg` just delegates to other ops anyway. However, we could change all `Copy` ops to be pure `ComputerOp`s and only the `CopyImg` is a `HybridOp` (as it is the only one which actually knows how to create an output). What I don't like about this idea is, that someone might want to use copy on lets say an arbitrary `RandomAccessibleInterval` and get back some `RandomAccessibleInterval` without having to decide how it actually looks like internally (convenience reasons). We would lose this functionality if we would e.g. change `CopyRAI` to be a pure `ComputerOp`. 

What do you think?

ctrueden:
@dietzc Sure, I'm OK with the op structure as written! I made a few other nitpicky comments. But in general this looks super great!

ctrueden:
I did a few cleanups. There is still more that could be done, but in the interest of getting this onto master now, I'm merging it.

Issue #234, opened 2015-10-01
Speed up the Geometric3DFeaturesTest


ctrueden:
On my system, the master branch of imagej-ops now builds in 77 seconds. With `Geometric3DFeaturesTest` disabled, it builds in 28 seconds. So this test costs nearly a minute of computation to run.

We need to keep the unit tests of ops as fast as possible. I often [build all commits of a topic branch](https://github.com/scijava/scijava-scripts/blob/d892adc0092c220ee1e597b9fb5a1fb067e4509b/check-branch.sh) before merging, which can take 10-20 minutes or more, depending on the number of commits—and now takes nearly 3x as long as before.

Would it be possible to slim down the `Geometric3DFeaturesTest` to run on smaller inputs?

If not, we could flag the tests as "expensive" by respecting the value of `imagej.ops.expensive.tests` (see [here](https://github.com/imagej/imagej-ops/blob/imagej-ops-0.20.0/src/test/java/net/imagej/ops/thread/ChunkerBenchmarkTest.java#L55) for an example).

dietzc:
@ctrueden is this ok for you now?

ctrueden:
It is, now that I fixed the compile error on master (d9cb54cf28c3d6227f95b9687ccf8d66fc897ac3). :wink:

ctrueden:
Thanks guys!

PR #237, opened 2015-10-05
Reuse vertex interpolator

tibuch:
Hi @dietzc and @ctrueden 

I fixed the MarchingCubes implementation. Now only one instance of the vertex interpolator is created and the algorithm runs 3x faster.

And I replaced the test label with a smaller one ( #234 ).


dietzc:
great. Thanks!

ctrueden:
Thanks @tibuch!

PR #239, opened 2015-10-05
Add tamura 2d features

dietzc:
thanks to @angrauma for the implementation of tamura 2d features

@ctrueden if you don't see something obvious, I'm happy to merge.


ctrueden:
:+1:

PR #242, opened 2015-10-07
RAIs: add null checks to in()

dietzc:
If `in()` equals null, we may want to pass at least the minimum required class definition (which is `RAI`). @ctrueden does this make sense?


ctrueden:
There was some weird style reformatting mixed into this commit, so I fixed it. Otherwise, great!

PR #243, opened 2015-10-07
CachedOpEnvironment: CachedFunctionOp extends AbstractOp

dietzc:
see: https://github.com/imagej/imagej-ops/commit/84a25cba0c67bb9e87b123777e6af55c438ce425#commitcomment-13598119


ctrueden:
Beautiful!

PR #246, opened 2015-10-09
Issue222 geom namespace

tibuch:
Hi @ctrueden and @dietzc 

I merged geometric2D and geometric3D into geom. 

I created the interface Polytope with the method getPoints(). Polytope is implemented by ThePolygon (only a new class which extends imglib2-roi Polygon and Polytope) and Mesh. With Polytope I managed to simplify some features like Solidity. 
One problem is that I can't rasterize a Mesh. This means that some features still need a specific implementation for 2D and 3D. 
As far as I am right now everything works but one can still improve some implementations and probably remove some specific 2D/3D implementations. I would suggest to merge this and add other improvements with different PRs. 

It would be nice if Polytope and Mesh are part of imglib2-roi. But I guess we first need a Mesh-contains implementation.


dietzc:
Hi @tibuch thanks for the effort!! I will review everything over the weekend!

dietzc:
I alao refactored the `geom` package a bit. It looks good to me now. If @ctrueden and @tibuch arw satisfied, I think we can merge!

tibuch:
Hi @dietzc thank you for reviewing and refactoring! It looks really nice and I learned some stuff again.

I am not 100% sure about the CentroidMesh.class. Apart from this I think we can merge.

dietzc:
Does this http://wwwf.imperial.ac.uk/~rn/centroid.pdf help for the mesh centroid calculation?

PR #248, opened 2015-10-09
Create math assign

LeonYang5114:
Assign sets the value of the output to a constant and ignore the input. It is added to the RealMath and the auto generated math classes, which make assigning constant values to image easier.

### DIFF
--- a/src/main/java/net/imagej/ops/math/MathNamespace.java
+++ b/src/main/java/net/imagej/ops/math/MathNamespace.java
@@ -7,13 +7,13 @@
  * %%
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
- * 
+ *
  * 1. Redistributions of source code must retain the above copyright notice,
  *    this list of conditions and the following disclaimer.
  * 2. Redistributions in binary form must reproduce the above copyright notice,
  *    this list of conditions and the following disclaimer in the documentation
  *    and/or other materials provided with the distribution.
- * 
+ *
  * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
  * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
  * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
@@ -29,6 +29,8 @@
  */
 
 package net.imagej.ops.math;
+
+import org.scijava.plugin.Plugin;
 
 import net.imagej.ops.AbstractNamespace;
 import net.imagej.ops.Namespace;
@@ -47,8 +49,6 @@ import net.imglib2.type.numeric.RealType;
 import net.imglib2.type.numeric.integer.ByteType;
 import net.imglib2.type.numeric.real.DoubleType;
 
-import org.scijava.plugin.Plugin;
-
 /**
  * The math namespace contains arithmetic operations.
  *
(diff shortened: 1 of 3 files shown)


ctrueden:
Nice initial version. However, I do not think we want any signatures that "ignore the input."

We want the following:
* `AssignTypeToType` – `assign(T out, T in)` – calls `out.set(in)` to copy the value from `in` to `out`.
* `AssignTypeToIterable` – `assign(Iterable<T> arg, T value)` – assigns the value of `value` to all elements of `iterable`, by leaning on an inplace `map` call with an `AssignType` op. (This could potentially be generalized even further than that, actually, thanks to the matching framework. `T` might not have to be restricted to `Type<T>` for this one.)
* `assign(Iterable<T> out, Iterable<T> in)` – iterates over both things in parallel, assigning the corresponding value of `in` to that of `out`. I am not sure we want to create this right away. Maybe we should generalize it further.

@dietzc Your thoughts here? Maybe what we are trying to do is redundant with existing ops? But we couldn't find suitable ones. The main use case is to __set an image or region to a constant value__. This came up when @LeonYang5114 and I thought about how best to migrate the [NanBackground](https://github.com/imagej/imagej-plugins-commands/blob/imagej-plugins-commands-0.6.0/src/main/java/net/imagej/plugins/commands/imglib/NanBackground.java) command to Ops.

dietzc:
It seems that `assign(Iterable<T> out, Iterable<T> in)`  can be implemented with `Map` and is something we could use for the `CopyComputerOp`? https://github.com/imagej/imagej-ops/tree/copy

`assign(T out, T in)` can be `assign(T out, V in)` with `assign(T out, T in)` as special case implementation? This somehow reminds me of  `convert`  (with same types) .

`assign(Iterable<T> arg, T value)` absolutely makes sense.


`

ctrueden:
> `assign(T out, T in)` can be `assign(T out, V in)` with `assign(T out, T in)` as special case implementation? This somehow reminds me of  `convert`  (with same types) .

Yeah, we discussed the `CopyRealTypes` op of the `convert` namespace, which does exactly that (typed on separate `I extends RealType<I>` and `O extends RealType<O>`). The problem is that it is lossy because it [goes through `double`](https://github.com/imagej/imagej-ops/blob/e660c34ca04a727d2f96d1ec662a685201f9f9f7/src/main/java/net/imagej/ops/convert/copy/CopyRealTypes.java#L50). So we definitely need one that is typed on the same `T extends Type<T>` for both input and output, so the implementation can be simply `output.set(input)`.

> is something we could use for the `CopyComputerOp`?

Yeah, the overlap here is big. Maybe we shouldn't call it `assign` in the case of images, since that might imply a shallow variable assignment, when what is actually happening is a per-pixel value assignment.

@LeonYang5114 and I considered calling this op `set`, but decided on `assign` to avoid terminology confusion—we thought the `math` namespace would be a good place for `assign`, and `math.set` sounds like [math sets](https://en.wikipedia.org/wiki/Set_(mathematics))).

So in summary!

```java
public class AssignTypeToType<T extends Type<T>> extends AbstractComputerOp<T, T> implements Ops.Math.Assign {
	@Override
	public void compute(final T in, final T out) {
		out.set(in);
	}
}

public class AssignTypeOverIterable<T extends Type<T>> extends AbstractInplaceOp<Iterable<T>> implements Ops.Math.Assign {
	@Parameter
	private T value;

	@Override
	public void compute(final Iterable<T> arg) {
		for (final T t : arg) {
			t.set(value);
		}
	}
}

// Or we could get fancy: extend HighLevelHybridOp, initialize a map, and call it:

public class AssignTypeOverIterable<T extends Type<T>> extends HighLevelInplaceOp<Iterable<T>> implements Ops.Math.Assign {
	@Parameter
	private T value;

	@Override
	protected InplaceOp<Iterable<T>> createWorker(final Iterable<T> t) {
		final T type = Util.getTypeFromInterval(t);
		return (InplaceOp) ops.inplace(Ops.Map.class, t, //
			ops().computer(Ops.Math.Assign.class, type, value));
	}
}
```

And we'll hold off on an image-to-image one, in favor of @dietzc's `copy` branch—which will likely end up reusing the `math.assign` ops to do their work.

We should also hold off on assign `HybridOp` implementations, even though it will make `assign` less consistent with the other primary math ops (`add`, `subtract`, `multiply` and `divide`), but 

_There was something else I was going to write here, but that was two days ago now and I forgot. So in the interest of submitting this comment, let this serve as a caveat that everything I wrote above is only 80% thought through._

dietzc:
Just to make sure: `Assign` implementations will base on `Map`, right?

LeonYang5114:
Hi @ctrueden @dietzc I have created the two classes based on Curtis' code. I have a few questions.

1. Should assign lives in math package? Op that implements Math interfaces are all ComputerOp, and all of them handles only RealType. The AssignTypeOverIterable is logically an InplaceOp, and the two assign Op are general enough to handle Type. I think it might not be very proper to let assign live inside math.
2.  <code>final T type = Util.getTypeFromInterval(t);</code>
This line in Curtis' code does not work for t as an Itervable, since that method also requires t as a <a href="https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/util/Util.java#L751">RandomAccessable</a>. I instead use (T) null to indicate the type. Will this work correctly?
3. I have removed the Assign classes in the image-related templates (e.g. ConstantToArrayImage.Assign). We should implement these functionalities in the CopyComputerOp branch, is that what you mean? or should I add those classes back?

Thanks!

LeonYang5114:
Sorry I did not read the copy package until now. I think the two classes above have too much overlap with the <a href="https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/copy/CopyType.java">CopyType</a> and the <a href="https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/copy/CopyIterableInterval.java">CopyIterableInterval</a>. And I think the copy package should be a better place than the math.assign.

But I also think that it might be quite handy to have a lossy assign in the RealMath which ignores the input and do a .setReal(value) to the output. What do you think? @ctrueden

ctrueden:
@LeonYang5114 Ah yeah. The copy namespace actually wasn't merged until recently. You're right that `CopyType` and `CopyIterableInterval` are exactly the same.

I still don't understand why we would want a "lossy assign" though?

LeonYang5114:
@ctrueden I think a lossy assign might be convenient, like this one in the deprecated <a src="https://github.com/imagej/imagej-deprecated/blob/master/src/main/java/net/imglib2/ops/operation/real/unary/RealConstant.java">RealConstant</a>. Otherwise I think the only way to achieve this right now is to use Converter in imglib2, which is not an Op, and the logic is not quite "mathematic."

LeonYang5114:
@ctrueden Another reason is that, we already have the <a href="https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/math/RealMath.java#L1103">RealMath.Zero</a>, why not generalize it to set the real part of the output to be any double we passed as parameter?

PR #249, opened 2015-10-11
Infer name and aliases from interface constants

### LINKED ISSUE #122: Make declaration of OPS names simpler
Right now, when implementing a core op, you have to do two things:
* Write `name=Ops.Foo.Bar.NAME` for the name in the `@Plugin` declaration
* Write `implements Ops.Foo.Bar` in the class declaration.

These two things are redundant, and like all duplicate data structure sorts of issues in programming, should be avoided. As things stand, writing `ops.run(Ops.Foo.Bar, ...)` and `ops.run(Ops.Foo.Bar.NAME, ...)` might have different results (due to ops that forgot to do one of the other of the above things), which could end up being very surprising.

One way to achieve unification here would be to make the `name=asdf` declarations unnecessary by inferring from the implemented interfaces. That is: if name is unset, look at the interfaces implemented by the class, and look for `NAME` constants declared in them.

There is one problem: we would need to load the class and supertypes in order to inspect the implemented interfaces—which slows down plugin discovery substantially. Although this would be done lazily when `getName` is first called on the `PluginInfo`, I suspect that that method is called early somewhere in many cases—e.g., when `OpService#getOps()` is called, it needs to know all names. I think right now, the framework doesn't need to load all the op implementation classes to do that, but it would if we made this change.

A good way around the above problem would be to subtype each op—i.e., write `@Plugin(type = Ops.FooBar.class)` instead of just `@Plugin(type = Op.class)`. Then look for the `NAME` constant specifically on the type of plugin being declared. That would be pretty slick, as it would limit upfront class loading to only the base interfaces.

ctrueden:
Fixes #122.


ctrueden:
@dietzc I am fairly confident these are good changes, but feel free to take a quick look before merging if you like.

dietzc:
seems that our `geom` changes I just merged conflict with this branch. sry.

ctrueden:
Rebased and fixed conflicts. Merging before any further conflicts arise.

dietzc:
:+1: thanks curts.

PR #260, opened 2015-10-13
IterableInterval Util class

zjpetersen:
Added a utility class 'IIs.java' for working with IterableIntervals.


ctrueden:
@zjpetersen Let's go through this PR in person tomorrow afternoon!

ctrueden:
Unfortunately, there are type safety problems with the current `RAIs`/`IIs` helper classes. I am thinking about how best to resolve them. The gist is that we cannot actually guarantee generified type parameters of e.g. `RandomAccessibleInterval<T>` without that `T` being specified as an input somehow—also passed to the ops matcher as a concrete object of some sort.

PR #262, opened 2015-10-15
add local binary pattern 2d feature

angrauma:
Add the implementation and test of local binary pattern in 2d. 
Implementation is only validated by hand using a small toy example.

@dietzc can you please have a look.


dietzc:
@tibuch can you review?

seebacherd:
Looks good to me, can be merged.

PR #263, opened 2015-10-16
Create Conditional Op

LeonYang5114:
This Op works as a conditional operator. 

Not sure do we have other Ops that do the same thing, but since this is quick to implement, I just code it up to make things go faster if we really need it.

Also I don't think that the name is good, but for now I cannot come up with a better one. Please nitpick :-)

### DIFF
--- a/src/main/java/net/imagej/ops/Conditional.java
+++ b/src/main/java/net/imagej/ops/Conditional.java
@@ -1,0 +1,62 @@
+/*
+ * #%L
+ * ImageJ software for multidimensional image processing and analysis.
+ * %%
+ * Copyright (C) 2014 - 2015 Board of Regents of the University of
+ * Wisconsin-Madison, University of Konstanz and Brian Northan.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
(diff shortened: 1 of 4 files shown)


ctrueden:
It seems like this op should be named `if`. The class you created might be `IfBoolType`. The `trueValue` and `falseValue` need to be parameters to the op, and there should be no public constructor. And it needs a `@Plugin(type = Ops.If.class)` annotation.

Should there be another class `IfCondition` that takes a `Condition<T>` as input instead of a `BoolType`? Or is that overcomplicating things?

We should also decide whether to type the boolean one on `BoolType` or `Boolean`. Since it's the input, the object does not need to be mutable, so `Boolean` would be OK and not require the ImgLib2 library—hence allowing it to be later moved to a more general `scijava-ops` layer at a future point. And we have (or will have) a SciJava `Converter` implementation to go from `BoolType` to `Boolean`, so we won't have to create the `BoolType` version separately.

@dietzc Does this all make sense?

dietzc:
it does. I just try to find an use-case for this `Op`? Not saying I don't like it. I just don't understand it :-)

ctrueden:
@dietzc At the very least, so you can write `if(condition, t, f)` statements as part of an `eval`?

dietzc:
Which absolutely makes sense. Didn't think about `eval` at all. Thanks @ctrueden !

LeonYang5114:
@ctrueden @dietzc I came up with this Op when I was working on the NanBackground class. With this Op and a Map, we can modify an Image given its threshold output. 
Using Boolean as input type definitely make sense to me. And with the IfCondition, we can do some simple thresholding + modification in just one Map call. That might also be convenient.

LeonYang5114:
@ctrueden Java does not allow "if" as a package name, so I put the IfBoolean and IfCondition in a package called "conditional." Other things are just as you specified. Is there anything I should change?

LeonYang5114:
A more up-to-date and cleaner branch is created. This branch is now closed.

PR #264, opened 2015-10-18
CachedOpEnvironment: use hash object instead of long

dietzc:
Avoids auto-boxing when passed to cache, which confuses several cache
implementation (e.g. guava). Guava cache didn't work with passed primitive `long`.

@ctrueden you think thats OK?

### DIFF
--- a/src/main/java/net/imagej/ops/cached/CachedOpEnvironment.java
+++ b/src/main/java/net/imagej/ops/cached/CachedOpEnvironment.java
@@ -109,23 +109,6 @@ public class CachedOpEnvironment extends CustomOpEnvironment {
 	}
 
 	/**
-	 * Calculates hash given input, {@link FunctionOp} and args
-	 */
-	private <I, O> long hash(final I input, final FunctionOp<I, O> delegate,
-		final Object[] args)
-	{
-		
-		long hash = input.hashCode() ^ delegate.getClass().getSimpleName()
-			.hashCode();
-
-		for (final Object o : args) {
-			hash ^= o.hashCode();
-		}
-
-		return hash;
-	}
-	
-	/**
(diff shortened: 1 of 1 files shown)


src/main/java/net/imagej/ops/cached/CachedOpEnvironment.java
```diff
@@ -244,10 +228,42 @@ public O createOutput(I input) {
...
-		
+
 		@Override
 		public CachedHybridOp<I, O> getIndependentInstance() {
 			return this;
 		}
 	}
+
+	/**
+	 * Simple helper class to generate an object which
```
ctrueden: Unfinished sentence.

src/main/java/net/imagej/ops/cached/CachedOpEnvironment.java
```diff
@@ -53,9 +53,8 @@
 
 	@Parameter
 	private CacheService cs;
-	
-	public CachedOpEnvironment(final OpEnvironment parent)
-	{
+
+	public CachedOpEnvironment(final OpEnvironment parent) {
```
ctrueden: Style changes are mixed in.

src/main/java/net/imagej/ops/cached/CachedOpEnvironment.java
dietzc: oh wow... I will fix that

ctrueden:
Seems OK to me!

dietzc:
I will get rid of the style changes (thank you mars...grrr...) and merge afterwards. thanks!

dietzc:
@ctrueden if you like it, you can merge it. Thanks!

PR #267, opened 2015-10-19
LocalBinaryPatterns: Code-Formatting

dietzc:
@ctrueden simply formatting the code with imagej-ops. Can you quickly have a look if you see something weird? if not, mars1 seems to have fixed the formatter problems!


ctrueden:
Looks OK to me. There are still some formatting issues not caught by the cleanup command though, such as:
* `final private` instead of `private final`
* use of `m_` prefixes
* use of fully qualified class names such as `java.util.Iterator` instead of import statements

PR #272, opened 2015-10-23
fix calculation of width/height in zernike computer

dietzc:
@angrauma doesn't this make more sense? I don't understand what you do with  `Dimension-min`?


angrauma:
@dietzc yes, this makes more sense! Thanks!

dietzc:
@LeeKamentsky do you also agree?

dietzc:
@LeeKamentsky any comment? :-)

PR #276, opened 2015-11-06
Fix copyimg bug

LeonYang5114:
@ctrueden @dietzc There is a small bug in the CopyImg Op. Its compute() method copy the output into the input, Since the unit test only checks the equality between the latest copy of the input and output, so the bug is not detected. Now I modify the unit test so that I first copy the input by iteration, and then compare the output to the copy-by-iteration input.

### DIFF
--- a/src/main/java/net/imagej/ops/copy/CopyImg.java
+++ b/src/main/java/net/imagej/ops/copy/CopyImg.java
@@ -76,7 +76,7 @@ public class CopyImg<T extends NativeType<T>> extends
 
 	@Override
 	public void compute(final Img<T> input, final Img<T> output) {
-		copyComputer.compute(output, input);
+		copyComputer.compute(input, output);
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


dietzc:
thanks for the fix @LeonYang5114. It would be great if you can separate format commits from actual changes in the code. This makes reviewing the code much easier for everybody. After you fixed it in this PR I will review an merge.

Thanks!!

Christian

LeonYang5114:
@dietzc Hi the format commit is separated from the actual changes now.

I wonder is there any easy way to handle the auto format? like automatically separate the formatting from the actual work? Sometimes I just can't help using ctrl+shift+F when I am coding.

dietzc:
@LeonYang5114 there is no automatic way to handle auto-format. basically you have to hope that the one who committed the class has the same auto-formatter as you do (the imagej format style). if the file was committed in a wrong format, you should first make a format-commit. Switching off your eclipse save actions might also help.

anyway, thanks for your changes!

PR #277, opened 2015-11-09
Hyperslice has a type now

dietzc:
this is useful if you want to ensure a certain type on the object which is sliced. makes code much cleaner and correct.


dietzc:
@ctrueden if you could quickly review.

ctrueden:
:yum:

PR #278, opened 2015-11-10
Remove unnecessary parameter of OpService

LeonYang5114:
This commit removes all the parameter of OpService that can be replaced by calling ops() instead. The import is modified accordingly in each class.


ctrueden:
Wow, I thought I had caught all of these when I did the big changeover to `OpEnvironment`. Guess I missed quite a few... or maybe they got introduced in pending PRs that were since merged. Either way: thanks!

PR #279, opened 2015-11-13
Add tree-based viewer for available ops

hinerm:


hinerm:
@ctrueden let me know if you would prefer this was in a different component, more modular, etc...

hinerm:
moving to imagej-ui-swing

PR #280, opened 2015-11-19
Add Stats.Percentile Op

imagejan:
I added a `stats.percentile` op, modeled after `stats.median`.

Interestingly, `Ops.list` already contained entries for the `Ops.Stats.Percentile` and `Ops.Stats.Quantile` interfaces. To complete the picture, we could add a `stats.quantile` op as well.

To stay DRY, we would need to base the Median op onto Percentile (with `percent=50`), but I am unsure how to best do this. Comments are welcome.


ctrueden:
Awesome work, @imagejan. Thank you. DRYing up the code would be a great hackathon topic.

PR #281, opened 2015-11-25
Test op interface

LeonYang5114:
All missing marker interfaces are added, and a test for Op interface is created.

In the OpInterfaceTest, I am not quite sure how to handle the potential InstantiableException.


ctrueden:
Awesome, thank you.

PR #283, opened 2015-12-01
Rename MapIterableToRAIParallel

LeonYang5114:
One of this Map's input type is actually IterableInterval instead of
Iterable, so its name is changed to MapIterableIntervalToRAIParallel.


ctrueden:
:+1:

PR #286, opened 2015-12-02
Change the input type of the BinaryOp as ComputerOp

LeonYang5114:
This is just for PR. I will try to develop based on this commit and see if everything works as expected.

The previous input type of BinaryOp as ComputerOp is BinaryInput, which
exposes the "Binariness" of the Op when it works as a normal, or unary
ComputerOp. It enforces the user to create some kind of BinaryInput in
order to use it as ComputerOp.

As we want the BinaryOp to be a generalization of ComputerOp, we should
make it compatible with its unary version, e.g. having the intput1 as
the input, and input2 as the constant parameter.


LeonYang5114:
see https://github.com/imagej/imagej-ops/commit/a190e0778c367651547a38471974e381599fee0a

PR #287, opened 2015-12-02
Array test

bnorthan:
This pull request:

Adds a simple test that explicitly adds a constant to an array.  This is done to verify that ```ConstantToArrayImage``` is as fast as we expect. 

Changes ```AbstractOpTest``` to explicitly return array images when generating byte test images (the test images were allready ```ArrayImg```, just not returned as such).  In the future we can add functions to return ```PlanarImg``` and ```CellImg``` in order to explicitly run tests on these types.


LeonYang5114:
I think the ability to generate different types of images makes sense in AbstractOp. I am not sure doing an explicit iteration on the array is necessary, but probably that's good since we are in benchmark. @ctrueden what do you think?

PR #289, opened 2015-12-03
add hole detection and extraction op

seebacherd:
add two new ops. one detects if an object in a binary image has holes, the other extracts this holes from the object in the binary image.
also added testcases and a test image.

Ops as requested by @dietzc.


dietzc:
awesome. thank you!! I added some comments in your code. Overall it looks great. Are there variants for 3D available?

seebacherd:
@dietzc That really depends on what you define as a hole in 3D. Do you mean holes in the surface or closed holes inside the object?

dietzc:
second one :-)

seebacherd:
Okay, I will take a look at that problem. I'll let you know as soon as I found something.

tibuch:
@seebacherd I did some testing and found a case where holes are detected.

extract_holes

This is the test image:
test

imagejan:
@tibuch What's detected in your example I wouldn't have classified as "holes". That's rather "cavities" or "dents"?
If that's what this Op is supposed to do, I recommend renaming the op to `DetectCavities` to avoid any confusion with holes (like it's also used in ImageJ's [Analyze Particles](https://github.com/imagej/ImageJA/blob/ff6a0883fbcfa007355941c34c9e9b1ff286086b/src/main/java/ij/plugin/filter/ParticleAnalyzer.java#L63) plugin)

tibuch:
@imagejan I agree with you. 

This example should show that some miss classifications could happen.

imagejan:
@tibuch ah, alright, thanks for clarifying :smile:

dietzc:
@seebacherd any comment?

seebacherd:
Yes, the problem is that this implementation is a heuristic. I'm trying to find holes in one iteration over the image, which leads to the problem that for example a cavity is mistakenly classified as a hole. To get exact results we could use the already implemented connected component analysis, but that would take more time.

dietzc:
is there an exact algorithm to determine holes in objects?

seebacherd:
Okay I've reimplemented the algorithms, not it is a little bit more elaborate, but the results look quite promising. For @tibuch test image we get the following results:

tibuch_results

For an image with actual holes we get the following results:

img_with_holes_result

dietzc:
@ctrueden I'm not sure if I like the name `BooleanTypeNamespace`. But `BinaryNamespace` also feels strange. Any ideas?

ctrueden:
> I'm not sure if I like the name BooleanTypeNamespace. But BinaryNamespace also feels strange. Any ideas?

morphology

imagejan:
> morphology

Note that there's also [Grayscale Morphology](https://en.wikipedia.org/wiki/Mathematical_morphology#Grayscale_morphology), though (and a very good Fiji plugin: [Gray Morphology](http://imagej.net/Gray_Morphology)). And the Maximum and Minimum image filters would belong into that family as well...

How about `BinaryMorphologyNamespace` ?

dietzc:
Wanted to mention the same as @imagejan. Nice suggestion.

Side discussion: In the long run we could also get rid of the `image` namespace, as I think that we more and more tend to group things by functionality rather than by input type.

ctrueden:
Is it really that bad to put all morphology ops (binary and grayscale) into the same namespace though?

tinevez:
@ctrueden I don't think so. In theory operations are the same, just the operands change between binary and grayscale. Plus we can have morphology operations on Strings.

ctrueden:
> Plus we can have morphology operations on Strings.

:+1: ! :grin:

imagejan:
I agree. If all those ops can go in this name space, `ops.morphology` is a great name :smile:

dietzc:
I updated the PR and cleanup the implementations a bit. If someone could quickly review, this would be great.

ctrueden:
I am cleaning this up now.

PR #290, opened 2015-12-04
Enhance logic

LeonYang5114:
Create the Conditional interface and two of its implements, If and Default. They might be useful when working with thresholded images and using eval Op.

See discussion in #263

### DIFF
--- a/src/main/java/net/imagej/ops/logic/Condition.java
+++ b/src/main/java/net/imagej/ops/logic/Condition.java
@@ -30,16 +30,13 @@
 
 package net.imagej.ops.logic;
 
-import net.imagej.ops.special.computer.UnaryComputerOp;
-import net.imglib2.type.logic.BoolType;
-
 /**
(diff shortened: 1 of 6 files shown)


ctrueden:
@LeonYang5114 Can you please rebase this over the latest master so that it can merge?

LeonYang5114:
@ctrueden Thanks for the reminder

LeonYang5114:
@ctrueden this branch is now up-to-date. However, I am not able to improve other condition ops, as they are actually not "operator"s, but building blocks of the conditions. Tell me if you have any idea.

ctrueden:
> Tell me if you have any idea.

Probably we should make the package as a whole less convoluted. We need `greaterThan`, `lessThan`, etc. ops that actually work intuitively, after all.

Also, the `conditional` op should declare `if` as an alias, so that you can write e.g. `if(equal(a, b), c, d)` in an `eval` statement. This is a simple change to `Ops.list`.

PR #291, opened 2015-12-04
Modify fft test

LeonYang5114:
Right now the FFTTest is too slow (around 15 second on my laptop). I change the iteration to testing only two edge cases. Not sure if this modification makes the test much less powerful, but now the test takes only 2.5 seconds, which is acceptable.

### DIFF
--- a/src/test/java/net/imagej/ops/filter/FFTTest.java
+++ b/src/test/java/net/imagej/ops/filter/FFTTest.java
@@ -60,8 +60,8 @@ public class FFTTest extends AbstractOpTest {
 	 */
 	@Test
 	public void testFFT3DOp() {
-		final int min = expensiveTestsEnabled ? 115 : 15;
-		final int max = expensiveTestsEnabled ? 120 : 20;
+		final int min = expensiveTestsEnabled ? 115 : 9;
+		final int max = expensiveTestsEnabled ? 120 : 11;
 		for (int i = min; i < max; i++) {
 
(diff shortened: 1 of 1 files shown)


LeonYang5114:
>@LeonYang5114 we need a blank image here. generateFloatArrayTestImg creates a random image. The idea is that we create a blank image, then add a sphere, to form a test image with an object in it.

I think [generateFloatArrayTestImg](https://github.com/imagej/imagej-ops/blob/master/src/test/java/net/imagej/ops/AbstractOpTest.java#L137-L151) does not put random values as long as `fill` is false. It is of course ok to use create, but I think if we are in a unit test and there is a helper method to achieve the same purpose, would it be better to reuse the code?

Thanks for mentioning the expensive test! I will try that.

bnorthan:
@LeonYang5114 

> I think generateFloatArrayTestImg does not put random values as long as fill is false. It is of course ok to use create, but I think if we are in a unit test and there is a helper method to achieve the same purpose, would it be better to reuse the code?

I forgot that there was a flag to skip the random values.  In that case it is fine to use the helper method.  I made the tests sizes even smaller to speed it up a little more.  Is there is anything else to do on this pull request??   I'd like to merge it so I can rebase off of it in the fft_refactor branch.

LeonYang5114:
@bnorthan I think everything looks good. I am not sure if I should merge the branch, as I created it, so please merge it if you want. Thanks!

PR #294, opened 2015-12-11
Optimize geom test

LeonYang5114:
This branch is created with the intention to make the geom test run faster. At first I make only the GeomTest support expensive test, but I recognize that AbstractFeatureTest actually has the two helper methods to create image labeling, so I let the GeomTest support expensive test.

Right now, only GeomTest uses the createLabelRegion2D/3D. But in the future, if any test class extends AbstractFeatureTest, it should have separate expected values for the expensive and inexpensive images.

The last commit is kind of trival. I did that because I have no idea what cause the performance problem at first, but I think that commit improve the code a little bit. If you don't like it, I can remove that commit or put it in another branch.

After these modifications, the running time of the GeomTest drops from about 5 sec to less than 2 sec.  I think we might not want to shrink the image size further, which could compromise the test.


ctrueden:
Nice work.

PR #298, opened 2015-12-16
Fft refactor3 WIP - for review

bnorthan:
@dietzc 

Hi Christian - After reading the paper you sent I made more changes to the deconvolution design to make it easier to add multi-view deconvolution.  So this branch is the one to review. 

I split Richardson Lucy into [RichardsonLucyCorrection](https://github.com/imagej/imagej-ops/blob/fft_refactor3/src/main/java/net/imagej/ops/deconvolve/RichardsonLucyCorrection.java) and [RichardsonLucyUpdate](https://github.com/imagej/imagej-ops/blob/fft_refactor3/src/main/java/net/imagej/ops/deconvolve/RichardsonLucyUpdate.java).  

To implement multi-view we can call RichardsonLucyCorrection for each view, then RichardsonLucyUpdate.  Also we can override RichardsonLucyUpdate, (see [RichardsonLucyUpdateTV](https://github.com/imagej/imagej-ops/blob/fft_refactor3/src/main/java/net/imagej/ops/deconvolve/RichardsonLucyTVUpdate.java)).

A good starting point for reviewing is [RichardsonLucyFunction](https://github.com/imagej/imagej-ops/blob/fft_refactor3/src/main/java/net/imagej/ops/deconvolve/RichardsonLucyFunction.java) or [RichardsonLucyTVFunction](https://github.com/imagej/imagej-ops/blob/fft_refactor3/src/main/java/net/imagej/ops/deconvolve/RichardsonLucyTVFunction.java)

Send me questions if you find anything confusing.  

Some commits are WIP.  And I need to fix the deconvolve namespace.  I want to get some feedback on the design before I do that though.


bnorthan:
@dietzc here is an example of how to run it... this will change a bit when we fix up the namespace but it should work for now. 

```java
// create the ImageJ application context with all available services
		final ImageJ ij = net.imagej.Main.launch(args);

		String imName = "/home/bnorthan//Dropbox//Deconvolution_Test_Set//Phantoms//Convolved3D_edge.tif";
		String psfName = "/home/bnorthan//Dropbox//Deconvolution_Test_Set//Phantoms//AberratedPSF3D_N.tif";

		// load the dataset
		final Dataset image = ij.dataset().open(new File(imName).getAbsolutePath());
		final Dataset psf = ij.dataset().open(new File(psfName).getAbsolutePath());

		// display the dataset
		ij.ui().show(image);
		ij.ui().show(psf);

		// richardson lucy 10 iterations, acceleration=true
		Img<?> deconvolvedrl = (Img<?>) ij.op().deconvolve().richardsonLucy(image, psf, new long[] { 20, 20, 20 }, null, null,
				null, null, null, null, 10, false, true);
		ij.ui().show(deconvolvedrl);

		// total variation richardson lucy, 10 iterations, regularization factor=0.002 
		Img<?> deconvolvedrltv = (Img<?>) ij.op().deconvolve().richardsonLucyTV(image, psf, new long[] { 20, 20, 20 }, null,
				null, null, null, null, null, 10, 0.002, false, true);
		ij.ui().show(deconvolvedrltv);
	
		// non-circulant richardson lucy 10 iterations, acceleration=true  
		Img<?> deconvolvedrlnc = (Img<?>) ij.op().deconvolve().richardsonLucy(image, psf, new long[] { 20, 20, 20 }, null,
				null, null, null, null, null, 10, true, true);
		ij.ui().show(deconvolvedrlnc);
		
		// non-circulant richardson lucy 10 iterations, accelration=true, regularization factor=0.002
		Img<?> deconvolvedrltvnc = (Img<?>) ij.op().deconvolve().richardsonLucyTV(image, psf, new long[] { 20, 20, 20 }, null,
				null, null, null, null, null, 10, 0.002, true, true);

		ij.ui().show(deconvolvedrltvnc);
```

ctrueden:
Wow, awesome!! :trophy:

dietzc:
I will review and implement the Multiview Deconvolution of @StephanPreibisch over Chrismas.

dietzc:
and: I'm really! looking forward doing this. Thank you very much @bnorthan.

ctrueden:
@dietzc You are a rock star. :guitar:

dietzc:
@bnorthan as this is a huge refactoring, I suggest an iterative review process. Here are some general comments:

* The `initialize() ` of an `Op` shouldn't do any computation (e.g. initialization of images or so). Rather this method should be used to initialize all other `Op`s, e.g. using `Computers`, `Functions`, `RAI`s, `Iterables` or `ops().op(...)`, such that in the `run` or `compute` method no call to the `OpEnvironment` has to be performed (because it's expensive).

* Don't use `Abstract*` types to declare variables. Rather use the `Interface`s whenever possible.

* `DivideHandleZero` should be a `BinaryHybridOp` which is then used in combination with a `Map`, as soon as we have an implementation for the `MapBinaryXY`.

* Try to put `@Override` annotations which makes the code easier to understand.

* Declare fields as `private` whenever possible. 

In a next step I will start reviewing the hierarchy of FFT. I hope this helps!

dietzc:
Concerning architecture etc: I will try to implement some stuff against the framework and potentially while doing so, potentially also suggest some changes. Would it be OK if I work on your branch or should I create another branch and file PR's against your branch?

bnorthan:
Either way is fine.  Your comments are helpful, thanks.  

Most of your suggestions are straight forward.  

> Don't use Abstract* types to declare variables. Rather use the Interfaces whenever possible.

The main reason I declared variables in an abstract class is for 'temporary' images needed for calculations.  It may make sense to just make these optional parameters.  The caller might have memory already allocated and want to reuse it.

dietzc:
Hi @bnorthan,

I used the free afternoon to go on with my review. Here are some of my suggestions/comments:

* Introduce a `FFTNamespace` which offers all kind of methods (for example `fft()`, `ifft()` etc).
* Move Padding operations into `image()` namespace, as this is a very general concept and maybe remove `PadInput` and rather have a bit of redundant code in the individual pad mechanisms.
* Simplify abstraction hierarchy of `FFT` methods (including `DefaultConvolveFFT`): It seems that some classes are doing more or less the same (e.g. `CreateFilterMemory` and `ConvolveFFTRAI` are doing redundant stuff). In my optimal world, `DefaultConvolveFFT` would simply call the `fft()` and `ifft()` methods and let them do the job (allocate memory etc). Also, often you can avoid `I extends RandomAccessibleInterval<T>`. Instead you could simply write `RandomAccessibleInterval<T>`.

* Remove `FFTSize` and put it into `DefaultFFTOp` as this should never be overwritten anyway.

Unfortunately, I don't have more time today. In general, we really should try to avoid too complicated abstraction hierarchies in favour of having `Op`s calling other `Op's if required and encapsulating (and hiding) functionality inside `Op`s.

bnorthan:
@dietzc thanks for looking at this.  I haven't had a chance to look at your changes yet, so I'll comment more when I look at them.  Some quick comments 

> Remove FFTSize and put it into DefaultFFTOp as this should never be overwritten anyway.

```FFTSize``` may have to be overridden sometimes.  For example different FFT libraries have different fast sizes.  Some GPU libraries are only efficient at power of 2 sizes.  Mines and JTransform have different 'fast' sizes.   FFT size is tightly coupled to the FFT implementation... so perhaps it does not have to be an op.  On the other hand there are scenarios (especially for iterative algorithms) where the caller may want to set up FFT memory once and keep re-using it.  Thus the caller may want to access ```FFTSize``` before calling multiple ffts. 

> Simplify abstraction hierarchy of FFT methods (including DefaultConvolveFFT): It seems that some classes are doing more or less the same (e.g. CreateFilterMemory and ConvolveFFTRAI are doing redundant stuff).

It can definitely be simplified somewhat.  However creating the FFTs for a filter is tricky because you need to create memory for the kernel and the image.  Sometimes you want the border size to be computed from the Image Size and Kernel size as to avoid wrap around.  Alternatively you may just specify a border size.  Also the kernel needs to be handled a little differently as the center needs to be at the origin (see ```PaddingIntervalOrigin```) ```CreateFilterMemory``` handles these dleetails.

I'll look at what you've done, and look for opportunities to simplify the abstraction.

bnorthan:
Hi Christian

I rolled back your commit but took some of the ideas from your commit and implemented them.  

First see https://github.com/imagej/imagej-ops/blob/fft_refactor3/src/main/java/net/imagej/ops/filter/fft/DefaultFFTOp.java

This is very generic.  It does not use RandomAccessibleInterval, just Interval.  It takes a 'pad'. 'create' and 'fft' op as input, and then creates memory, pads and performs the fft. 

I then implemented a 'FFTMethods' version... https://github.com/imagej/imagej-ops/blob/fft_refactor3/src/main/java/net/imagej/ops/filter/fft/FFTMethodsOp.java

I wanted to make sure all FFTMethods code is separate, to make it easier to implement other versions in the future.

bnorthan:
More updates

This class convolves RAIs by passing a multiply op to FFTMethodsLinearFFTFilter:

https://github.com/imagej/imagej-ops/blob/fft_refactor3/src/main/java/net/imagej/ops/filter/convolve/ConvolveFFTRAI.java

FFTMethodsLinearFFTFilter implements a linear FFT filter using FFTMethods fft ops:

https://github.com/imagej/imagej-ops/blob/fft_refactor3/src/main/java/net/imagej/ops/filter/FFTMethodsLinearFFTFilter.java

DefaultLinearFFTFilter is the generic FFT Filter, the ffts could be overridden, possibly by GPU ops...

https://github.com/imagej/imagej-ops/blob/fft_refactor3/src/main/java/net/imagej/ops/filter/DefaultLinearFFTFilter.java

ctrueden:
This is getting really long... the main thing from a history standpoint is to get rid of the WIP commits. (It's ideal if every commit compiles with passing tests (for git bisect) but in the interest of time, we can probably compromise on that this time.)

It might actually be easiest to just squash the history and update it to latest API on master—which I hope is now more stable than ever with all the binary special op stuff.

@bnorthan Can you take a crack at it? Just squash everything and update and force push? Just let us know if you need guidance.

bnorthan:
@ctrueden I agree this is way to long, I'll  squash and update to the latest master.  Is it safe to do that now??  Or are there any disruptive changes pending in the next couple of days?

ctrueden:
I assume this can be closed in favor of #336?

bnorthan:
Yes, this can be closed but not deleted (yet).  I want to move some of @dietzc comments to an issue.  Some of his suggestions (for example "Introduce a FFTNamespace") have not been done yet.

Issue #297, opened 2015-12-16
CopyIterableInterval always creates a copy of type DoubleType. 


tibuch:
Here is a JUnit Test:
https://github.com/imagej/imagej-ops/blob/copyIterableInterval/src/test/java/net/imagej/ops/copy/CopyIterableIntervalTest.java#L94-L107

Issue #296, opened 2015-12-14
Remove unsafe built-in op methods


ctrueden:
Right now, each op has a general-purpose method in its namespace; e.g.:
```java
	/** Executes the "ascii" operation on the given arguments. */
	@OpMethod(op = Ops.Image.ASCII.class)
	public Object ascii(final Object... args) {
		return ops().run(Ops.Image.ASCII.NAME, args);
	}
```

Unfortunately, these signatures mask many problems with calling code, since any time mismatches will be hidden when the method call resorts to the `Object...` signature instead of the intended one.

Better to be explicit and require calls to `run` (`ops.run(MyOp.class, arg1, arg2, ... argN)`).

I started work on the [remove-varargs branch](https://github.com/imagej/imagej-ops/compare/remove-varargs) but it revealed many issues with several op packages—all of which should ideally be addressed as part of this work.

So I'm punting this effort till the next hackathon...

PR #299, opened 2015-12-16
Fix CopyIterableInterval

dietzc:
Fixes https://github.com/imagej/imagej-ops/issues/297.


tibuch:
This looks good to me.

ctrueden:
Thanks @dietzc, certainly this is a step forward. So LTGM for me. But I just want to mention that I think the `create` and `copy` ops need some redesign in general. I bumped up against it again when working on #296, which I now have to put on hold till the hackathon.

PR #300, opened 2015-12-16
DefaultCreateImg: Auto-guess type if possible

dietzc:
Inspired by https://github.com/imagej/imagej-ops/issues/297, I enhance the `DefaultCreateImg` such that it can guess the output type from the `dims` object, if `dims` is an `IterableInterval` or `RAI`.

A more general approach might be to add an "typed" interface in imglib2.


ctrueden:
Nice!

PR #301, opened 2015-12-16
Image moment calculation edits

djniles:
Found some discrepancies in image moment calcs HuMoment2 (sign change in equation) and the normalized central moments (integer division of exponent). Remaining discrepancies were due to Matlab numerical error.

### DIFF
--- a/src/main/java/net/imagej/ops/imagemoments/hu/DefaultHuMoment2.java
+++ b/src/main/java/net/imagej/ops/imagemoments/hu/DefaultHuMoment2.java
@@ -76,6 +76,6 @@ public class DefaultHuMoment2<I extends RealType<I>, O extends RealType<O>>
 		double n20 = normalizedCentralMoment20Func.compute1(input).getRealDouble();
 		double n02 = normalizedCentralMoment02Func.compute1(input).getRealDouble();
 
-		output.setReal(Math.pow(n20 - n02, 2) - 4 * (Math.pow(n11, 2)));
+		output.setReal(Math.pow(n20 - n02, 2) + 4 * (Math.pow(n11, 2)));
 	}
 }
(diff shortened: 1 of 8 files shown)


ctrueden:
@djniles Awesome, thank you very much for following up with this!

@seebacherd Here's the PR [mentioned on the forum](http://forum.imagej.net/t/imagemoments-in-imagej-ops-potential-miscalculations/521).

dietzc:
are the unit tests still passing?

djniles:
@dietzc It looks like previously the unit tests had been passing because an absolute tolerance of 1e-3 was used for expected numbers 1e-11 or smaller.  If I changed the tolerance to be relative instead of absolute (i.e., 1e-14 for an expected value of ~1e-11) Hu moments 2-7 all failed the unit tests prior to the corrections.  After making the corrections, all of the moments passed the unit tests with 1e-3 relative tolerance.

dietzc:
@seebacherd can you review this PR and find a better test image?

seebacherd:
Unfortunately I can neither commit to this PR here or directly in djniles project. I've calculated the Image Moment values with OpenCV for the following image:

lenna

And also updated the tests to use the new values:

[ImageMomentsTest.txt](https://github.com/imagej/imagej-ops/files/65582/ImageMomentsTest.txt)


But there still seem to be some differences between the OpenCV values and our implementation.

ctrueden:
> Unfortunately I can neither commit to this PR here or directly in djniles project.

@seebacherd What you can do is branch from the PR, do your extra commits (and/or rewrite the existing commits), and then push that new branch.

dietzc:
So, how to proceed?

ctrueden:
Did @seebacherd create his own branch with updates? If so, he can file a new PR that supercedes this one.

djniles:
For what it's worth, here are my numbers for Lena using Matlab.  I also include my code for calculating the moments.  It looks like a couple of the moments (5 and 7) disagree by quite a bit, but I'm not sure why.

https://gist.github.com/djniles/ee00f9dc54fc32e906a1

ctrueden:
@tibuch What is the status of this branch? Does the work continue elsewhere? I am confused...

tibuch:
Thanks @djniles for the bugfix! I just updated the test-values and set the test-epsilon to 10e-8. 

I opened another PR with my additional commit: https://github.com/imagej/imagej-ops/pull/491

PR #304, opened 2015-12-21
Refactor AbstractCenterAwareNeighborhoodBasedFilter

dietzc:
since now for each and every pixel there have been several calls to the op-service. should be way faster now!!


dietzc:
I addressed the above mentioned issues. better?

dietzc:
I also generalized the type of the output to `IterableInterval`.

stelfrich:
> I addressed the above mentioned issues. better?

Since monologues are really boring: Definitely better, there is just one call to `Computers.unary()` left per input `RAI`.

Seems good to merge to me, @dietzc.

PR #306, opened 2015-12-30
Logic and Math ops pull request

jaywarrick:
I assume you'll have suggestions before potentially merging, but would like to see similar capabilities added to imagej-ops eventually. Two primary commits associated with this pull request. It builds with maven on my computer passing all tests.

The first provides logical UnaryComputerOps for RealType inputs. These can be used with existing Map functionality to perform "inplace" logical "Blitter" operations on RealType images.

The second provides basic math BinaryComputerOps for RealType inputs. To use these with images, I copied and edited existing Map classes to extend to BinaryComputerOps. It can be expanded more to encompass a wider range of inputs eventually. For now though, the Map infrastructure is used to create a BinaryMapIIAndRAIToRAI.

Apologies ahead of time... Probably inappropriately, an additional MapIterableIntervalToSamplingRAI class was included in the second commit. This was needed to due to the need to override the conforms() method of the MapIterableIntervalToRAI class which requires inputs to be the same dimension AND size. When using sampling regions (see Regions.sample()), the dimensions are the same but not the size. Maybe there is a better way? I haven't used Views yet. In general though, I wonder if an object could be provided for determining if inputs conform (similar to how a Comparator object can be provided to a TreeMap upon construction to perform sorting of keys). Default would be produce existing behavior or there could be a few default conform testing classes used regularly? Then I wouldn't need this slight variant of the MapIterableIntervalToRAI class to be able to use the RealType logic ops with LabelRegions etc the way I would like to.

Let me know your thoughts.

Thanks!

Jay


jaywarrick:
I see now there there are a few pull requests that relate to binary ops and logic from @hinerm. If changes are needed to keep consistent or there is overlap to remove. Just let me know.

Cheers

dietzc:
hi @jaywarrick 

thanks for your efforts. @LeonYang5114 and @ctrueden worked a lot on these kind of ops over the last couple of weeks. I think they can comment in more detail (status quo etc).

Some general comments from my side:
* I don't understand what "blitter like" means.
* There are lot of commits in your PR like 'Merge remote-tracking branch 'imagej/master'. Please use fetch/rebase.
* There are several commits without an associated GitHub account. 
* In https://github.com/jaywarrick/imagej-ops/commit/5335cc56fac5301189790bf8e1ba36c6fd9acfea for example the `NotEqual`: I'm not sure if the output of a boolean operation should ever be a `RealType`. Rather, we should force it to be a `BooleanType` and have `Converter`s from the corresponding type to `BooleanType`?
* Why aren't these `Op`s `BinaryFunction`s?

Cheers,

Christian

jaywarrick:
Thanks for the comments @dietzc. I figured there would be things like that to address.

* Blitter-like refers to the Imagej1 Blitter interface that was used to define most of the logic and math operations for ImageProcessor (http://rsb.info.nih.gov/ij/developer/api/ij/process/Blitter.html)
* Sorry, I generally use the GitHub desktop app which doesn't do rebase operations, only merge. I'll try to figure out rebasing from the command line, but switching to using command line more generally where it can all sink in might take a while given other aspects of my research I have to attend to.
* Not sure about the commits with no associated GitHub account. Not sure how that can/did happen. Thoughts? Easy to remedy?
* Re: RealType output of boolean operations... I was thinking from the historical image processing perspective here. I should have been thinking more generally as you point out. My thinking was based upon ImageJ1 behavior. When you do a boolean operation between two 8-bit images, the results of the boolean operation are pixel values of either 0 or 255 (i.e., MAX_VALUE). I was trying to extend that mentality to other bit depths. Your idea for conversion makes sense but I think it is actually easier to return a boolean type directly and convert to RealType if desired, right? Maybe that is what you were saying? Right now the op calculates the boolean value internally, interprets the result and returns the appropriate RealType. Does the automatic converter mechanism also work on outputs in addition to op inputs? Maybe the code I submitted is better intended as a set of ops that are slightly more optimized for this particular type of behavior (i.e., return 0 or MAX_VALUE vs return boolean T or F)?
* Re: Why aren't these [RealLogic] Ops BinaryFunctions: I wanted BinaryInplaceOp behavior but it seemed as thought that concept was potentially shelved based upon what I could see occuring in topic branches and pull requests. I think description for the UnaryComputerOp interface just said not to alter the input variable whereas BinaryComputerOp says not to alter either input. Thus, it seemed technically ok to use the UnaryComputerOp as a BinaryInplaceOp. This is easy enough to change if a BinaryInplaceOp interface is planned.

Cheers,

Jay

ctrueden:
I'll do a full review of all open PRs at the hackathon.

ctrueden:
@jaywarrick Could you please have a look at the current state of master—what @LeonYang5114 has done recently regarding binary math ops using our new `BinaryOp`-based interfaces?

Let us know if there is something more you need there. We do still need to binarize the logic namespace, but @LeonYang5114 will do that soon.

Issue #231, opened 2015-09-29
Wrapped FunctionOp


dietzc:
Hi @ctrueden,

there is some problem with `FunctionOp`s which are discovered using the `OpService`. Assume that I try to resolve a `FunctionOp` based on an input of type `A` and some output of type `O`. The `OpService` finds an `FunctionOp` `B to O` as there exists a converter from `A` to `B`. However, the resulting `FunctionOp` will only accept `A`s in the `compute(A in)`. Shouldn't the `OpService` wrap the orginal `FunctionOp`, such that the wrapped `FunctionOp` uses a converter inside the `compute(B...)`.

ctrueden:
Hmm, good idea.

There may be performance issues with the wrapper calling into the `ConvertService` every time `compute` is called, though. So we may want to also cache the `Converter` instance used.

dietzc:
which could be done in the initialize of the wrapped op?

ctrueden:
The wrapper will need to do the caching, probably.

dietzc:
https://github.com/imagej/imagej-ops/commit/1afa9f001548f9fbac087cdc99a23821bffbdd7f Wouldn't this be enough, if the `OpService` instantiates this `Op`?

ctrueden:
Yeah. But I wouldn't label the worker `FunctionOp` or `Converter` with `@Parameter`, since the `ConvertedFunctionOp` is not actually declared as an `Op` plugin.

And the framework still needs to instantiate the `ConvertedFunctionOp`, which requires determining and setting the associated `Converter`. Probably, the [`specialOp` method](https://github.com/imagej/imagej-ops/blob/a83d59247c16637eb4228c6edf29ce2d12fe4f20/src/main/java/net/imagej/ops/AbstractOpEnvironment.java#L742-L750)) should be revised to support this.

dietzc:
agreed. You always would have to create a `ConvertedFunctionOp`or `ConvertedComputerOp` if the special input of the `FunctionOp` or `ComputerOp` is converted. This would make the complete system more consistent.

dietzc:
added test-case demonstrating the problem

https://github.com/imagej/imagej-ops/tree/converter-bug

PR #307, opened 2015-12-31
Featuresets pull request

jaywarrick:
Just to revisit the idea of merging featuresets into master, here is a pull request that we can chat about.

We could try to pull to master or optionally to the existing featuresets branch. I chose to do it this way as the featuresets branch was so out of date, it is easier to see what is changed relative to master (potentially incurring a bit of lost, but confusing, commit history).

Builds and passes existing tests (more needed?).

Let me know your suggestions.

Thanks


dietzc:
Hi @jaywarrick,

Please find some general comments here https://github.com/imagej/imagej-ops/pull/306.

Additonally: I actually closed my former PR with the code you are more or less resubmitting, because I there are some serious limitations with the design of `FeatureSet`s which have to be resolved first. For example the usage of `attr`s is really tailor made such that the `FeatureSet`s can be used in KNIME. I will discuss a better design with @ctrueden on the Konstanz hackathon next week and follow up in this thread. In general I would love to see `FeatureSet`s in ops, but we have to be careful with the design such that everybody can easily use and adapt it,

Best,

Christian

PS: This PR also mixes some topics (e.g. fixing Hu moments and introducing `FeatureSet`s).

jaywarrick:
Thanks again for the comments. I completely understand. If you feel that a better implementation is pending soon, I'm happy to wait. I'll just put the feature sets into my project code for now so I can depend upon a release version of imagej-ops.

As far as potential things to think about during the hackathon, here are some of my thoughts that very likely overlap with some of the things you probably have in mind already, but just in case...

* It seems like input conversion gets a bit tricky with OpRef-based FeatureSets. For example, creating a Geometric2DFeatureSet<LabelRegion<Integer>,DoubleType> feature set, individual feature ops are created that take Polygons for many of the features (e.g., BoundarySize) but some features still use the LabelRegion IterableInterval directly (e.g., Size). Currently, because evalFunction of AbstractOpRefFeatureSet calls compute1 instead of ops().run(Op, args...), inputs aren't converted from LabelRegion to Polygon so you get an error. So the ```evalFunction(final UnaryFunctionOp<Object, ? extends O> func, final I input)``` has to be overridden if autoconversion functionality is desired. I'm not sure how much of a computational cost is incurred by calling ops().run(Op, args...) but maybe this could be considered? Alternatively, it could be made more clear or potentially the following bullet point might be addressed in a way that would take care of this nuance.

* If evalFunction is overridden to use ops().run(Op, args...) in my example, the AbstractOpRefFeatureSet will call each feature op, many of which require a Polygon input. Thus, each of these features will separately convert the LabelRegion to a Polygon, incurring extraneous computation compared to a strategy that could keep track of all the desired input types in the featureset, convert once, and supply everybody their desired input type. Might this approach be considered?

* I think there should be an AbstractFeatureSet that the existing AbstractCachedFeatureSet extends (i.e., AbstractCachedFeatureSet is simply an AbstractFeatureSet that is initialized with a CachedOpEnvrionment). This way it will be easier to build feature sets that either utilize or avoid caching depending upon the application. It would then also seem good to also create two versions of AbstractOpRefFeatureSet, one with caching and one without. For my particular use case, it seems that caching balloons memory, slows computation (by constantly searching the cache), and introduces erroneous results for large caches. Is there a potential alternative strategy that would eliminate creation and maintenance of two largely identical libraries of featuresets that only differ with respect to caching?

I have no idea what you all feel as reasonable to potentially address in the hackathon but those would be some things that I would bring up if I were there. Good luck at the hackathon and thanks for all your work that you put into this library.

Best,

Jay

jaywarrick:
It appears that the first bullet point in the previous comment is being addressed in this issue https://github.com/imagej/imagej-ops/issues/231

Cheers,

Jay

dietzc:
@jaywarrick I started a new attempt on `FeatureSet`s https://github.com/imagej/imagej-ops/tree/feature-sets-table. Can't promose anything and it's far from being complete. I hope I will find some time next week to work on this. I will open a PR early for discussion with you and others,... sounds good?

PR #308, opened 2016-01-05
Refactor math

LeonYang5114:
This branch refactor the current math package quite a lot. To sum up:
1. create RealBinaryMath
2. make the RealMathTest complete
3. integrate BinaryOp and InplaceOp into NumericTypeBinaryMath
4. refactor the image-related math operations (they are auto-generated)
    - add more types to the ConstantToXXXImage
    - rename ops with improper names (e.g. named on Img)
    - extends the functionality of some ops (make them BinaryOp and/or InplaceOp as proper)
5. cleanup, auto-format, small bug, etc.

This branch is not ready to merge since the II/RAI/Img related ops do not have unit tests yet. Since I have made quite a few commits, it would be great if someone can review before I move on.

### DIFF
--- a/src/main/java/net/imagej/ops/math/MathNamespace.java
+++ b/src/main/java/net/imagej/ops/math/MathNamespace.java
@@ -211,6 +211,15 @@ public class MathNamespace extends AbstractNamespace {
 		I1 extends RealType<I1>, I2 extends RealType<I2>, O extends RealType<O>> O
 		add(final O out, final I1 in1, final I2 in2)
 	{
+		final Class<?> outClass = out.getClass();
+		final Class<?> in1Class = in1.getClass();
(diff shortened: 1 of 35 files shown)


ctrueden:
Awesome work. Nice clean history.

A couple of quick comments:
* We may want to rename the `FooToBarOutputWhiz` classes to be a little more succinct. Maybe `As` instead of `Output`? Like `IIAndRAIAsRAI` or something...
* We had agreed not to use `II` for `IterableInterval` though at this point I can't remember why not. It seems inconsistent with our decision to allow `RAI` for `RandomAccessibleInterval`. I think I'd rather use `II` also, personally, for succinctness and consistency. (There are currently 17 classes on master that spell out `IterableInterval` instead of abbreviating `II` which we should then rename.)

PR #311, opened 2016-01-07
Refactor map (for PR)

LeonYang5114:
Create the MapBinaryComputer. This branch is based on the refactor-math branch. See #308 

This pull request is opened for peer review purpose.

### DIFF
--- a/src/main/java/net/imagej/ops/OpEnvironment.java
+++ b/src/main/java/net/imagej/ops/OpEnvironment.java
@@ -143,7 +143,8 @@ public interface OpEnvironment extends Contextual {
 	 *         outputs will be given.
 	 */
 	@OpMethod(op = net.imagej.ops.run.RunByType.class)
-	default <OP extends Op> Object run(final Class<OP> type, final Object... args)
+	default <OP extends Op> Object run(final Class<OP> type,
+		final Object... args)
 	{
 		return OpUtils.run(module(type, args));
 	}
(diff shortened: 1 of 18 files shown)


LeonYang5114:
The unit tests for the II/RAI/Image related math ops are now added. @ctrueden

dietzc:
It's hard to review the `MapBinary` if there are so many other changes from https://github.com/imagej/imagej-ops/pull/311. Can you split the `MapBinary` work, then I won't review things twice. 

Thanks a lot ;-)

LeonYang5114:
@dietzc Sorry about that. This branch is based on another branch which has a open pull request #308. That branch focuses on the math stuff, and this branch has the dependency on that because the unit tests for the Maps uses some binary math ops. Is there any trick I can use to separate the two branches?

ctrueden:
I merged #311, and am rebasing this branch over master, to avoid the duplicate commits.

ctrueden:
Great. All commits compile with passing tests, and the history is clean. So I am merging.

My only real concern is the binary inplace map stuff. Do we really need that? What is the point of it?

LeonYang5114:
The `MapBinaryInplace` is experimental, and we would probably remove that. I am still playing around with it but push it to the branch accidently (as you can see I did not have test for that).

I create this because user might want to map a binary operation on two images, and write the result to one of the inputs in-place (suppose the user does not want to create a new image).
One big problem is that the cast from BinaryOp to InplaceOp is unsafe (as you mentioned), and we could not check that in the `conforms`, as we don't know which input the user would like to mutate, if we support the mutation of either.
A simple solution would be only supporting the mutation of the first input. And if the user want to mutate the second one, he/she should create a new image instead. Do you think this is better? @ctrueden

PR #312, opened 2016-01-07
Defaultcreateimg binary

bnorthan:
Make DefaultCreateImg binary and add a test for it.

### DIFF
--- a/src/main/java/net/imagej/ops/create/img/DefaultCreateImg.java
+++ b/src/main/java/net/imagej/ops/create/img/DefaultCreateImg.java
@@ -30,46 +30,36 @@
 
 package net.imagej.ops.create.img;
 
-import net.imagej.ops.AbstractOp;
+import org.scijava.plugin.Parameter;
+import org.scijava.plugin.Plugin;
+
 import net.imagej.ops.Ops;
-import net.imagej.ops.special.Output;
+import net.imagej.ops.special.AbstractBinaryFunctionOp;
 import net.imglib2.Dimensions;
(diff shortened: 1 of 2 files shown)


ctrueden:
I'm confused... isn't this what we talked about doing in the Gitter? So we should merge this, right?

ctrueden:
@bnorthan If the reason you closed is to work on a shinier version of the patches somehow, please note that you can force push to the same topic branches to update the PR, instead of closing and opening another.

bnorthan:
I couldn't figure out how to use the same op instance to create images of different types.  So that got me thinking maybe it should unary after all, with type as a required parameter...

```
BinaryFunctionOp<Dimensions, FloatType, Img<FloatType>> op =
			(BinaryFunctionOp) Functions.binary(ops, DefaultCreateImg.class,
				Img.class, Dimensions.class, new FloatType());
// this is fine
Img<FloatType> test1 = op.compute1(ArrayImgs.bytes(new long[] { 10, 10 }));

// but the type parameter in compute2 is redundant??
Img<FloatType> test3 =op.compute2(ArrayImgs.bytes(new long[] { 10, 10 }), new FloatType());

```
I played around with wildcards a bit, to see if I could create images of different types with the same op instance... but couldn't get it to work, so figured I'd wait and see what you guys did.

ctrueden:
I believe the `create.img` design has been made pretty good now with 267e2655191cf10da6589ba9cd3864db664808b1. As always, please speak up if you have any additional suggestions or concerns about it.

Issue #309, opened 2016-01-05
Any usable op should rarely extend AbstractOp directly


LeonYang5114:
Unless for rare reasons, a usable (not abstract) op should not extend AbstractOp directly, since we have created the `NullaryOp` family, and those ops should at least extend e.g. `NullaryFunctionOp` or `NullaryComputerOp` or whatever `SpecialOp` as proper.

For example, the [DefaultCreateImg](https://github.com/imagej/imagej-ops/blob/76e7bba1bd541efd5f47b13f91cf379ffd9b4223/src/main/java/net/imagej/ops/create/img/DefaultCreateImg.java#L57-L58) could extend `AbstractUnaryFunctionOp<Dimension, Img<T>>` instead.

ctrueden:
Addressed by #315.

PR #315, opened 2016-01-11
Make op implementations extend proper abstract SpecialOp

LeonYang5114:
Although many ops are modified, the changes are similar so I do not
split the commit. All of them include changing "extends AbstractOp" to
"extends AbstractNullary(Unary)FunctionOp", renaming variables, and
removing redundant methods/variables/imports. The logic of the classes
are not changed, and the way to retrieve these ops remains the same.
This is achieved by only making the required parameters as inputs.

See #309


ctrueden:
Thanks!

PR #316, opened 2016-01-13
Do op matching only in initialize() method (in most cases)

LeonYang5114:
~~This branch is not ready to be merged, but I would like to know if there's any comment.~~
This branch is ready now.

I search the usage of `ops()` outside `initialize()` by calling
`grep -nR -B 3 --include="*.java" --exclude="*Namespace*" --exclude="*Test*" "ops()" | less`
and highlight the `initialize()`s. It seems that quite a bit of the ops have already done `initialize()`. 

I recognize that there are some rare cases that we might not want to do initialization, or we need to refactor the op a bit. I will share with you guys those cases once I am done with all ops (currently working on package starting by 't').

I also find that I have to do a lot warning suppression in order to initialize, because the `XXXFunctionOp`s returned by the `Functions` are always raw. Not sure if that's good practice.

@ctrueden @dietzc


LeonYang5114:
- Some ops are undergone refactoring (e.g. filter package) or need to be refactor before doing the initialize(), so I did not include them in this branch.
- [NormalizeIterableComputer](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/image/normalize/NormalizeIterableComputer.java#L70-L71) has a matching call that cannot be initialized, since one of the additional parameters is dependent on the input. Probably need to refactor [NormalizeRealTypeComputer](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/image/normalize/NormalizeRealTypeComputer.java). [NormalizeIterableIntervalFunctionOp](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/image/normalize/NormalizeIterableIntervalFunction.java) could be combined with it to make a Hybrid.
- [Hyperslice](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/slicewise/Hyperslice.java#L218-L219) calls a matching in its `get()` method, which is very wasteful. However, it could not be initialized unless the `CropOp`s are changed to binary ops.
- [AbstractApplyThresholdIterable](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/threshold/AbstractApplyThresholdIterable.java#L50) calls a matching in its `compute1()`. Since the additional parameter is dependent on the input, this could not be initialized, unless all the `ApplyThreshold` ops are changed to binary ops.

ctrueden:
Fabulous. I will try to find time to review this in detail tomorrow morning. We should also address your remaining points by filing issues and/or solving them as time allows.

ctrueden:
:sparkling_heart:

PR #317, opened 2016-01-13
Thinning from Knime/Knip (cont'd)

kephale:
@dietzc For your review!


LeeKamentsky:
Interesting, guys. Do you all have a medial axis transform in addition to the thinning approach? We use a hybrid in CellProfiler for skeletonization:

* Calculate the distance to the edge for every point in the foreground (distance-transform)
* Points along the border have the same distance to the edge - you want to visit the ones with the fewest neighbors (0 to 8 in the 3x3 grid) first so you will erode corners before edges and this results in fewer spurs in your skeleton.
* Order the points from lowest distance to greatest, then by the # of neighbors, from lowest to highest and then use a pseudo-random permutation to break ties (this lessens artifacts caused by applying a consistent bias in order to the skeletonization)
* Do the thinning operation in-place, with a cursor that iterates the points according to the order. CellProfiler uses a 512-entry boolean table to represent all possible 3x3 combinations. The logic of the table is:
 * If the center point is false, the entry's value is false
 * If there are 2 or fewer pixels, the entry's value is true (keep)
 * Determine the number of connected components in the 3x3 grid and the number of connected components in the 3x3 grid with the center point being false. If these numbers are different, keep (removing the point changes the connectivity, breaking the skeleton).
 * Otherwise set the center point to false.

IMHO, our skeletons are very aesthetic - they have very few spurs and an appearance that honors the form of the original object. Also, the running time may be less since there's only the single pass when thinning (but I'm not sure how expensive the distance transform is or whether some sort of marching strategy could be used in its place to get the distance ordering more quickly).

I have a feeling that the above doesn't quite fit into the ThinningOp framework and is something different. Nevertheless, the code here looks like it captures the classic algorithms correctly and is worthy even in the presence of our approach.

Our code is here:
https://github.com/CellProfiler/centrosome/blob/master/centrosome/cpmorphology.py#L3676
and here https://github.com/CellProfiler/centrosome/blob/master/centrosome/_cpmorphology2.pyx#L20

kephale:
Oh cool! It does strike me that with fairly minimal effort the thinning methods in this PR could be expressed in the same form as your approach in CellProfiler. Something like, ask a ThinningStrategy for a traversal order and whether it is a multi-pass method. That might facilitate some interesting hybrids of your distance transform ordering and the current ThinningStrategy's mechanisms for deciding to remove pixels. How would you feel about that?

LeeKamentsky:
I think it's doable. I think you need a distinction between a transforming method that uses your buffer-flipping technique and a mutating method which copies the source to the destination initially, then operates only on the destination (my method wouldn't work if it were only to examine the source pixels - it depends on the prior thinning operations). I love the idea of asking the strategy for the cursor, very nice encapsulation of the behavior.

ctrueden:
@kephale I rebased this over the latest master. However, before we merged, there are two remaining issues:

1. Would it make sense to have each thinning algorithm be its own op? So `morphology.thinGuoHall`, `morphology.thinZhangSuen`, etc.? Then we can dispense with the weird intermediate service layer. If people want to add more thinning algorithms, they just make a new `morphology.thinFoo`. Users probably want to control which one gets called always, right? After all, there is no automatic selection of the "best" thinning strategy for any circumstance, is there? Would such a thing even make sense?
2. We need unit tests.

kephale:
@ctrueden should be good to go.
Unit tests have been added
Each thinning op has its own name

ctrueden:
Thanks @kephale!! :beers:

PR #318, opened 2016-01-13
Refactor stats ops

imagejan:
The `stats.median` and `stats.percentile` ops now make use of `stats.quantile` via op chaining.


ctrueden:
Thanks!

PR #321, opened 2016-01-14
RAIToMeshConverter: Fix supports

bnorthan:
Modify RAIToMeshConverter.supports() based on how LabelRegionToPolygonConverter.supports() is implemented.   This is done to prevent errors during op matching.  

In supports():

1. Check that source object exists.  
2. Check that the number of dimensions are 3
3. Check that destClass is a Mesh.


ctrueden:
Thanks!

Actually, @hinerm was bitten by this problem in a different scenario, and fixed a problem with the SJC converter framework in general (https://github.com/scijava/scijava-common/commit/6db9fe07f4432754e4a458c7748eb37b273310b7 has the details). So I guess it's good we didn't merge this earlier, or we wouldn't have that SJC bug-fix.

PR #322, opened 2016-01-14
Slicewise works on RandomAccessibleInterval

imagejan:
Thanks to @ctrueden, fixed input and output types of the two method signatures of the built-in slicewise op.


ctrueden:
:+1:

PR #323, opened 2016-01-15
Refactor normalize

LeonYang5114:
This branch addresses one of the issue mentioned in #316 that the `Ops.Image.Normalize` ops could not implement `initialize` method unless they are refactored.

Now the logic for handling null parameters are moved to the upstream ops, the `NormalizeIterableIntervalComputer/Function`, so that it is feasible to initialize the helper ops. In order to make the NormalizeRealTypeComputer reusable, I also have to make the parameters of it modifiable through a method `setup(...)`. This is not very clean, but as this op is just a helper and not even a `Plugin`, I think adding this additional signature will not be too bad.

However, I am not able to merge the two ops into a `hybrid` in a nice and clean way, since the `function` one has an additional parameter `isLazy`. If the two ops are combined, all the computer method signature in the `ImageNamespace` will have to include an `isLazy` parameter, which does not make sense (`isLazy` is only used in Function compute1(...)).

I also make the Function op a `FunctionViaComputer`, since I think its logic works in that way.

### DIFF
--- a/src/main/java/net/imagej/ops/image/ImageNamespace.java
+++ b/src/main/java/net/imagej/ops/image/ImageNamespace.java
@@ -239,7 +239,7 @@ public class ImageNamespace extends AbstractNamespace {
 		return ops().run(Ops.Image.Normalize.NAME, args);
 	}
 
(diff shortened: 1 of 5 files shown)


ctrueden:
:+1:

PR #326, opened 2016-01-17
Updating threshold ops to work on IterableIntervals instead of Img.

kephale:
Generalizes threshold ops to use IterableIntervals


ctrueden:
Thank you! :v:

PR #327, opened 2016-01-18
Removed unnecessary computations from local thresholding

stelfrich:
Resulting from the restructuring of `Unary/Binary/NullaryOps` the input of local thresholding ops was computed a second time.


ctrueden:
Closing in favor of #328.

By the way, in future, you can use `git push --force` to update the same branch, instead of opening a second PR.

stelfrich:
@ctrueden Sorry, I was a little confused b/c I hadn't submitted the PR on purpose (at least I wasn't aware of it..). Will do that with future PRs..

PR #329, opened 2016-01-19
small create kernel refactoring

EikeHeinz:
I have renamed the abstract classes for the creation of gauss and log kernels, because not all kernels need a sigma to create them.
On top of that i talked with @dietzc about the calibration that is currently implemented in the gauss and log kernels and removed the corresponding parts from the implementation, since this can be replaced by processing the input image beforehand.

### DIFF
--- a/src/main/java/net/imagej/ops/create/AbstractCreateGaussianKernel.java
+++ b/src/main/java/net/imagej/ops/create/AbstractCreateGaussianKernel.java
@@ -1,5 +1,5 @@
 /*
- * #%L
+* #%L
  * ImageJ software for multidimensional image processing and analysis.
  * %%
  * Copyright (C) 2014 - 2016 Board of Regents of the University of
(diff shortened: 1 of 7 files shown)


ctrueden:
Great. Have you tested that the op still behaves as before? E.g., the [Create and Convolve Points](https://github.com/imagej/imagej-scripting/blob/master/src/main/resources/script_templates/Tutorials/Create_and_Convolve_Points.py) example script in ImageJ?

You can test as follows:

* Build a new imagej-ops
* Drop it into an ImageJ installation
* Run the Script Editor
* From the menu, choose Templates &#9654; Tutorials &#9654; Create and Convolve Points

If it still behaves the same as before the change, I'll merge it!

ctrueden:
We tested, and it is working! Thanks!

bnorthan:
f.y.i.  I added ```testCreateAndConvolvePoints``` to the convolve unit test.  It follows the jython script exactly and asserts that the output image has the same sum, min and max values as the output from the jython script. 

https://github.com/imagej/imagej-ops/blob/master/src/test/java/net/imagej/ops/filter/convolve/ConvolveTest.java

ctrueden:
Thanks @bnorthan !

Issue #200, opened 2015-09-14
CustomOpEnvironment: Split into two versions with different behaviour


dietzc:
For the scenario in which you may want to pass additional `CommandInfo`s to a parent `OpEnviroment` one may want to chose between two options:

* Merge the existing `OpIndex` with the added `CommandInfo`s. 
* Try to match `Op`s first given the added `CommandInfo`s and just fall-back to the existing `OpIndex` if we couldn't match an `Op` in the first try.

Naming suggestions: `JointCustomOpEnviroment` and `PrioritizedCustomOpEnvironment`.

ctrueden:
`CustomOpEnvironment` is actually both, because it sets the additional ops to `FIRST_PRIORITY`.

PR #332, opened 2016-01-21
[DO NOT MERGE] Fast op index

ctrueden:
FOR DISCUSSION ONLY (for now)

It's not too much faster. And I don't know whether it is always faster. See commit messages.

For your perusal, @dietzc. Somewhat relevant to #200.


ctrueden:
Closing for now, since it does not seem urgent. We can reopen later if needed.

PR #333, opened 2016-01-22
Add method for extending RAIs keeping the interval

stelfrich:
Extending a `RAI` with an `OutOfBoundsFactory` and setting the interval of the output to the interval of the input is a recurring pattern when working with `Neighborhoods`, e.g. local thresholding or filtering.

@ctrueden There are some style glitches in RAIs.java that could be fixed while at it. Should I do that?


ctrueden:
Thanks @stelfrich !

Probably we should do a full style cleanup at some point. No rush though.

PR #334, opened 2016-01-22
Refactor apply threshold

LeonYang5114:
The sole purpose of this branch is to `initialize` the op matching call in [`AbstractApplyThresholdIterable`](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/threshold/AbstractApplyThresholdIterable.java#L50), which might not be called frequently, but each time, it will trigger a series of op matching calls, since it is a high level op.

This is achieved by making `ApplyThresholdComparable` and `ApplyConstantThreshold` binary ops. [This line of code](https://github.com/imagej/imagej-ops/blob/28ba5ab2fac591af6ac4d1a782b1705a94387943/src/main/java/net/imagej/ops/threshold/apply/ApplyConstantThreshold.java#L71) does not seem to be elegant to me, but I cannot come up with other ideas right now.

### DIFF
--- a/src/main/java/net/imagej/ops/threshold/AbstractApplyThresholdImg.java
+++ b/src/main/java/net/imagej/ops/threshold/AbstractApplyThresholdImg.java
@@ -31,6 +31,7 @@
 package net.imagej.ops.threshold;
 
 import net.imagej.ops.Ops;
+import net.imagej.ops.special.computer.Computers;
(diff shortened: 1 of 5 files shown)


ctrueden:
I don't see an easy way around the `setInput2` call on the chained op. Intuitively it seems fine, so let's not worry about it for the moment.

PR #335, opened 2016-01-26
LocalMean Op that operates on RAIs

stelfrich:
Currently, the local thresholding ops (this PR is about `LocalMean` for now) are implemented to work on `Iterable<T>` (which are the generated neighborhoods of on an input image). This requires them to be called with `ThresholdNamespace.apply(...)` to really work as __local__ thresholding methods. Just by taking a look at the method definitions in the `ThresholdNamespace` it is hardly possible to understand this necessity.

Open questions:
1) Is it desirable to change the `ThresholdNamespace.localMean` (for example) definition to operate on RAIs (as proposed by this PR which breaks the completeness test of the `ThresholdNamespace`)?


dietzc:
local mean ops usually always operate on `RAI`s. the current naming is simply a bit confusing from my point of view. The operations which map from `Iterable<T>` to an `BooleanType` can be accessed via the namespace, but this doesn't make sense. Usually, the term local thresholding is used to apply these map operations on an entire image neighborhood-wise. 

What do you think?

stelfrich:
> What do you think?

Sticking to the example, I would suggest that we replace
```java
public <T extends RealType<T>> BitType localMean(final BitType out, final Pair<T, Iterable<T>> in, final double c)
```

with 

```java
public <T extends RealType<T>> RandomAccessibleInterval<BitType> localMean(final RandomAccessibleInterval<BitType> out, final RandomAccessibleInterval<T> in, final double c)
```

This, however, raises the question of how those method calls are handled internally: A call to `ThresholdNamespace.apply` with the according op on the input image or as suggested by the PR by an additional op (see [LocalMeanRAI.java](https://github.com/stelfrich/imagej-ops/blob/ee267c6bec05b6333ba6287b2d5422da184e62bb/src/main/java/net/imagej/ops/threshold/localMean/LocalMeanRAI.java)).
While the former requires less boilerplate code, the latter option makes the newly created ops available for use with `OpService.run`, right?

dietzc:
I see. So for me a nice solutions means that from a user point of view things are easily accessible and any `Namespace`s contain methods which can be used from scripts etc. So in this case there definitively should be an `LocalThresholdMeanOp` which has a `RAI<T>` as input and a `IterableInterval<BitType>` as output. Same for `LocalThresholdBernsenOp` etc. They may internally base on `MapNeighborhood` (as the basic implementations will work based on `Shape`s) and they will also utilize some kind of `Computer<Iterable<T>,BitType>` which maps from a `Neighborhood` to a `BitType` (i.e. performs the thresholding). However, the `Computer<Iterable<T>, BitType>` does not necessarily be of type `LocalXY`. Potentially the `ComputerOp` is just a `public` inner-class of the corresponding `LocalXY`. Depends whether the `Computer` might be useful in other contexts.

Does this make sense?

stelfrich:
> So in this case there definitively should be an `LocalThresholdMeanOp` which has a `RAI<T>` as input and a `IterableInterval<BitType>` as output. 

:+1:

> they will also utilize some kind of `Computer<Iterable<T>,BitType>` which maps from a `Neighborhood` to a `BitType` (i.e. performs the thresholding)

That will actually not work, because in this case the `Computer` doesn't have the value of the (center) pixel that is thresholded (that was the original motivation behind `CenterAwareComputerOp` as implemented by @Squareys, I guess).

dietzc:
then replace `Computer` with `CenterAwareComputerOp` in my text ;-). Still, these `Op`s are actually no `LocalXY` ops...

stelfrich:
> Still, these Ops are actually no `LocalXY` ops...

True. But they aren't really reusable either due to their "weird" `ValuePair<T, IterableInterval<T>` input. If they were just operating on `IterableInterval` one could apply them to complete images. But even then, it doesn't make sense..

> Potentially the `ComputerOp` is just a `public` inner-class of the corresponding `LocalXY`

Rookie questions: can inner-classes also be discovered with the typical ops matching? Plus, what interface would such an inner-class implement (read: do I always need an interface for a class to be discoverable as op?)

dietzc:
>  If they were just operating on IterableInterval one could apply them to complete images. But even then, it doesn't make sense. 

True, but still I wouldn't expose them in any namespace.

> Rookie questions: can inner-classes also be discovered with the typical ops matching? Plus, what interface would such an inner-class implement (read: do I always need an interface for a class to be discoverable as op?) 

Good question. I think if they are static, yes. If not, we can still put them in the same package but don't interface them as `LocalXY`.

ctrueden:
> can inner-classes also be discovered with the typical ops matching?

Yes, but they must be static.

> what interface would such an inner-class implement (read: do I always need an interface for a class to be discoverable as op?)

It is best practice to always implement `Ops.Foo.Bar`, and write `@Plugin(type = Ops.Foo.Bar.class)`. Do you have some reason not to do this in any case?

dietzc:
these `Op`s shouldn't really be exposed in any namespace. I think thats why he is asking

ctrueden:
Currently, there is no way to flag an op as exempt from the namespace completeness test. But I guess we could add this if others are sure we want that.

dietzc:
@stelfrich do these ops actually need to be discovered? at least all of them? or can they also be simply hidden in the `LocalXY` op?

stelfrich:
> @stelfrich do these ops actually need to be discovered?

I can't come up with a use-case in which I wanted to re-use them. So no, I don't think that they need to be discovered. However, how would I use them from within `LocalXY` if they are not discoverable?

ctrueden:
> how would I use them from within LocalXY if they are not discoverable?

You could always use the good ol' `new` keyword!

stelfrich:
> You could always use the good ol' `new` keyword!

Sure. Just thought there was a more "elegant" way..

dietzc:
`new`. hip again :+1:

ctrueden:
Yeah, you basically have to ask yourself: do you want the extensibility provided by ops? If so, your internal ops need to be plugins, which get discovered, and can hence be overridden. Then you can use the op matcher, etc.

Or: do you just want to do stuff internally, with no meddling from the outside? In that case, just write "plain old Java code" however you would normally.

stelfrich:
@dietzc I have applied your suggested naming scheme (also applied to the `ThresholdNamespace`). Along the way, I have changed the output of `LocalThresholdMean` to be `IterableInterval` instead of `RAI` which also propagated to the `OpEnvironment`. What's your opinion on the general refactoring? Comments on the implementation are also more than welcome.

stelfrich:
I have thought about this a bit more and came up with some refactoring. But still, I don't really like how [LocalThresholdMean](https://github.com/stelfrich/imagej-ops/blob/248c601578090b7aed7d8047a2686df18cede765/src/main/java/net/imagej/ops/threshold/localMean/LocalThresholdMean.java) is implemented, especially the fields of the inner-class.

dietzc:
why don't you like it? it think it's not too bad?

stelfrich:
@dietzc It feels odd to duplicate the parameters to the inner-class. But in the end, it will get the work done.

What about renaming of the ops, i.e. `LocalX` to `LocalThresholdX` with inner-classes `LocalXComputer`? Or should we just keep the names of `LocalX` with my proposed implementation. This way a `LocalX` op really will operate on local neighborhoods and the computation will be done in `LocalXComputer`.

dietzc:
I'd make the inner-class private and non-static.

about naming: you mean `LocalMeanThreshold` and `LocalMeanThresholdComputer`?

stelfrich:
Included in #345.

Issue #313, opened 2016-01-08
Tweak Map


LeonYang5114:
The newly merged branch #311 has some issues as indicated by @dietzc and @ctrueden, and I summarize them here so that we could work on fixing it in the hackathon or later.

- [x] Naming. We agreed on calling `RAI` for `RandomAccessibleInterval` but not `II` for `IterableInterval`, in naming ops. However, using `IterableInterval` when naming `Map` ops will generate names such as `MapIterableIntervalAndIterableIntervalOutputIterableInterval`. (This issue also exists in the math template ops [IIToIIOutputII](https://github.com/imagej/imagej-ops/blob/master/src/main/templates/net/imagej/ops/math/IIToIIOutputII.vm). Names such as `MapTwoIterableIntervalToIterableInterval` had been suggested.

- [ ] Avoid method calls within loops. For example [here](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/map/MapIIAndIIToRAI.java#L81). The `op` can be retrieved outside the loop.

- [x] MapBinaryInplace. Do we actually need this interface and its implementations? `Map`s extending this interface are supposed to support mutation of either `in1()` or `in2()`, but this introduce some type unsafety, if the current implementation is not changed. We could probably remove this interface, and any binary `Map` extending `MapInplace` mutates the first input by default.

- [x] MapBinaryComputer implementations not complete. Right now we have 3 binary `Map` ops:
II+II=>II, II+II=>RAI, II+RAI=>RAI. To make the collection complete, we still need
II+RAI=>II, RAI+II=>II, 
RAI+II=>RAI, RAI+RAI=>II
(RAI+RAI=>RAI is not necessary since one of the RAIs could be converted into II)
The problem is, this proliferates the namespace, and introduces ambiguity. Do we want more of these binary `Map` ops? or do we want to remove some we currently have? or right now is fine?

More issue will be added later.

ctrueden:
> MapBinaryComputer implementations not complete.

Unfortunately, it seems likely we will need all the binary map combinations, due to limitations in how converters work (they operate independently on each parameter, without knowledge of the other parameters).

> Avoid method calls within loops.

In the example, typically there is no difference on modern JVMs, but it should be tested. If it's easy to factor out, there is no harm, though.

> not `II` for `IterableInterval`

[...]

PR #337, opened 2016-01-27
Tweak map

LeonYang5114:
This PR is addressing #313 ~~but has not finished.~~

The problem I have right now is, should I auto-generate all the inplace maps also? There are only 8 inplace maps, not too many; but auto-generate them seems to be consistent with generating the other maps.

As mentioned in the issue, I also recognize that there is some possibility to use lambda expression to make the structure better and the code more readable. I think I will work on that later.


LeonYang5114:
The map package is now done refactoring: all versions of unary/binary/inplace II/RAI are complete; many are auto generated; many tests are also auto generated; tests are passed.

Only thing remaining is that we should use lambda expression to make MapUtils less wordy, or create Cursor similar to TwinCursor. Anyway, they should not be done in this branch. @ctrueden

ctrueden:
Looks awesome. The only thing I would change is rename `MapUtils` to `Maps` since that is our convention for utility classes. I will merge and then do it.

PR #338, opened 2016-01-28
Add workaround for a bug in maven-compiler-plugin

LeonYang5114:
`useIncrementalCompilation` feature is supposed to be turned on by default, but it is actually not working. However, this bug could not be easily recognized, since we all use the incremental build provided by Eclipse, and every call to `mvn` tells us the module is up to date.

However, when you modify a template (either, .vm or .list, which I did frequently recently), the auto-generated source files could not be incremental built by Eclipse, and therefore it triggers the complete recompilation of the project, which takes me about 40 extra seconds every time :-(

To generate this bug, just turn off Project -> Build Automatically, and modify 1 file and then call `mvn`. The whole project will be recompiled.

See https://issues.apache.org/jira/browse/MCOMPILER-209 for information about the bug.

It has been tested that the version 3.5 still has this bug, and that by doing this workaround, the useIncrementalCompilation feature is turned on. To verify, repeat the instruction above to generate the bug, and now it should show that "Compiling 1 source file to ..." and takes significantly less time to finish.

Not sure if there are better solutions, and whether projects need this workaround too. @ctrueden

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -323,6 +323,14 @@
 			</plugins>
 		</pluginManagement>
 		<plugins>
+			<!--  Workaround for a bug in maven-compiler-plugin.
+			See https://issues.apache.org/jira/browse/MCOMPILER-209 -->
+			<plugin>
+				<artifactId>maven-compiler-plugin</artifactId>
+				<configuration>
+					<useIncrementalCompilation>false</useIncrementalCompilation>
+				</configuration>
+			</plugin>
 			<plugin>
 				<groupId>org.codehaus.mojo</groupId>
 				<artifactId>groovy-maven-plugin</artifactId>


ctrueden:
Thank you for researching this, @LeonYang5114. Fascinating that the flag (naively) seems to do the exact opposite of what's expected.

Nearly every component in the SciJava software stack compiles in 3-5 seconds or less... except Ops. I believe that the Java-8-isms are making the compiler think much harder about what to do, since Ops used to compile in a few seconds, but now takes an order of magnitude longer.

So for now, let's just merge this PR as is, setting the incremental flag to false for Ops only. And in the future, we could move it upstream into `pom-scijava`, if we deem it useful for other components.

hinerm:
@ctrueden @LeonYang5114 - looks like the compilation speed issue [may be fixed](https://bugs.openjdk.java.net/browse/JDK-8055984) (or at least improved) with `jdk9b84`.. I haven't been able to confirm locally that this actually fixes Ops.

hinerm:
see also [SO](http://stackoverflow.com/questions/30707387/troubleshoot-slow-compilation)

Issue #314, opened 2016-01-11
ScaleImg Bug - out is always the same image type of in


LeonYang5114:
[ScaleImg](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/image/scale/ScaleImg.java#L92) always generate the `out`  of the same image type of the `in`, while `out` could be of much larger size. For example:
```
// generate an (Integer.MAX_VALUE x 1) ArrayImg
final Img<ByteType> in = ArrayImgs.bytes(new byte[Integer.MAX_VALUE - 2], Integer.MAX_VALUE - 2, 1);
final double[] scaleFactors = new double[] { 1.0, 2.0 };
Img<ByteType> out = ops.image().scale(in, scaleFactors, new NLinearInterpolatorFactory<ByteType>());
// stack trace:
// java.lang.RuntimeException: Number of elements in Container too big, use for example
// CellContainer instead: 4294967290 > 2147483647
// ...
```
It should use a more sophisticated way to create the out.

In addition, [this line](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/image/scale/ScaleImg.java#L94) is not working as expected. In the current implementation, `in` and `out` should always have the same `IterationOrder`, since they have the same `ImageFactory`.

I would fix this bug later.

ctrueden:
@dietzc and I agree that this is a bug.

The problem is that `ScaleImg` does:
```java
		final Img<T> out = input.factory().create(newDims, input.firstElement()
			.createVariable());
```
When actually it would be better to use a `create.img` op instead, which decides how to create the best `Img`.

I will try to address this at the same time I clean up the `create` namespace in general.

PR #339, opened 2016-01-30
Fix ScaleImg bug of assumption about IterationOrder

lnyng:
See #314 

The previous code generate the output image using the image factory of the input image. Since the output image has different dimension of the input image, the result could cause errors. For example, scaling up an large size image, so that the result image has more number of pixels than Integer.MAX_VALUE.

However, this case rarely happens in practice (scaling up an already large image?), and I find that no test on this feature could be run in reasonable amount of time. One significant reason is that, creating an `CellImg` of size 50000x50000 takes too much time (I never see it finish. Takes at least a few minutes), while in comparison, creating two `ArrayImg` of size 50000x25000 takes a few seconds. I think this is due to the use of default cell size in `DefaultCreateImgFactory`. I will file another issue and work on that later.

### DIFF
--- a/src/main/java/net/imagej/ops/image/scale/ScaleImg.java
+++ b/src/main/java/net/imagej/ops/image/scale/ScaleImg.java
@@ -30,16 +30,19 @@
 
 package net.imagej.ops.image.scale;
 
+import net.imagej.ops.Contingent;
 import net.imagej.ops.Ops;
+import net.imagej.ops.special.computer.Computers;
+import net.imagej.ops.special.computer.UnaryComputerOp;
 import net.imagej.ops.special.function.AbstractUnaryFunctionOp;
-import net.imglib2.Cursor;
+import net.imglib2.FinalDimensions;
 import net.imglib2.FinalInterval;
-import net.imglib2.FlatIterationOrder;
-import net.imglib2.RandomAccess;
 import net.imglib2.RandomAccessible;
 import net.imglib2.img.Img;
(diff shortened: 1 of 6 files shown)


dietzc:
what about using `Copy` here instead of hard coded loops?

LeonYang5114:
@dietzc The branch is updated, and also include code that supports creating Cellimg. I think the `initialize` method I wrote in the `ScaleImg` op is not very nice, but I could not come up with a way to improve right now. Can you check? Thanks

src/main/java/net/imagej/ops/image/scale/ScaleImg.java
```diff
@@ -52,56 +55,75 @@
...
+			new net.imglib2.realtransform.Scale(scaleFactors))), new FinalInterval(
+				newDims));
+
+		copyImg = (UnaryComputerOp) Computers.unary(ops(),
+			Ops.Copy.IterableInterval.class, Img.class, interval);
+	}
+
+	@Override
+	public Img<T> compute1(Img<T> input) {
+		final long[] newDims = new long[in().numDimensions()];
```
ctrueden: Whoa. The `in()` method should never be called from `compute1`. Why are we not using `input` instead here?

ctrueden:
`ScaleImg` became `DefaultScaleView`, so this bug is no longer relevant.

I almost cherry-picked the improvements to `DefaultCreateImgFactory`, but ultimately decided against it. We can add in the `targetCellSize` again when the need arises.

PR #341, opened 2016-02-02
CenterAwareComputerOp as BinaryComputerOp

stelfrich:
In the original implementation of `CenterAwareComputerOp` et al, those ops were implemented as `UnaryOp`s with a `Pair<T, Iterable<T>>` as sole input. With the advent of `BinaryComputerOp`s and the according maps, this special case (shadowing two inputs using `Pair`) and related implementations are no longer required.


ctrueden:
So beautiful! Thank you!

ctrueden:
FYI, I pushed fixes for my niggling comments above.

stelfrich:
> FYI, I pushed fixes for my niggling comments above.

Thanks, @ctrueden!

PR #345, opened 2016-02-09
Refactor local thresholding ops

stelfrich:
With this PR, local thresholding ops operate on `RAI`s while the local computations (on neighborhoods) are delegated to inner classes (`BinaryOp`s as introduced by #341). Convenience methods in the `ThresholdNamespace`, i.e. `localXY()` are also renamed to `localXYThreshold()` to break API (behavior changes from pseudo-local to local). 

Includes previous work as discussed in #335.

cc: @dietzc


dietzc:
to me this looks great! thanks @stelfrich. Tests passing?

@ctrueden API break OK with you?

stelfrich:
> Tests passing?

Sure do.
> API break OK with you?

With regard to API breaks, the most critical change is that the previous pattern of using `ThresholdNamespace.apply` (see https://github.com/stelfrich/imagej-ops/commit/30b5748bfdbeb268c255fe0a2320a3a3dc31f0ea#diff-9c234bde5f090e438954cb97bbd181a8L132 and https://github.com/stelfrich/imagej-ops/commit/30b5748bfdbeb268c255fe0a2320a3a3dc31f0ea#diff-9c234bde5f090e438954cb97bbd181a8L146) for applying a `LocalThresholdMethod` to a complete image is now replaced by the respective `localXYThreshold` methods in `ThresholdNamespace`.

dietzc:
I think its fine. We can merge

Issue #176, opened 2015-08-03
Clarify OpMethods


Squareys:
Hello everybody!

I'd like to clarify how OpMethods in `Namespace` implementations and `OpService` should look like. Consider a common op hierarchy for a statistical max op:
 - `Op.Stat.Max`
 - `MaxOp`
 - `DefaultMax`

Two variants for its OpMethods are possible.
 - first:
```
@OpMethod(op = net.imagej.ops.stat.max.DefaultMaxFilter.class)
public <T extends RealType<T>> T max(final T out, final Iterable<T> input)
{
	final T result =
		(T) ops().run(net.imagej.ops.stat.max.DefaultMax.class, out, input);
	return result;
}
```
 - second:
```
@OpMethod(op = net.imagej.ops.stat.max.DefaultMaxFilter.class)
public <T extends RealType<T>> T max(final T out, final Iterable<T> input)
{
	final T result = (T) ops().run(Ops.Stat.Max.class, out, input);
	//                           ! ^^^^^^^^^^^^
	return result;
}
```

In my opinion, the second version is the "correct" one. That is because the first one prevents Ops from choosing the best option (which is one of its big features after all). A more optimized version for a compatible set of parameters could be never be chosen.

The second one allows this, though. Both are accepted by the NamespaceTest, but if missing, the test will suggest the first one.

Is my thinking correct, or did I miss something?

Greetings, Squareys.

ctrueden:
Firstly, I assume your reference to a `DefaultMaxFinder` is a typo, and should be `DefaultMax`. Correct?

When there is no clash between multiple op implementations, the current pattern is:
```java
@OpMethod(op = net.imagej.ops.stat.max.DefaultMax.class)
public <T extends RealType<T>> T max(final T out, final Iterable<T> input)
{
    final T result =
        (T) ops().run(net.imagej.ops.stat.max.DefaultMax.class, out, input);
    return result;
}
```
In other words: as specific as possible. This ensures you get exactly the expected op when calling its associated built-in method signature.

If multiple ops clash, the current rule of thumb is to choose the "most specific common type" of all involved ops in the clash. Sometimes this ends up being the `Ops.Foo.Bar` interface, but often times there is something more specific.

The thinking was to "lock down" the built-in op signatures as much as possible, since each one typically corresponds to one concrete implementation.

[...]

PR #348, opened 2016-02-10
Make op method extensible

LeonYang5114:
See #176 for the discussion.
Filter and logic namespaces are not included in this refactoring, since that involves more intricate refactoring. They will be done separately.

Since this refactoring reveals some bugs, bug fixes are included in this branch for convenience.

### DIFF
--- a/src/main/java/net/imagej/ops/OpEnvironment.java
+++ b/src/main/java/net/imagej/ops/OpEnvironment.java
@@ -312,7 +312,7 @@ public interface OpEnvironment extends Contextual {
 	/** Executes the "eval" operation on the given arguments. */
 	@OpMethod(op = net.imagej.ops.eval.DefaultEval.class)
 	default Object eval(final String expression) {
-		final Object result = run(net.imagej.ops.eval.DefaultEval.class,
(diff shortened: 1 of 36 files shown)


dietzc:
shouldn't this work also comprise cleaning up methods like https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/copy/CopyNamespace.java#L89, as this case is covered for example by https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/copy/CopyNamespace.java#L126 (the `OpMatcher` will find the `CopyArrayImg` as the op with the highest priority if the passed `IterableInterval`s are actually `ArrayImg`s.

ctrueden:
@dietzc I agree that that is the next step, yeah. But let's get this merged first, then do another PR for that ASAP.

dietzc:
Ok.makes sense.

dietzc:
looks good to me. @LeonYang5114 tests passing? anything else you want to work on before merging? @ctrueden OK with merging?

ctrueden:
[256px-smirc-thumbsup svg](https://commons.wikimedia.org/wiki/File:SMirC-thumbsup.svg)

PR #349, opened 2016-02-10
Sobel and hessian filter

EikeHeinz:
This adds an implementation of a separated sobel kernel, a sobel filter and a hessian filter. Both filters use a derivative filter that uses the separated sobel kernel. Tests are included.
~~I've tested the sobel filter on several images and compared these to the already filtered versions.
If required I can add some assert statements to the SobelFilterTest case.~~

### DIFF
--- a/src/main/java/net/imagej/ops/create/CreateNamespace.java
+++ b/src/main/java/net/imagej/ops/create/CreateNamespace.java
@@ -730,6 +730,16 @@ public class CreateNamespace extends AbstractNamespace {
 		return result;
 	}
 
+	// -- kernelSobel --
+
+	@OpMethod(op = net.imagej.ops.create.kernelSobel.CreateKernelSobel.class)
+	public <T extends ComplexType<T>> RandomAccessibleInterval<T> kernelSobel(final T outType) {
+		@SuppressWarnings("unchecked")
+		final RandomAccessibleInterval<T> result = (RandomAccessibleInterval<T>) ops().run(
+			net.imagej.ops.create.kernelSobel.CreateKernelSobel.class, outType);
+		return result;
+	}
+
 	// -- labelingMapping --
 
 	@OpMethod(
--- a/src/main/java/net/imagej/ops/filter/FilterNamespace.java
+++ b/src/main/java/net/imagej/ops/filter/FilterNamespace.java
@@ -675,6 +675,19 @@ public class FilterNamespace extends AbstractNamespace {
 		return result;
 	}
 
+	// -- hessian --
+
+	@OpMethod(op = net.imagej.ops.filter.hessian.HessianRAI.class)
+	public <T extends RealType<T>> CompositeIntervalView<T, RealComposite<T>> hessian(
+		final RandomAccessibleInterval<T> in)
+	{
+		@SuppressWarnings("unchecked")
+		final CompositeIntervalView<T, RealComposite<T>> result =
+			(CompositeIntervalView<T, RealComposite<T>>) ops().run(
+				net.imagej.ops.filter.hessian.HessianRAI.class, in);
+		return result;
+	}
+
 	// -- ifft --
 
 	/** Executes the "ifft" operation on the given arguments. */
--- a/src/main/java/net/imagej/ops/filter/FilterNamespace.java
+++ b/src/main/java/net/imagej/ops/filter/FilterNamespace.java
@@ -980,6 +980,31 @@ public class FilterNamespace extends AbstractNamespace {
 		return result;
 	}
 
+	// -- Sobel
+
+    @OpMethod(op = net.imagej.ops.filter.sobel.SobelRAI.class)
+	public <T extends RealType<T>> RandomAccessibleInterval<T> sobel(
+		final RandomAccessibleInterval<T> in)
+	{
+		@SuppressWarnings("unchecked")
+		final RandomAccessibleInterval<T> result =
+			(RandomAccessibleInterval<T>) ops().run(
+				net.imagej.ops.filter.sobel.SobelRAI.class, in);
+		return result;
+	}
+
+	@OpMethod(op = net.imagej.ops.filter.sobel.SobelRAI.class)
+	public <T extends RealType<T>> RandomAccessibleInterval<T> sobel(
+		final RandomAccessibleInterval<T> out,
+		final RandomAccessibleInterval<T> in)
+	{
+		@SuppressWarnings("unchecked")
+		final RandomAccessibleInterval<T> result =
+			(RandomAccessibleInterval<T>) ops().run(
+				net.imagej.ops.filter.sobel.SobelRAI.class, out, in);
+		return result;
+	}
+
 	/** Executes the "variance" filter operation on the given arguments. */
 	@OpMethod(op = net.imagej.ops.filter.variance.DefaultVarianceFilter.class)
 	public <T extends RealType<T>> IterableInterval<T> variance(final IterableInterval<T> out,
--- a/src/main/java/net/imagej/ops/filter/FilterNamespace.java
+++ b/src/main/java/net/imagej/ops/filter/FilterNamespace.java
@@ -51,6 +51,8 @@ import net.imglib2.type.Type;
 import net.imglib2.type.numeric.ComplexType;
 import net.imglib2.type.numeric.NumericType;
 import net.imglib2.type.numeric.RealType;
+import net.imglib2.view.composite.CompositeIntervalView;
+import net.imglib2.view.composite.RealComposite;
 
 import org.scijava.plugin.Plugin;
 
@@ -924,6 +926,35 @@ public class FilterNamespace extends AbstractNamespace {
 		@SuppressWarnings("unchecked")
 		final RandomAccessibleInterval<T> result = (RandomAccessibleInterval<T>) ops()
 				.run(Ops.Filter.PadShiftFFTKernel.class, in1, in2, fast);
+		return result;
+	}
+
+	// -- partial derivative --
+
+	/** Executes the "partial derivative" operation on the given arguments */
+	@OpMethod(op = net.imagej.ops.filter.derivative.PartialDerivativeRAI.class)
+	public <T extends RealType<T>> RandomAccessibleInterval<T> partialDerivative(final RandomAccessibleInterval<T> in, final int dimension) {
+		@SuppressWarnings("unchecked")
+		final RandomAccessibleInterval<T> result =
+			(RandomAccessibleInterval<T>) ops().run(net.imagej.ops.filter.derivative.PartialDerivativeRAI.class, in, dimension);
+		return result;
+	}
+
+	/** Executes the "partial derivative" operation on the given arguments */
+	@OpMethod(op = net.imagej.ops.filter.derivative.PartialDerivativeRAI.class)
+	public <T extends RealType<T>> RandomAccessibleInterval<T> partialDerivative(final RandomAccessibleInterval<T> out, final RandomAccessibleInterval<T> in, final int dimension) {
+		@SuppressWarnings("unchecked")
+		final RandomAccessibleInterval<T> result =
+			(RandomAccessibleInterval<T>) ops().run(net.imagej.ops.filter.derivative.PartialDerivativeRAI.class, out, in, dimension);
+		return result;
+	}
+
+	/** Executes the "partial derivative" operation on all dimensions */
+	@OpMethod(op = net.imagej.ops.filter.derivative.PartialDerivativesRAI.class)
+	public <T extends RealType<T>> CompositeIntervalView<T, RealComposite<T>> allPartialDerivatives(final RandomAccessibleInterval<T> in) {
+		@SuppressWarnings("unchecked")
+		final CompositeIntervalView<T, RealComposite<T>> result =
+			(CompositeIntervalView<T, RealComposite<T>>) ops().run(net.imagej.ops.filter.derivative.PartialDerivativesRAI.class, in);
 		return result;
 	}
 
--- a/src/main/java/net/imagej/ops/filter/derivative/PartialDerivativeRAI.java
+++ b/src/main/java/net/imagej/ops/filter/derivative/PartialDerivativeRAI.java
@@ -1,0 +1,179 @@
+/* #%L
+ * ImageJ software for multidimensional image processing and analysis.
+ * %%
+ * Copyright (C) 2014 - 2016 Board of Regents of the University of
+ * Wisconsin-Madison, University of Konstanz and Brian Northan.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * 
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ * #L%
+ */
+
+package net.imagej.ops.filter.derivative;
+
+import net.imagej.ops.Ops;
+import net.imagej.ops.special.chain.RAIs;
+import net.imagej.ops.special.computer.BinaryComputerOp;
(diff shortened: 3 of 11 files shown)


src/main/java/net/imagej/ops/create/CreateNamespace.java
```diff
@@ -543,6 +543,33 @@ public Object kernelLog(final Object... args) {
...
+	@OpMethod(op = net.imagej.ops.create.kernelSobel.CreateKernelSobelSeparated.class)
+	public <T extends ComplexType<T>> Img<T> kernelSobelSeparated() {
+		@SuppressWarnings("unchecked")
+		final Img<T> result = (Img<T>) ops().run(
+			net.imagej.ops.create.kernelSobel.CreateKernelSobelSeparated.class);
+		return result;
+	}
+
+	@OpMethod(op = net.imagej.ops.create.kernelSobel.CreateKernelSobelSeparated.class)
+	public <T extends ComplexType<T>> Img<T> kernelSobelSeparated(final Type<T> outType) {
```
dietzc: can't you use `final T outType` rather than `final Type<T> outType`? @ctrueden?

src/main/java/net/imagej/ops/filter/derivative/DirectionalDerivativeRAI.java
```diff
@@ -0,0 +1,97 @@
...
+
+		addOp = RAIs.binaryComputer(ops(), Ops.Math.Add.class, in(), in());
+
+		yConvolver = RAIs.computer(ops(), Ops.Filter.Convolve.class, in(), kernelY);
+		xConvolver = RAIs.computer(ops(), Ops.Filter.Convolve.class, in(), kernelX);
+
+		copyRAI = RAIs.computer(ops(), Ops.Copy.RAI.class, in());
+		createRAIFromRAI = RAIs.function(ops(), Ops.Create.Img.class, in());
+
+		IntervalView<T> filter = Views.rotate(kernelY, 0, 1);
```
dietzc: no need to create variable here. you can put `Views.rotate(..)` directly in the ops call.

src/main/java/net/imagej/ops/create/CreateNamespace.java
ctrueden: Yes, always use `T`, not `Type<T>`. I actually just fixed all of the known offenders with 9316184ca99e9ca34b05e414b3a80fb51e352340. Let's try to keep things clean!

src/main/java/net/imagej/ops/filter/hessian/HessianRAI.java
```diff
@@ -0,0 +1,79 @@
...
+	@SuppressWarnings("unchecked")
+	@Override
+	public void initialize() {
+		createRAIFromDim = Functions.unary(ops(), Ops.Create.Img.class, RandomAccessibleInterval.class,
+				Dimensions.class, Util.getTypeFromInterval(in()).createVariable());
+		createRAIFromRAI = RAIs.function(ops(), Ops.Create.Img.class, in());
+
+		derivativeComputers = new UnaryComputerOp[in().numDimensions()];
+		for (int i = 0; i < in().numDimensions(); i++) {
+			derivativeComputers[i] = RAIs.computer(ops(), Ops.Filter.DirectionalDerivative.class, in(), i);
```
dietzc: only works if `DirectionalDerivative` is also working for n-d...

src/main/java/net/imagej/ops/filter/sobel/SobelRAI.java
```diff
@@ -0,0 +1,73 @@
...
+
+		Sqr squareOp = ops().op(Ops.Math.Sqr.class, RealType.class, RealType.class);
+		squareMapOp = RAIs.computer(ops(), Ops.Map.class, in(), squareOp);
+		Sqrt sqrtOp = ops().op(Ops.Math.Sqrt.class, RealType.class, RealType.class);
+		sqrtMapOp = RAIs.computer(ops(), Ops.Map.class, in(), sqrtOp);
+		addOp = RAIs.binaryComputer(ops(), Ops.Math.Add.class, in(), in());
+
+		derivativeComputers = new UnaryComputerOp[in().numDimensions()];
+		for (int i = 0; i < in().numDimensions(); i++) {
+			derivativeComputers[i] = RAIs.computer(ops(), Ops.Filter.DirectionalDerivative.class, in(), i);
```
dietzc: sure that `Sobel`works like this in n-d?  See e.g. http://www.aravind.ca/cs788h_Final_Project/gradient_estimators.htm#idsb

dietzc:
@iarganda do you want to review?

iarganda:
Sure, I'll have a look as soon as possible! Thanks!

dietzc:
@iarganda @hornm any comments?

src/main/java/net/imagej/ops/filter/derivative/PartialDerivativesRAI.java
```diff
@@ -0,0 +1,51 @@
+package net.imagej.ops.filter.derivative;
```
stelfrich: License header is missing.

src/main/java/net/imagej/ops/filter/derivative/PartialDerivativesRAI.java
```diff
@@ -0,0 +1,51 @@
...
+import net.imglib2.type.numeric.RealType;
+import net.imglib2.view.Views;
+import net.imglib2.view.composite.CompositeIntervalView;
+import net.imglib2.view.composite.GenericComposite;
+import net.imglib2.view.composite.RealComposite;
+
+import org.scijava.plugin.Plugin;
+
+@Plugin(type = Ops.Filter.AllPartialDerivatives.class, name = Ops.Filter.AllPartialDerivatives.NAME)
+public class PartialDerivativesRAI<T extends RealType<T>, C extends GenericComposite<T>>
```
stelfrich: I could safely remove the declaration of `C`. Am I missing something?

src/main/java/net/imagej/ops/filter/derivative/PartialDerivativesRAI.java
```diff
@@ -0,0 +1,51 @@
...
+		extends AbstractUnaryFunctionOp<RandomAccessibleInterval<T>, CompositeIntervalView<T, RealComposite<T>>>
```
stelfrich: The output could be "generalized" to a `RandomAccessibleInterval<? extends Composite<T>>`.

src/main/java/net/imagej/ops/filter/derivative/PartialDerivativesRAI.java
```diff
@@ -0,0 +1,51 @@
...
+		createRAIFromRAI = RAIs.function(ops(), Ops.Create.Img.class, in());
+		derivativeComputers = new UnaryComputerOp[in().numDimensions()];
+		for (int i = 0; i < in().numDimensions(); i++) {
+			derivativeComputers[i] = RAIs.computer(ops(), Ops.Filter.PartialDerivative.class, in(), i);
+		}
+	}
+
+	@Override
+	public CompositeIntervalView<T, RealComposite<T>> compute1(RandomAccessibleInterval<T> input) {
+		List<RandomAccessibleInterval<T>> derivativeList = new ArrayList<>();
```
stelfrich: Just personal preference: I'd call it `derivatives`. This way, you could change it to something else (e.g. a Set) without the need for changing the variable name to keep consistency.

stelfrich:
Hey @EikeHeinz,
I have added some code-related comments to your latest commits. Some general remarks that I couldn't pin to a specific location in the commits:

- [ ] Missing license headers
- [ ] Rename the test to `PartialDerivativeFilterTest`

src/main/java/net/imagej/ops/create/kernelSobel/CreateKernelSobelSeparated.java
```diff
@@ -0,0 +1,69 @@
+package net.imagej.ops.create.kernelSobel;
```
stelfrich: Missing license header.

src/main/java/net/imagej/ops/filter/derivative/PartialDerivativeRAI.java
```diff
@@ -0,0 +1,149 @@
+package net.imagej.ops.filter.derivative;
```
stelfrich: Missing license header.

src/main/java/net/imagej/ops/filter/FilterNamespace.java
```diff
@@ -622,7 +624,7 @@ void correlate(final RandomAccessibleInterval<I> raiExtendedInput,
 			raiExtendedInput, raiExtendedKernel, fftInput, fftKernel, output,
 			performInputFFT, performKernelFFT);
 	}
-
+
```
stelfrich: According to the [coding style](http://imagej.net/Coding_style) empty lines should be empty, i.e. there shouldn't be any whitespaces (tabs, spaces, ...).

EikeHeinz:
Thank you for your comments @stelfrich, I'll look into them as soon as possible.
Note: I think, that the returned CompositeView doesn't contain all derivatives (I'm looking into this). Please keep that in mind, if you are planning to use it.

src/main/java/net/imagej/ops/filter/derivative/PartialDerivativeRAI.java
```diff
@@ -0,0 +1,149 @@
...
+		RandomAccessibleInterval<T> kernel = ops().create().kernelSobelSeparated();
+		// kernel A contains 1 2 1
+		IntervalView<T> kernelA = Views.hyperSlice(Views.hyperSlice(kernel, 3, 0), 2, 0);
+		// kernel B contains -1 0 1
+		IntervalView<T> kernelB = Views.hyperSlice(Views.hyperSlice(kernel, 3, 0), 2, 1);
+
+		// add dimensions to kernel if input has more than 2 dimensions to
+		// properly rotate the kernel
+		if (in().numDimensions() > 2) {
+			MixedTransformView<T> expandedKernelA = Views.addDimension(kernelA);
```
stelfrich: Generalize `expandedKernelA` to be `RandomAccessible<T>`.

src/main/java/net/imagej/ops/filter/derivative/PartialDerivativeRAI.java
```diff
@@ -0,0 +1,149 @@
...
+			for (int i = 0; i < in().numDimensions() - 3; i++) {
+				expandedKernelA = Views.addDimension(expandedKernelA);
+				expandedKernelB = Views.addDimension(expandedKernelB);
+			}
+			long[] dims = new long[in().numDimensions()];
+			for (int j = 0; j < in().numDimensions(); j++) {
+				dims[j] = 1;
+			}
+			dims[0] = 3;
+			Img<DoubleType> kernelInterval = ops().create().img(dims);
```
stelfrich: You don't seem to be using the created image? Replace that line with `Interval kernelInterval = new FinalInterval(dims);`

src/main/java/net/imagej/ops/filter/derivative/PartialDerivativeRAI.java
```diff
@@ -0,0 +1,149 @@
...
+	private UnaryComputerOp<RandomAccessibleInterval<T>, RandomAccessibleInterval<T>> kernelBConvolver;
+
+	private UnaryComputerOp<RandomAccessibleInterval<T>, RandomAccessibleInterval<T>>[] kernelAConvolverArray;
+
+	@SuppressWarnings("unchecked")
+	@Override
+	public void initialize() {
+		RandomAccessibleInterval<T> kernel = ops().create().kernelSobelSeparated();
+		// kernel A contains 1 2 1
+		IntervalView<T> kernelA = Views.hyperSlice(Views.hyperSlice(kernel, 3, 0), 2, 0);
```
stelfrich: The separated kernels could be `RandomAccessibleIntervals`s? Though you might want to get a second opinion here, I am still not too firm with `Views`..

src/main/java/net/imagej/ops/filter/derivative/PartialDerivativesRAI.java
```diff
@@ -0,0 +1,51 @@
...
+import net.imglib2.RandomAccessibleInterval;
+import net.imglib2.type.numeric.RealType;
+import net.imglib2.view.Views;
+import net.imglib2.view.composite.CompositeIntervalView;
+import net.imglib2.view.composite.GenericComposite;
+import net.imglib2.view.composite.RealComposite;
+
+import org.scijava.plugin.Plugin;
+
+@Plugin(type = Ops.Filter.AllPartialDerivatives.class, name = Ops.Filter.AllPartialDerivatives.NAME)
```
stelfrich: Missing documentation.

stelfrich:
> Note: I think, that the returned CompositeView doesn't contain all derivatives (I'm looking into this). Please keep that in mind, if you are planning to use it.

I won't be using it until some time next week. No worries 😄

ctrueden:
Thanks everyone for reviewing.

Regarding the license headers, I am less worried about these, since they can be easily added after the fact using `mvn license:update-file-header`. The other comments, though, would be good to resolve. I especially would like to see no WIP commits in the commit history. Every commit needs to compile with passing tests.

src/main/java/net/imagej/ops/filter/derivative/PartialDerivativesRAI.java
EikeHeinz: You're right, removed it.

src/main/java/net/imagej/ops/filter/derivative/PartialDerivativesRAI.java
ctrueden: Also, you don't have to write `name = ...`—it will be inferred from the plugin type.

stelfrich:
I stumbled across an issue with using `PartialDerivativesRAI` on the [shape-index-map branch](https://github.com/imagej/imagej-ops/tree/shape-index-map). There, I am reusing an instance of `PartialDerivativesRAI` for the second-order partial derivatives (see [DefaultHessian#L95](https://github.com/imagej/imagej-ops/commit/19286fcd133292e56fce12e76094e33205d85be1#diff-48f3439ec116b8b17310596132b19f37R95)). When used the second time, the input will be of type `ConvertedRandomAccessibleInterval` which cannot be cast to `Img` in [`PartialDerivative#createOutput`](https://github.com/imagej/imagej-ops/blob/sobel-hessian-filter/src/main/java/net/imagej/ops/filter/derivative/PartialDerivativeRAI.java#L148).

I have added a [dirty hack](https://github.com/imagej/imagej-ops/commit/19286fcd133292e56fce12e76094e33205d85be1#diff-c2c9bf09cc635e9d62649b237be3ddd0R128) on that branch to circumvent a matching of `CreateImgFromImg` to `Ops.Create.Img`. But doesn't solve the underlying issue that `Ops.Create.Img` is matched too specifically to be reused later on. Maybe @dietzc has an opinion on that?

dietzc:
why do we need to cast it to `Img` at all? This shouldn't ever be required.

stelfrich:
> why do we need to cast it to `Img` at all? This shouldn't ever be required.

It is done implicitly when `PartialDerivatives` is fed an `Img`. In [PartialDerivative#L128](https://github.com/imagej/imagej-ops/pull/349/files#diff-c2c9bf09cc635e9d62649b237be3ddd0R128) the createop is initialized with the concrete input instead of a `Class`. Hence, the concrete type of createOp will be `CreateImgFromImg`.

If I force that createOp to be `CreateImgFromRAI` everything works out fine due to its more general but applicable nature. For me that raises the question if a matching can be too specific?

ctrueden:
@EikeHeinz Any updates?

EikeHeinz:
I'm working on the last issues. When I solved them I'll rebase the branch to resolve the conflicts and clean up the commit history. Probably during the next week.

ctrueden:
@EikeHeinz Awesome, thank you!

stelfrich:
@EikeHeinz Have you gotten around to working on the issues?

EikeHeinz:
@stelfrich Yes, I think I have resolved most or all issues, except the one I'm having with the CompositeView. I'll look into that one again tomorrow and hopefully fix it.

EikeHeinz:
@stelfrich Seems like the rebase requires some more work to do. I'm going to solve the new issues over the next few days.

stelfrich:
@EikeHeinz Were you able to resolve the issues and compute all derivatives? If you need a second opinion/review, I'd be glad to help!

dietzc:
did you see: https://github.com/imglib/imglib2-algorithm/pull/29?  does this help?

EikeHeinz:
Sorry for the late answer.
@stelfrich yes I've been able to resolve the remaining issues. I'm currently working on some of the points you've mentioned that I haven't fixed until now.
@dietzc Thanks for the info. I'll have a look.

src/main/java/net/imagej/ops/filter/FilterNamespace.java
```diff
@@ -500,7 +503,27 @@
 				performInputFFT, performKernelFFT);
 		return result;
 	}
+	
+	// -- directional derivative --
+	
+	/** Executes the "directional derivative" operation on the given arguments */
+	@OpMethod(op = net.imagej.ops.filter.derivative.PartialDerivativeRAI.class)
+	public <T extends RealType<T>> RandomAccessibleInterval<T> directionalDerivative(final RandomAccessibleInterval<T> in, final int dimension) {
```
stelfrich: This one and the next method are a left-over from the rename to `partialDerivative`, I guess? If so, they can now be removed.

src/main/java/net/imagej/ops/create/kernelSobel/CreateKernelSobelSeparated.java
```diff
@@ -0,0 +1,70 @@
...
+ * 
+ * @author Eike Heinz, University of Konstanz
+ *
+ * @param <T>
+ *            type of input
+ */
+
+@Plugin(type = Ops.Create.KernelSobel.class, name = Ops.Create.KernelSobel.NAME)
+public class CreateKernelSobelSeparated<T extends ComplexType<T>>
+		extends AbstractNullaryFunctionOp<RandomAccessibleInterval<T>> {
```
stelfrich: `implements Ops.Create.KernelSobel`

EikeHeinz:
@stelfrich Thanks for your comments and help.
@hornm @dietzc I've made a rebase and cleaned up the commit history. Can one of you have a look please?

ctrueden:
Rebased over the latest master to resolve conflicts.

I almost merged this, but there is one inconsistency I want to fix first: is the op name `create.kernelSobel` Or `create.kernelSobelSeparated`? We need to make the naming consistent before we can merge it. @EikeHeinz Any preference?

EikeHeinz:
We can name it `kernelSobel`, because it's the shorter call, but if there are plans to add the not separated Sobel Kernel we might also call it `kernelSobelSeparated`. I have no preference.

ctrueden:
I tried to research the difference between separable and non-separable Sobel operators, but wikipedia just describes it as separable. I assume the separable version is the original/canonical one? In which case, calling it only `kernelSobel` here makes sense. If a non-separable version is added, we could call it `kernelSobelNonSeparable` or some other distinguishing name.

ctrueden:
Woo hoo, merged! Thank you @EikeHeinz!

PR #350, opened 2016-02-11
Distinguish MapUnaryComputers and MapBinaryInplaces

LeonYang5114:
Both categories of the maps have the same signature as (II/RAI, II/RAI, op). If the providing op is both a unary computer and a binary inplace op (e.g. the NumericTypeBinaryMath ops), the matching for the two
categories of maps will conflict. Now the unary computer maps have higher priority. If want to use binary inplace maps, Inplaces.binary(...) should be called.

### DIFF
--- a/src/main/java/net/imagej/ops/map/MapIIAndIIInplace.java
+++ b/src/main/java/net/imagej/ops/map/MapIIAndIIInplace.java
@@ -44,7 +44,7 @@ import org.scijava.plugin.Plugin;
  * @author Leon Yang
  * @param <EA> element type of inputs + outputs
  */
-@Plugin(type = Ops.Map.class, priority = Priority.HIGH_PRIORITY + 1)
+@Plugin(type = Ops.Map.class, priority = Priority.LOW_PRIORITY)
 public class MapIIAndIIInplace<EA> extends
 	AbstractMapBinaryInplace<EA, IterableInterval<EA>> implements Contingent
 {
--- a/src/main/java/net/imagej/ops/map/MapIIAndIIInplaceParallel.java
+++ b/src/main/java/net/imagej/ops/map/MapIIAndIIInplaceParallel.java
@@ -47,7 +47,7 @@ import org.scijava.plugin.Plugin;
  * @author Leon Yang
  * @param <EA> element type of inputs + outputs
  */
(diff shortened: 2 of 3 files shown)


dietzc:
@ctrueden @LeonYang5114 I'm also a bit puzzled, which of the two arguments would be used for the `inplace` in case of a binary `inplace` (`mutate1` vs `mutate2`)? Is there a convention? Maybe we should name the `Map` more explicit? I'm not sure.

ctrueden:
The intent of the framework is that no op should be considered multiple arities at once. Use separate ops for that.

LeonYang5114:
@dietzc 
If `ops.run(...)` is used, `mutate1` is always called (by convention). If `mutate2` want to be called, an `BinaryInplaceOp` instance need to be retrieved explicitly:
```
BinaryInplaceOp<II, II> op = Inplaces.binary(ops, Ops.Map.class, II.class, helperOp);
```
(this is expected to work, but as I am writing this comment, I find a bug in `MapIIAndIIInplaceParallel#conforms` which leads to NPE when `in1()` or `in2()` is null)
or this:
```
BinaryInplaceOp<II, II> op = Inplaces.binary(ops, Ops.Map.class, firstII, secondII, helperOp);
```

dietzc:
> The intent of the framework is that no op should be considered multiple arities at once. Use separate ops for that.

so the op should be split up? do I understand this correctly @LeonYang5114, @ctrueden?

ctrueden:
@dietzc wrote:
> so the op should be split up?

No, my comment was insufficiently informed.

I finally made time to actually look at this patch. What Leon said was:

@LeonYang5114 wrote:
> If the providing op is both a unary computer and a binary inplace op (e.g. the NumericTypeBinaryMath ops)

I looked at `NumericTypeBinaryMath` but the ops it contains are actually only `BinaryHybridCFI` which is fine. So they are arity 2. But all `BinaryComputerOp`s also extend `UnaryComputerOp`, which fulfills what Leon is saying above: a `BinaryHybridCFI` has the API for `UnaryComputerOp` as well as `BinaryInplaceOp`, which causes the clash in question.

In any case, these changes seem fine! (I say without having tested it.)

PR #352, opened 2016-02-13
Create NoParam

lnyng:
NoParam is a null instance of a given class. It behaviors almost exactly the same as a null object except that it still retains the class information. It is useful for chain initialization when a class object is used to do the matching at the top level.

@ctrueden All the tests are passing for now, but I think this NoParam could very likely introduce some bugs. I am also not sure if the `NoParamService` is necessary. Any comment would be great.

### DIFF
--- a/src/main/java/net/imagej/ops/DefaultNoParamService.java
+++ b/src/main/java/net/imagej/ops/DefaultNoParamService.java
@@ -1,0 +1,131 @@
+/*
+ * #%L
+ * ImageJ software for multidimensional image processing and analysis.
+ * %%
+ * Copyright (C) 2014 - 2016 Board of Regents of the University of
+ * Wisconsin-Madison, University of Konstanz and Brian Northan.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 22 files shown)


stelfrich:
Just my 2 cents: `OpUtils.isNullParam()` instead of `== null` seems like a harsh change especially since we can't warn people that with this implementation `in() == null` and `out() == null` effectively evaluates to `false`. The question is, in which instances we use `out() == null` and `in() == null`, respectively, and if those cases can be eliminated. The aforementioned concerns, however, remain.

`Optionals` aren't really of use either because we can't preserve the class type. This way we could at least have changed eliminated the `== null` checks in favor of a more standard approach.

dietzc:
I'm not sure about this. I share the concerns of @stelfrich. Shouldn't this is part of the work on `generic`s which is done by @ctrueden? There we introduce `Token`s for generic type matching.

dietzc:
@ctrueden any opinion on this?

ctrueden:
I will have an opinion eventually, but I've been too busy to dig into it yet.

ctrueden:
The `Nil`, to be introduced with generic type matching, will solve this.

PR #356, opened 2016-02-15
Add support for travis-ci

gab1one:
This adds the `.travis.yml` file needed to build imagej-ops on travis.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,7 @@
 	</parent>
 
 	<artifactId>imagej-ops</artifactId>
-	<version>0.27.1-SNAPSHOT</version>
+	<version>0.28.1-SNAPSHOT</version>
 
 	<name>ImageJ Ops</name>
 	<description>ImageJ Operations: a framework for reusable algorithms.</description>


dietzc:
should be use imagej-ops as a test project for travis @ctrueden? I like the idea having a continuous integration also for branches.

hinerm:
>I like the idea having a continuous integration also for branches.

Jenkins is capable of building branches and [pull requests](https://wiki.jenkins-ci.org/display/JENKINS/GitHub+pull+request+builder+plugin). I like Travis in general, but what is the intended value added by using Travis? Would it be less effort to tweak our established Jenkins jobs?

ctrueden:
## Arguments for Travis ##
* It is great to be able to tweak builds merely by pushing `travis.yml` files to Git. This tracks how the build is being done at _that time_ in the source, whereas right now, our Jenkins configuration is not "in sync" with the Git revisions. (While the revision hash is present in the console output of past CI builds, those builds can nonetheless not be reproduced exactly and reliably—the Jenkins job might have been edited since the build took place.)
* If Travis can fully replace the standard CI build job of each project, it would drastically reduce the number of jobs listed on our Jenkins. Hence, the job list would be easier to read, less overwhelming.
* Using Travis for our core projects is "eating the community dog food" in that it provides an extensible model for third parties to use CI with their own projects, in exactly the same way we do (minus the snapshot deploys in some cases, perhaps).

## Arguments against Travis ##
* Potential delays in building.
* Travis needs credentials to the ImageJ Maven repository for snapshot deployment.
* Reliance on a third-party service which we don't control.
* _Much_ less configurable than our Jenkins. Very limited in what we can do.
* For several reasons, Travis can never fully replace our Jenkins setup. So it does not reduce our maintenance burden.

## Neutral comments ##
* I agree with @hinerm that wanting to test PRs using CI is not an argument for or against Travis. We can do it either way.

ctrueden:
Anyone have any following comments or discussion? @stelfrich? @gab1one?

stelfrich:
**tl;dr** I think it would be a good idea to have CI for PRs independent of the implementation. We don't need Travis as long as the Jenkins infrastructure can handle the load (and is maintained). 

---
It's hard for me to evaluate most of the _cons_ since I am not familiar with the (extent of the) current Jenkins configuration.

> Travis needs credentials to the ImageJ Maven repository for snapshot deployment.

With regard to https://docs.travis-ci.com/user/encryption-keys/ I don't think that this should be too much of a problem.

> Harder to execute jobs which are downstream from the standard snapshot builds.

@ctrueden Could you provide an example for such a job? Can they be implemented as shell scripts that we can put in the repositories? Those could be [executed easily](https://docs.travis-ci.com/user/customizing-the-build/#Implementing-Complex-Build-Steps) from Travis (although I think you are aware of that..).

ctrueden:
@gab1one Why did you close this without explanation?

gab1one:
I reset the master branch of my fork to imagej-ops/master, which autoclosed the pullrequest.
So this was an accident, proves that one should always work on a topic branch :smile:

ctrueden:
Ah, I see. Yeah, topic branches for sure.

stelfrich:
Just for reference: this PR has been superseded by https://github.com/imagej/imagej-ops/commit/d85c4204fc052ba8bf81df476e84d955d5f8aa7b.

PR #357, opened 2016-02-16
Fix alternative parameter in MapNeighborhood

stelfrich:
Alternative input declaration for ops matching was `RandomAccessibleInteval` instead of `IterableInterval` (if the input of `MapNeighborhood` had already been set).

See Gitter:
> @LeonYang5114: In [MapNeighborhood](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/map/neighborhood/MapNeighborhood.java#L74), should the highlighted line use `IterableInterval.class` instead?
> @stelfrich: I think that line is correct, since `MapNeighborhood` is typed on `RandomAccessibleInterval` inputs, while outputs are `IterableInterval`.
> @LeonYang5114: but that line of code is _matching a helper op_ (i.e. the `map` we get in that line is not a `MapNeighborhood`, but just a simple `MapUnaryComputers.IIToII`), and [`shape.neighborhoodsSafe(in())`](https://github.com/imglib/imglib2-algorithm/blob/master/src/main/java/net/imglib2/algorithm/neighborhood/Shape.java#L150) actually returns an `IterableInterval<Neighborhood<T>>`. 
> @stelfrich: you are right, sorry for the confusion!

### DIFF
--- a/src/main/java/net/imagej/ops/map/neighborhood/MapNeighborhood.java
+++ b/src/main/java/net/imagej/ops/map/neighborhood/MapNeighborhood.java
@@ -71,7 +71,7 @@ public class MapNeighborhood<I, O> extends
 	public void initialize() {
 		map = (UnaryComputerOp) Computers.unary(ops(), Map.class,
 			IterableInterval.class, in() != null ? shape.neighborhoodsSafe(in())
-				: RandomAccessibleInterval.class, getOp());
+				: IterableInterval.class, getOp());
 	}
 
 	@Override


stelfrich:
- [ ] if tests pass but behavior is erroneous, a new test needs to be added to test the fixed behavior, to avoid regressions (@ctrueden, Gitter)

LeonYang5114:
I agree with you that this behavior is actually not erroneous, but just that the converter service is too "smart."

Do you think always use `IterableInterval.class` would be safe in this case? so we never call `shape.neighborhoodSafe(in())` even `in()` is non-null.

stelfrich:
> Do you think always use `IterableInterval.class` would be safe in this case? so we never call `shape.neighborhoodSafe(in())` even `in()` is non-null.

We would (obviously) be losing information about the return type of `shape.neighborhoodSafe()`. I don't see that additional information used during matching at the moment. We would, however, not be able to have a highly specialized `MapIIofNeighborhoodsToII` if we were to take into account the type parameters in future implementations of Ops matching. I would opt for keeping the `shape.neighborhoodSafe(in())` call.

All in all, I would suggest to merge this PR without additional tests. In general, `RAI`s can be converted to `II`s regardless of their type parameters. In this very instance, `IItoIIParallel` is matched even if there is a `RAItoIIParallel` due to the higher priority. Hence, I'd say that both versions are equally correct, the one in this PR is, however, more intuitive to understand.

What's your take on it, @ctrueden, @dietzc?

dietzc:
writing a test here is super hard, because you would have to write a buggy `Op`. So I'll just merge it as is.

PR #358, opened 2016-02-18
create morphology namespace

LeonYang5114:
@ctrueden 
Here is some rudimentary work on creating the morphology namespace.

A couple questions:
1. Can we come up with better naming for the ops?
2. Do we need the `Inplace` functionalities for each of the morphology operation? (e.g. [`Erosion#erodeInPlace`](https://github.com/imglib/imglib2-algorithm/blob/master/src/main/java/net/imglib2/algorithm/morphology/Erosion.java#L899-L910), which is actually not really "in place," since it creates a new `Img` and copies it back to the given `RAI`)
3. Do we want to combine the `Single` and `Multi` shape versions of each morphology ops?
4. Do we want to use `Map`, `Chunck` etc to re-implement the code in imglib2?

To do:
- [x] Write test cases
- [x] Include the `TopHat` and `BlackTopHat` operations

**==== Update ====**
This branch is now ready for review. It does two things:

1. Refactor the `map.neighborhood` package. This is done because I think the two `MapNeighborhood` should logically share the same interface, and creating that interface makes it possible to reuse a `MapNeighborhood` with different `Shape`s, which is necessary for morphology operations with a `List` of `Shape`s.

2. Create the morphology namespace and implementations. The implementation is adapted from imglib2, but mostly relies on the Op framework (i.e. `MapNeighborhood`, `Ops.Create.Img`). 

Some notes:
All the morphology ops requires extended RAI (or one that is large enough), but they do not have an optional out of bound factory parameter. Not sure if we really want that.
Only the Dilate and Erode have `Default` ops, which operate on a single shape. Other morphology ops do not have that because `SomeType` can be converted to `List<SomeType>` with a single element, and only the Dilate and Erode would have different implementations between the `Default` and `List` versions.

@ctrueden @dietzc any comment?

### DIFF
--- a/src/main/java/net/imagej/ops/morphology/Morphologies.java
+++ b/src/main/java/net/imagej/ops/morphology/Morphologies.java
@@ -1,0 +1,85 @@
+/*
+ * #%L
+ * ImageJ software for multidimensional image processing and analysis.
+ * %%
+ * Copyright (C) 2014 - 2016 Board of Regents of the University of
+ * Wisconsin-Madison, University of Konstanz and Brian Northan.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * 
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ * #L%
+ */
+
+package net.imagej.ops.morphology;
+
+import java.util.List;
+
+import net.imglib2.Interval;
+import net.imglib2.algorithm.morphology.MorphologyUtils;
+import net.imglib2.algorithm.neighborhood.Neighborhood;
+import net.imglib2.algorithm.neighborhood.Shape;
+import net.imglib2.type.logic.BitType;
+
+/**
+ * Utility class for morphology operations.
+ * 
+ * @author Leon Yang
+ */
+public class Morphologies {
+
+	private Morphologies() {
+		// NB: Prevent instantiation of utility class.
+	}
+
+	/**
+	 * Computes the min coordinate and the size of an {@link Interval} after
+	 * padding with a list of {@link Shape}s in a series morphology operations.
+	 * 
+	 * @param source the interval to be applied with some morphology operation
+	 * @param shapes the list of Shapes for padding
+	 * @return a size-2 array storing the min coordinate and the size of the
+	 *         padded interval
+	 */
+	public static final long[][] computeMinSize(final Interval source,
+		final List<Shape> shapes)
+	{
+
+		final int numDims = source.numDimensions();
+		final long[] min = new long[numDims];
+		final long[] size = new long[numDims];
+
+		for (int i = 0; i < numDims; i++) {
+			min[i] = source.min(i);
+			size[i] = source.dimension(i);
+		}
+
+		for (final Shape shape : shapes) {
+			final Neighborhood<BitType> nh = MorphologyUtils.getNeighborhood(shape,
+				source);
+			for (int i = 0; i < numDims; i++) {
+				min[i] += nh.min(i);
+				size[i] += nh.dimension(i) - 1;
+			}
+		}
+
+		return new long[][] { min, size };
+	}
+}
--- a/src/main/java/net/imagej/ops/morphology/close/ListClose.java
+++ b/src/main/java/net/imagej/ops/morphology/close/ListClose.java
@@ -1,0 +1,112 @@
+/*
+ * #%L
+ * ImageJ software for multidimensional image processing and analysis.
+ * %%
+ * Copyright (C) 2014 - 2016 Board of Regents of the University of
+ * Wisconsin-Madison, University of Konstanz and Brian Northan.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * 
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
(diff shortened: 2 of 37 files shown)


LeeKamentsky:
I was thinking of contributing some of CellProfiler's 3x3 2-D binary morphology operations. These are based on 512-element lookup tables for all possible binary 3x3 combinations. This page: http://www.broadinstitute.org/~leek/CPmanual/Morph.html gives you a list. We use branchpoints and endpoints all the time for analyzing skeletons of neurites. A lot of them have limited utility: "life" is CellProfiler's easter egg and one of my secret hopes is that someone will publish a paper where they used it.

Porting the code is dead simple: I'd write a Python script to print each of the tables as some sort of Java boilerplate and then write some abstract class to compute the operation with an override method that would supply the table for each class.

dietzc:
> 2. Do we need the Inplace functionalities for each of the morphology operation? (e.g. Erosion#erodeInPlace, which is actually not really "in place," since it creates a new Img and copies it back to the given RAI)

I vote against having `Inplace` ops, as all kinds of `Neighbood`-based operations can't be calculated inplace.

> Do we want to use Map, Chunck etc to re-implement the code in imglib2?

I would use it. In particular I wood use the `MapNeighborhood` functionality. https://github.com/tinevez/imagej-ops/commit/bfee6fe6fe7b69ebc59ba5c0b1b30e926e997cb7 was a good start.

> Do we want to combine the Single and Multi shape versions of each morphology ops?

What does multi-shape functionality mean? Applying various shapes to the same `Op`? Or decomposing a shape into two single shape for approximation? I'd make a version of both.

> Porting the code is dead simple: I'd write a Python script to print each of the tables as some sort of Java boilerplate and then write some abstract class to compute the operation with an override method that would supply the table for each class.

Sounds awesome. So we can special-case on these ops?

LeeKamentsky:
Re special case: Most of the CellProfiler morphological ops could be generalized to N-D, but the table lookup mechanism is prohibitively expensive to do for cases > 2D. For 3+D, I would have to think about each and my brain might explode :wink:.

ctrueden:
> I vote against having `Inplace` ops

Yes, I am against individual ops implementing `Inplace` interfaces when under the hood they are actually doing functional+copy. We can write very general function :arrow_right: inplace adapters for that, if the need arises.

LeonYang5114:
@LeeKamentsky Is there anything I can do to cooperate with your plan?

LeeKamentsky:
@LeonYang5114  If I were to rough-out an implementation of each on a branch, it would be nice to be able to have someone who's more involved in Ops make it conform better to the architecture. I'd like to just write the code and check that it works, then let someone else handle it from there.

LeonYang5114:
@LeeKamentsky I would be more than happy to help. Just let me know once you start working on that.

dietzc:
> Re special case: Most of the CellProfiler morphological ops could be generalized to N-D, but the table lookup mechanism is prohibitively expensive to do for cases > 2D. For 3+D, I would have to think about each and my brain might explode. 

:+1: I don't want your brain to explode. anyway, we can have specialized, fast 2d, 3d versions based on your implementation. many thanks!

src/main/java/net/imagej/ops/morphology/Morphologies.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.List;
+
+import net.imglib2.Interval;
+import net.imglib2.algorithm.morphology.MorphologyUtils;
+import net.imglib2.algorithm.neighborhood.Neighborhood;
+import net.imglib2.algorithm.neighborhood.Shape;
+import net.imglib2.type.logic.BitType;
+
+/**
+ * Adapted from {@link net.imglib2.algorithm.morphology.MorphologyUtils}
```
ctrueden: Please add an `@author` tag for this.

Also, can you briefly explain what this brings to the table beyond `MorphologyUtils`? Would it make sense to merge these methods back into the `MorphologyUtils` class itself, so that we don't need this class in Ops?

src/main/java/net/imagej/ops/morphology/Morphologies.java
```diff
@@ -0,0 +1,111 @@
...
+ */
+public class Morphologies {
+
+	private Morphologies() {
+		// NB: Prevent instantiation of utility class.
+	}
+
+	public static final long[][] computeMinSize(final Interval source,
```
ctrueden: If we keep this method here, please add javadoc explaining its function.

src/main/java/net/imagej/ops/morphology/Morphologies.java
```diff
@@ -0,0 +1,111 @@
...
+		for (int d = 0; d < ndims; d++) {
+			final long d1 = d < sourceDims ? source.dimension(d) : 1;
+			final long d2 = d < nhDims ? nh.dimension(d) : 1;
+			size[d] = d1 + d2 - 1;
+		}
+
+		return new long[][] { min, size };
+	}
+
+	public static final long[][] computeMinSize(final Interval source,
```
ctrueden: If we keep this method here, please add javadoc explaining its function.

ctrueden:
@LeonYang5114 You mentioned that you weren't sure the 2d4b1cd and 52b651f changes are good. Can you briefly explain your reservations?

@dietzc What do you think? Do you like these changes?

src/main/java/net/imagej/ops/morphology/Morphologies.java
LeonYang5114: `MorphologyUtils` is good enough (although I think it might be coded in a more concise way). The problem is that `MorphologyUtils` has package visibility, so that I cannot use it in `ops`. Any workaround?

ctrueden: > Any workaround?

File a PR against `imglib2-algorithm` making it public?

LeonYang5114:
@ctrueden Since the two commits make deep changes to the `mapNeighborhood` package, many ops are be involved (and some probably under development in other branches).

I also add an interface to abstract out the `getShape` method. I notice that it needs to extend the `UnaryComputerOp` to work. If this is good, I might also want to apply this pattern to the other `map` ops, i.e. make `MapBinaryComputer` extends `BinaryComputerOp` as well.

dietzc:
> @dietzc What do you think? Do you like these changes?

looks good to me. Just wondering if it would make sense to make `MapNeighborhood` a `BinaryOp` (second parameter `Shape`).  There are cases where you want to apply the same op on several variants of `Shape`s... any opinion?

src/main/java/net/imagej/ops/morphology/close/ListClose.java
```diff
@@ -0,0 +1,112 @@
...
+import net.imglib2.type.numeric.RealType;
+import net.imglib2.util.Util;
+import net.imglib2.view.Views;
+
+import org.scijava.plugin.Plugin;
+
+/**
+ * Computes the closing of a {@link RandomAccessibleInterval} using a
+ * {@link List} of {@link Shape}s. It is the caller's responsibility to provide
+ * a {@link RandomAccessibleInterval} with enough padding for the output.
```
dietzc: can't we have a default `OutOfBoundsFactory`?

LeonYang5114:
@dietzc Sounds a good idea. I will try it right now.

dietzc:
I'm not sure if I understand why we need `List<Shape>` signatures in the namespace? I think `List<Shape>` is only used internally for an approximate but faster implementation of several of the `Shape`s?! Don't we want to have this optimization hidden from the user? For example if I call a certain `Shape` then the `Op` internally recognizes that this `Shape` can somehow be decomposed and applies it. I don't see the use-case for a default user to use the signatures with `List<Shape>`, but I don't have too much experience with morph ops.

src/main/java/net/imagej/ops/morphology/close/ListClose.java
LeonYang5114: What if the input is large enough so that no out of bound access is needed? would that produce unexpected result? I am still not quite clear about how `OOBF` works

dietzc: It shouldn't affect the computation results.

LeonYang5114:
I definitely have less experience on morph than you. The reason I keep that is because I migrate those ops almost directly from imglib2-algorithm, and they have the `List<Shape>` signature in all morph operations. We could easily remove them, but I think we might want some discussion.

src/main/java/net/imagej/ops/morphology/close/ListClose.java
stelfrich: > It shouldn't affect the computation results.

Wait. That implies that the `OOBF` does not jump in if the underlying `source` is defined outside of the `Interval`?

dietzc: as far as I understand the concept, it's always source first, yes. but one should definitively test it.

dietzc:
Ok I had a look at imglib2-algorithm again. So it seems that https://github.com/imglib/imglib2-algorithm/blob/master/src/main/java/net/imglib2/algorithm/morphology/StructuringElements.java helps you decomposing your `Shape`s into `List<Shape>`. However, for certain `Shape`s we can automatically do this (e.g. https://github.com/imglib/imglib2-algorithm/blob/master/src/main/java/net/imglib2/algorithm/morphology/StructuringElements.java#L400) without loss of precision, for other `Shape`s we can only approximate the result (https://github.com/imglib/imglib2-algorithm/blob/master/src/main/java/net/imglib2/algorithm/morphology/StructuringElements.java#L65).

 Therefore the question is how we design the interfaces in `Ops`: Do we want to have something like `RectangleDialate` and `RectangleErode` such that we can override the naive implementation with the special (decomposing) one (and maybe with a flag: `approximateForSpeedup`?!), or should we simply keep everything as is and it's users responsibility to make things fast by using the `StructuringElements` helper?

Optimally, we would have something which detects that is a certain, decomposable `Shape` is passed to the `Erode`/`Dilate` and/or the user passed an optional flag `approximateForSpeedup` and the `Erode` and `Dilate` recognize that and either execute a method with `List<Shape>` or `Shape`. To accomplish this we would need to be able to register `MorphologicalShapeDecomposer` (better names welcome) per `Shape` which is used in the generic `Erode` / `Dilate`. Does this make any sense? @tinevez?

tinevez:
IMHO I would always keep these methods with two inputs: the source to operate on and the structuring element. 
Then it is the user responsibility to generate a structuring element with or without decomposition.Trying to automate the generate of a strel with flags is dangerous. As you pointed out, certain decomposition are only approximation, and sometimes a decomposition is penalising in time.

dietzc:
> IMHO I would always keep these methods with two inputs: the source to operate on and the structuring element. 

Alright! Thanks. we can add magic under the hood later, too.

LeonYang5114:
A few points have been addressed:
1. `Morphologies` has added author information and documentations. The corresponding method in `imglib2-algorithm` is changed to public recently, but no recent release has been made.
2. `MapNeighborhood` is now a `BinaryComputerOp` with `Shape` as the second input. It breaks some code, since now the `Shape` needs to go before the `Op` parameter, but the change is minimal. 
3. optional `OutOfBoundsFactory` is added to `Dilate` and `Erode`, and they work well.

I am not sure what is our final decision on the `List<Shape>` signature. My understanding is that we want to keep it for now, right? Anyway, if there is anything that needs improvement, feel free to comment.

@dietzc @ctrueden

LeonYang5114:
The branch is rebased to master and updated to use the util method in imglib2-algorithm. @ctrueden

PR #336, opened 2016-01-26
Major iteration of filter design

bnorthan:
This is a significant iteration of the filter (fft, convolution,
    deconvolution) design.  Due to concurrent changes in master, all the
commits were squashed into a single large commit.  This commit
refactors the code to work towards the following general goals:

1.  Use RandomAccessibleInterval instead of Img as op inputs.
2.  Favor composition over inheritance where possible.
3.  Derive from nullary/unary/binary functions/computers/hybrids.
4.  Perform op matching in initialize.

PR #359, opened 2016-02-19
Create fill

LeonYang5114:
As @dietzc commented in #336, we might want to have some fill ops in the image namespace, which is the goal of this branch.

Some infrastructures are implemented for the fill ops to be possible (MapNullaryComputer, Assign).

One problem is the naming of the fill ops: DefaultFillParallel is actually doing parallel filling implicitly, but I cannot figure out a better naming, and not sure FillInterval should be called FillRAI instead.

Any comment is highly appreciated.

### DIFF
--- a/src/main/java/net/imagej/ops/image/ImageNamespace.java
+++ b/src/main/java/net/imagej/ops/image/ImageNamespace.java
@@ -185,6 +185,25 @@ public class ImageNamespace extends AbstractNamespace {
 		return result;
 	}
 
+	// -- fill --
+
+	/** Executes the "fill" operation on the given arguments. */
+	@OpMethod(op = net.imagej.ops.Ops.Image.Fill.class)
+	public Object fill(final Object... args) {
+		return ops().run(Ops.Image.Fill.class, args);
+	}
+
+	/** Executes the "fill" operation on the given arguments. */
+	@OpMethod(op = net.imagej.ops.image.fill.DefaultFill.class)
+	public <T extends Type<T>> Iterable<T> fill(final Iterable<T> out,
+		final T in)
+	{
+		@SuppressWarnings("unchecked")
+		final Iterable<T> result = (Iterable<T>) ops().run(
+			net.imagej.ops.Ops.Image.Fill.class, out, in);
+		return result;
+	}
+
 	// -- histogram --
 
 	/** Executes the "histogram" operation on the given arguments. */
--- a/src/main/java/net/imagej/ops/map/MapNullaryII.java
+++ b/src/main/java/net/imagej/ops/map/MapNullaryII.java
@@ -1,0 +1,37 @@
+
+package net.imagej.ops.map;
+
+import net.imagej.ops.Ops;
+import net.imagej.ops.thread.chunker.ChunkerOp;
+import net.imagej.ops.thread.chunker.CursorBasedChunk;
+import net.imglib2.IterableInterval;
+
+import org.scijava.Priority;
+import org.scijava.plugin.Plugin;
+
+/**
+ * A paralleled version of {@link MapNullaryComputer} over an
+ * {@link IterableInterval}.
+ * 
+ * @author Leon Yang
+ * @param <O> element type of outputs
+ */
+@Plugin(type = Ops.Map.class, priority = Priority.LOW_PRIORITY + 1)
+public class MapNullaryII<O> extends
+	AbstractMapNullaryComputer<O, IterableInterval<O>>
+{
+
+	@Override
+	public void compute0(IterableInterval<O> output) {
+		ops().run(ChunkerOp.class, new CursorBasedChunk() {
+
+			@Override
+			public void execute(final int startIndex, final int stepSize,
+				final int numSteps)
+			{
+				Maps.map(output, getOp(), startIndex, stepSize, numSteps);
+			}
+		}, output.size());
+	}
+
+}
--- a/src/main/java/net/imagej/ops/OpEnvironment.java
+++ b/src/main/java/net/imagej/ops/OpEnvironment.java
@@ -57,6 +57,7 @@ import net.imagej.ops.math.MathNamespace;
 import net.imagej.ops.special.SpecialOp;
 import net.imagej.ops.special.UnaryOutputFactory;
 import net.imagej.ops.special.computer.BinaryComputerOp;
+import net.imagej.ops.special.computer.NullaryComputerOp;
 import net.imagej.ops.special.computer.UnaryComputerOp;
 import net.imagej.ops.special.inplace.BinaryInplace1Op;
 import net.imagej.ops.special.inplace.BinaryInplaceOp;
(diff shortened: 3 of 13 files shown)


dietzc:
I added some comments. many thanks @LeonYang5114!!

LeonYang5114:
@dietzc I created the FillInterval just for the `Hybrid` API, but it does seem to be redundant. I will fix these issues.

src/main/java/net/imagej/ops/image/ImageNamespace.java
```diff
@@ -185,6 +185,25 @@ public Object equation(final Object... args) {
...
+	// -- fill --
+
+	/** Executes the "fill" operation on the given arguments. */
+	@OpMethod(op = net.imagej.ops.Ops.Image.Fill.class)
+	public Object fill(final Object... args) {
+		return ops().run(Ops.Image.Fill.NAME, args);
+	}
+
+	/** Executes the "fill" operation on the given arguments. */
+	@OpMethod(op = net.imagej.ops.image.fill.DefaultFill.class)
```
dietzc: `op = Fill.class` not on `DefaultFill`, right?

src/main/java/net/imagej/ops/image/ImageNamespace.java
```diff
@@ -185,6 +185,25 @@ public Object equation(final Object... args) {
...
+	public <T extends Type<T>> Iterable<T> fill(final IterableInterval<T> out,
+		final Type<T> in)
```
dietzc: Please use `T` instead of `Type<T>` here.

src/main/java/net/imagej/ops/image/ImageNamespace.java
```diff
@@ -185,6 +185,25 @@ public Object equation(final Object... args) {
 		return result;
 	}
 
+	// -- fill --
+
+	/** Executes the "fill" operation on the given arguments. */
+	@OpMethod(op = net.imagej.ops.Ops.Image.Fill.class)
+	public Object fill(final Object... args) {
+		return ops().run(Ops.Image.Fill.NAME, args);
```
dietzc: Please call via `Fill.class`

src/main/java/net/imagej/ops/map/MapII.java
```diff
@@ -0,0 +1,37 @@
...
+import org.scijava.plugin.Plugin;
+
+/**
+ * A paralleled version of {@link MapNullaryComputer} over an
+ * {@link IterableInterval}.
+ * 
+ * @author Leon Yang
+ * @param <O> element type of outputs
+ */
+@Plugin(type = Ops.Map.class, priority = Priority.LOW_PRIORITY + 1)
```
dietzc: `MapNullaryII`?

src/main/java/net/imagej/ops/map/MapIterable.java
```diff
@@ -0,0 +1,23 @@
...
+import org.scijava.plugin.Plugin;
+
+/**
+ * A {@link MapNullaryComputer} over an {@link Iterable}.
+ * 
+ * @author Leon Yang
+ * @param <O> element type of outputs
+ */
+@Plugin(type = Ops.Map.class, priority = Priority.LOW_PRIORITY)
+public class MapIterable<O> extends AbstractMapNullaryComputer<O, Iterable<O>> {
```
dietzc: `MapNullaryIterable`

dietzc:
looks great. I added some minor comments. Afterwards I think we can merge!

LeonYang5114:
@dietzc I have updated the commits to reflect some of your comments. For the remaining suggestions:

1. `@OpMethod(type = net.imagej.ops.image.fill.Default.class)` should not be changed to `Ops.Image.Fill` since the `ImageNamespaceTest` (and other namespace tests as well) tests if all the existing ops with `@Plugin(type = Op.SomeOp.class` have their corresponding `OpMethod` implementations. 
2. I am not sure how to make `DefaultFill` depends on `T` instead of `T extends Type<T>`, since filling requires modification in place, which depends on `Type.set(Type)` method. If we only have `T`, do we want to use `CreateNativeType` and cast to `T`?
3. Do we have a preference on `Ops.XXX.class` over `Ops.XXX.NAME`? On #348 I did not change the `.NAME` usages.
Thanks for the comments!

dietzc:
> @OpMethod(type = net.imagej.ops.image.fill.Default.class) should not be changed to Ops.Image.Fill since the ImageNamespaceTest (and other namespace tests as well) tests if all the existing ops with @Plugin(type = Op.SomeOp.class have their corresponding OpMethod implementations. 

But here https://github.com/imagej/imagej-ops/pull/348/files you are working on exactly this?!

> I am not sure how to make DefaultFill depends on T instead of T extends Type<T>, since filling requires modification in place, which depends on Type.set(Type) method. If we only have T, do we want to use CreateNativeType and cast to T?

Just in the signature use `... fill(T type ,...)` instead of `... fill (Type<T> type, ...)`. You want it to be `T`.

dietzc:
> Do we have a preference on Ops.XXX.class over Ops.XXX.NAME? On #348 I did not change the .NAME usages.

personally, I have... but actually it doesn't matter. I prefer to use `.class` which is safer from my point of view.

LeonYang5114:
>Just in the signature use ... fill(T type ,...) instead of ... fill (Type<T> type, ...). You want it to be T.

Thanks for catching that.

>But here https://github.com/imagej/imagej-ops/pull/348/files you are working on exactly this?!

Actually I was changing the implementation of those op methods in that PR, instead of the `type` of them. Or should I change those `type`s? Then we probably also have to change the `NamespaceTest`s.

dietzc:
> Actually I was changing the implementation of those op methods in that PR, instead of the type of them. 

Now I get it. We list all `Op`s which could potentially be selected in the `OpMethod`. Yes. You are right. Sorry.

LeonYang5114:
@dietzc can you check if the PR looks ok?

src/main/java/net/imagej/ops/map/MapNullaryII.java
```diff
@@ -0,0 +1,37 @@
...
+ * @param <O> element type of outputs
+ */
+@Plugin(type = Ops.Map.class, priority = Priority.LOW_PRIORITY + 1)
+public class MapNullaryII<O> extends
+	AbstractMapNullaryComputer<O, IterableInterval<O>>
+{
+
+	@Override
+	public void compute0(IterableInterval<O> output) {
+		ops().run(ChunkerOp.class, new CursorBasedChunk() {
```
dietzc: can't the `ChunkerOp` be initialized beforehands?

dietzc:
besides the minor detail above, it's great. thanks. I will merge.

LeonYang5114:
I think I would make a topic branch for initializing the `ChunkerOp`s, since all the parallel ops' `ChunckerOp`s are not initialized right now.

Issue #353, opened 2016-02-13
Matching Problem?


dietzc:
Hi @ctrueden,

I created a small test which shows a potential problem with the matching. The test is selfexplaining, however if you have any questions I'm happy to answer them.

https://github.com/imagej/imagej-ops/tree/matching_problem

Thanks,

Christian

LeonYang5114:
This problem is already solved after the latest commits. But to answer the question, here are the reasons:

1. `Dimension` is not `Interval`, so it can only match `DefaultCreateImg`, which is fine.
2. matcher tries to match `CreateImgFromInterval` in the second case, which tries to a `DefaultCreatingImg` using the input (should be an `Interval` if it is not matched by class object, or null o/w). Since the class information is lost, the initialization fails.
3. If a null check is there (e.g. `input == null ? Interval.class : input`), the matching would be fine.

By the way, it is me that wrote the `initialize` code and forgot the do the null check :-P

dietzc:
But `Interval extends Dimension`, so the matching should work, right?

LeonYang5114:
Yes, it should work, as long as `CreateImgFromInterval` does a null check in the initialize method.

dietzc:
Hi @LeonYang5114,  have a look at the example code which I provided in mentioned branch. It doesn't work with `Interval.class` but works fine for `Dimension.class`, so the `null` check can't be the reason or do I miss something obvious here? I try to find some time to debug this next week!

Thanks for your efforts!

Christian

LeonYang5114:
I pushed a1d51d9 to your branch, which fixed the problem.

The problem is that the two lines of codes actually match with the different ops. If the `CreateImgFromInterval` op is removed, there will be no problem. Anyway, the problem has been fixed in the master branch.

dietzc:
perfect. thanks

PR #361, opened 2016-02-20
test IntervalView to RAI

bnorthan:
I added tests for copying an interval view of a larger image to another image.

### DIFF
--- a/src/main/java/net/imagej/ops/copy/CopyRAI.java
+++ b/src/main/java/net/imagej/ops/copy/CopyRAI.java
@@ -38,6 +38,7 @@ import net.imagej.ops.special.computer.UnaryComputerOp;
 import net.imagej.ops.special.function.UnaryFunctionOp;
 import net.imagej.ops.special.hybrid.AbstractUnaryHybridCF;
 import net.imglib2.RandomAccessibleInterval;
+import net.imglib2.type.NativeType;
 import net.imglib2.type.Type;
 import net.imglib2.util.Intervals;
 import net.imglib2.util.Util;
@@ -70,13 +71,17 @@ public class CopyRAI<T> extends
 
 	@Override
 	public void initialize() {
-		final Class<?> outTypeClass =
-			out() == null ? Type.class : Util.getTypeFromInterval(out()).getClass();
-		final T inType = Util.getTypeFromInterval(in());
-		final UnaryComputerOp<T, ?> typeComputer =
-			Computers.unary(ops(), Ops.Copy.Type.class, outTypeClass, inType);
+		final Object outType = out() == null ? Type.class : Util
+			.getTypeFromInterval(out());
+
+		final Object inType = in() == null ? NativeType.class : Util
+			.getTypeFromInterval(in());
+
+		final UnaryComputerOp<?, ?> typeComputer = Computers.unary(ops(),
+			Ops.Copy.Type.class, outType, inType);
 		mapComputer = RAIs.computer(ops(), Ops.Map.class, in(), typeComputer);
 		createFunc = RAIs.function(ops(), Ops.Create.Img.class, in(), inType);
+
 	}
 
 	@Override
--- a/src/test/java/net/imagej/ops/copy/CopyRAITest.java
+++ b/src/test/java/net/imagej/ops/copy/CopyRAITest.java
@@ -31,16 +31,25 @@
 package net.imagej.ops.copy;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
 
 import java.util.Random;
 
 import net.imagej.ops.AbstractOpTest;
+import net.imagej.ops.Ops;
+import net.imagej.ops.special.hybrid.Hybrids;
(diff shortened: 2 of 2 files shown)


bnorthan:
I am getting an error when trying to match the copy op.  Looks like ```CopyRAI.initialize``` is using in() to determine the type.

``` java
UnaryHybridCF<IntervalView<UnsignedByteType>, RandomAccessibleInterval<UnsignedByteType>> copy =
			(UnaryHybridCF) Hybrids.unaryCF(ops, Ops.Copy.RAI.class,
				RandomAccessibleInterval.class, IntervalView.class);

```

dietzc:
this is similar to https://github.com/imagej/imagej-ops/issues/353. You can try to match it against a super-class of `IntervalView`.

bnorthan:
@dietzc It looks like it matches with CopyRAI, but then fails in the initialize because there is no ```in()``` yet.  I noticed that CopyRAI checks to see if ```out()``` is null, but not ```in()```.  

```
final T inType = Util.getTypeFromInterval(in());
```

LeonYang5114:
You are right. The only thing needed to fix this is to add null check for the `in()`. 

But this could apply to many other ops, since they might be match with no  `in()` or `out()` in some future use cases, and they want to match some helper ops. Then we need to do the null check every time we use `in()` or `out()`, or even better, make the matcher smarter so that the matcher can handler this automatically.

bnorthan:
@LeonYang5114, I don't mind fixing small things like this as I encounter them... however it seems that the method to deal with ```null``` inputs is still being determined in #352.   So I don't want to change anything until there are examples of how the checks should work.

LeonYang5114:
I would suggest you to fix that if that's what hindering your work. I just don't think it looks very elegant but that's what we have been doing in some places (e.g. [HistogramCreate](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/image/histogram/HistogramCreate.java#L63), [MapNeighborhood](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/map/neighborhood/MapNeighborhood.java#L73), [DefaultMedian](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/stats/DefaultMedian.java#L62), [DefaultPercentile](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/stats/DefaultPercentile.java#L66)), and I think we would need to write more this kind of null check later.

bnorthan:
I've got it working, but the code is longer then I'd like 18de342e23a37b992d552086cc191ceabbadb205.

I first tried like this... but the weird thing was that it did not work for the case where in() was not null. 

``` java
@Override
	public void initialize() {
		final Class<?> outTypeClass =
			out() == null ? Type.class : Util.getTypeFromInterval(out()).getClass();
		
		final Class<?> inTypeClass = 
				in()==null ? NativeType.class :Util.getTypeFromInterval(in()).getClass();
		
		final UnaryComputerOp<?, ?> typeComputer =
			Computers.unary(ops(), Ops.Copy.Type.class, outTypeClass, inTypeClass);
		mapComputer = RAIs.computer(ops(), Ops.Map.class, in(), typeComputer);
		createFunc = RAIs.function(ops(), Ops.Create.Img.class, in(), inTypeClass);
	}
```

dietzc:
`Util.getTypeFromInterval(in()).getClass()` why do you ask for the class here and not directly pass the type to the `CopyOp` (as you have it?). I would do `Util.getTypeFromInterval(in()).createVariable()` and pass this if possible. just go for `Object`.

bnorthan:
@dietzc, like this??  Or did I misunderstand you?
```
final Class<?> outTypeClass =
			out() == null ? Type.class : Util.getTypeFromInterval(out()).getClass();
		
		final Object inType = 
        in()==null ? NativeType.class :Util.getTypeFromInterval(in());

			final UnaryComputerOp<?, ?> typeComputer =
			Computers.unary(ops(), Ops.Copy.Type.class, outTypeClass, inType);
			mapComputer = RAIs.computer(ops(), Ops.Map.class, in(), typeComputer);
			createFunc = RAIs.function(ops(), Ops.Create.Img.class, in(), inType);
```

LeonYang5114:
It looks good to me.

src/main/java/net/imagej/ops/copy/CopyRAI.java
```diff
@@ -70,13 +71,17 @@
 
 	@Override
 	public void initialize() {
-		final Class<?> outTypeClass =
-			out() == null ? Type.class : Util.getTypeFromInterval(out()).getClass();
-		final T inType = Util.getTypeFromInterval(in());
-		final UnaryComputerOp<T, ?> typeComputer =
-			Computers.unary(ops(), Ops.Copy.Type.class, outTypeClass, inType);
+		final Class<?> outTypeClass = out() == null ? Type.class : Util
```
dietzc: here we could also type on `Object`?

ctrueden:
I split out the `CopyRAITest` auto-formatting to its own commit, and merged it! Thanks!

PR #362, opened 2016-02-24
Create native test

bnorthan:
I added tests for creating native types.  At first the tests failed.  In ```CreateNativeTypeFromClass```, ```createWorker``` tried to match a ```Ops.Create.Object``` with an output of ```NativeType.class```, but ```NativeType``` is an interface so it failed.  

At least that was my understanding of what happened.  I tweaked ```CreateNativeTypeFromClass``` and now the test passes.  Let me know if the change I made makes sense, or if there is a better solution.


ctrueden:
Thanks @bnorthan. I totally forgot about that detail of interfaces vs. classes—totally my fault. The new unit test is much appreciated.

I just twiddled the order of the commits, so that both of the commits compile with passing tests. I do the same as you: write a unit test which fails, then fix it to work—but then I always reverse the order to avoid having a commit with a failing test on it. The narrative of the history becomes: 1) fix this bug; and then 2) add a test to ensure the bug never comes back.

bnorthan:
@ctrueden thanks, I just reversed the order of commit in #361 due to the same issue.

PR #364, opened 2016-02-25
Create extended rai

LeonYang5114:
Wrapper of a `link RandomAccessibleInterval` and a `OutOfBoundsFactory` that keeps the interval information. All the method except for accessing out of bounds data are handled by the source RAI. Notice that this class is very similar to [`ExtendedRandomAccessibleInterval`](https://github.com/imglib/imglib2/blob/master/src/main/java/net/imglib2/view/ExtendedRandomAccessibleInterval.java), which does not have the interval information.

Creating an instance of this class could be considered a shorthand for: `Views.interval(Views.extend(source, outOfBoundsFactory), source);`

This branch is experimental. Any comment would be appreciated.

### DIFF
--- a/src/main/java/net/imagej/ops/ExtendedRAI.java
+++ b/src/main/java/net/imagej/ops/ExtendedRAI.java
@@ -1,0 +1,166 @@
+/*
+ * #%L
+ * ImageJ software for multidimensional image processing and analysis.
+ * %%
+ * Copyright (C) 2014 - 2016 Board of Regents of the University of
+ * Wisconsin-Madison, University of Konstanz and Brian Northan.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * 
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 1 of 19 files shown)


bnorthan:
@LeonYang5114 feel free to also refactor [```ConvolveNaive```](refactor https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/filter/convolve/ConvolveNaive.java) as part of this.  I will just take your version.

dietzc:
isn't this a duplicate of the efforts in: https://github.com/imagej/imagej-ops/tree/views?

LeonYang5114:
I did not notice that we have that branch. But as I read some of the code, I think that's probably not duplicate efforts. What I am doing here is to make `ExtendedRandomAccessibleInterval` easier to use. The new class `ExtendedRAI` is exactly the same as `ExtendedRandomAccessibleInterval` except that it keeps the `Interval` information, so that we can typed on `ExtendedRAI` in an op to make `RAI` extension really _required_. I did not see similar logic in the branch you mentioned. Hope that makes sense.

dietzc:
Are you sure that we want to introduce `ExtenededRAI` on our side just for convenience reasons? We can have an `Op` in the aforementoined `views` branch which actually does `Views.interval(Views.extend(source, outOfBoundsFactory), source)`. Having a class `ExtendedRAI` doesn't really make sense, as also the optimizations within the `Views` logic would fail. In short: I'm against adding `ExtendedRAI` but certainly we should have an `Op` which encapsulates `Views.interval(Views.extend(source, outOfBoundsFactory), source)` in https://github.com/imagej/imagej-ops/tree/views. Can you make this happen @tibuch?

ctrueden:
@dietzc @tpietzsch What if the `ExtendedRandomAccessibleInterval` was still an `Interval`? That would solve our issue. Is there some harm in doing that?

dietzc:
`ExtendedRandomAccessibleInterval` can't be an `Interval` as it is defined in infinity. however, I opened a PR https://github.com/imglib/imglib2/pull/123 which I'm sure will trigger some discussions. Anyway, I vote _strongly_ against having any `ExtendedRAI` class or any other data-structure within ops which actually should live in `imglib2`. We may not diverge from `imglib2`s principles, rather we should just use them.

I will close this PR. If my PR at `imglib2`will not be accepted we will find another solution.

PR #366, opened 2016-02-26
Fix map compatible

LeonYang5114:
When RAI and II are involved, they do not necessarily have the same dimensions. Instead, as long as the II is contained by the RAI, mapping could proceed.


ctrueden:
Great!

PR #368, opened 2016-02-29
Allow to ignore certain ops from being cached.

dietzc:
Sometimes a developer may want to exclude certain operations from being cached in the `CachedOpEnvironment`, e.g. a common case is that one may not cache `Create.class` operations. This PR allows to specify which classes to ignore.


ctrueden:
Looks good!

PR #369, opened 2016-02-29
UnaryHybridCI: ComputerOps which can be used for inplace computations

dietzc:
see title. @ctrueden would you mind quickly reviewing this. Maybe I've overlooked something.


ctrueden:
Do you mind if I push some more commits on top of this?

dietzc:
of course not.

ctrueden:
OK, good enough for now. We don't have `BinaryHybridCI` or `BinaryHybridCI1` yet, but I think it's OK for the time being.

PR #371, opened 2016-03-02
Fix double min max

LeonYang5114:
`Double.MIN_VALUE`, which equals to the minimum absolute double value, is used as initial values for finding the maximum values. The correct way is to use `Double.NEGATIVE_INFINITY`. The usage of `Double.MAX_VALUE` is changed to `Double.POSITIVE_INFINITY` for consistency.

Notice that the [`DefaultCoarsenessFeature`](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/features/tamura2d/DefaultCoarsenessFeature.java#L115) has a kind of strange behavior, since an `ArrayImg` with data `{0, -1, 0, 0}` has 0 coarseness.


ctrueden:
Thanks!

Issue #372, opened 2016-03-03
Can't run Otsu anymore


dietzc:
Try to execute: `IterableInterval<BitType> otsu = ops.threshold().otsu(in);`. This doesn't work anymore. Fails with: 

```
java.lang.IllegalArgumentException: No matching 'net.imagej.ops.Ops$Threshold$Apply/net.imagej.ops.special.computer.BinaryComputerOp' op

Request:
-	net.imagej.ops.Ops$Threshold$Apply/net.imagej.ops.special.computer.BinaryComputerOp(
		null,
		ArrayImg,
		UnsignedShortType)

Candidates:
1. 	(Iterable out) =
	net.imagej.ops.threshold.apply.ApplyConstantThreshold(
		Iterable out,
		Iterable in1,
		RealType in2)
	Argument types do not match: 0: cannot coerce null -> java.lang.Iterable<net.imglib2.type.logic.BitType>
2. 	(BitType out) =
	net.imagej.ops.threshold.apply.ApplyThresholdComparable(
		BitType out,
		Comparable in1,
		Object in2)
	Argument types do not match: 0: cannot coerce null -> class net.imglib2.type.logic.BitType
3. 	(BitType out) =
	net.imagej.ops.threshold.apply.ApplyThresholdComparator(
		BitType out,
		Object in1,
		Object in2,
		Comparator comparator)
	Not enough arguments: 3 < 4

	at net.imagej.ops.DefaultOpMatchingService.findModule(DefaultOpMatchingService.java:110)
	at net.imagej.ops.OpEnvironment.op(OpEnvironment.java:219)
	at net.imagej.ops.special.SpecialOp.op(SpecialOp.java:271)
	at net.imagej.ops.special.computer.Computers.binary(Computers.java:267)
	at net.imagej.ops.threshold.AbstractApplyThresholdImg.initialize(AbstractApplyThresholdImg.java:60)
	at net.imagej.ops.threshold.ApplyThresholdMethod$Otsu.initialize(ApplyThresholdMethod.java:330)
	at net.imagej.ops.DefaultOpMatchingService.findModule(DefaultOpMatchingService.java:103)
	at net.imagej.ops.OpEnvironment.module(OpEnvironment.java:251)
	at net.imagej.ops.OpEnvironment.run(OpEnvironment.java:155)
	at net.imagej.ops.threshold.ThresholdNamespace.otsu(ThresholdNamespace.java:1048)
[...]

PR #373, opened 2016-03-03
Thresholding Fixes

dietzc:
This PR fixes https://github.com/imagej/imagej-ops/issues/372.

However, this is just a workaround, as there seem to be deeper problems with the `OpMatching` which I couldn't figure out. However, as a next step I will try to create a test which lets the `OpMatcher` fail and makes the problem more accessible. Basically, now I only changed the `Function.unary(..)` calls to `Computers.unary(...)` calls. I think the `canCast` in https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/DefaultOpMatchingService.java#L267 is correct for `Computers` but not for `Functions`.  But I'm not sure yet. I have to investigate further. However, the most important thing is now, that `otsu` etc can be executed again.

Best,

Christian

### DIFF
--- a/src/main/java/net/imagej/ops/threshold/AbstractApplyThresholdImg.java
+++ b/src/main/java/net/imagej/ops/threshold/AbstractApplyThresholdImg.java
@@ -31,6 +31,7 @@
 package net.imagej.ops.threshold;
 
 import net.imagej.ops.Ops;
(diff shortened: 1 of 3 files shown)


hinerm:
:+1: @dietzc  :+1:

PR #375, opened 2016-03-04
Filter refactor2

bnorthan:
This is an update of the filter re-factor based on feedback given in #336 

This pull request refactors the filter/convolution/deconvolution ops to work towards the following general goals:

1. Use RandomAccessibleInterval instead of Img as op inputs.
2. Favor composition over inheritance where possible.
3. Derive from nullary/unary/binary functions/computers/hybrids.
4. Perform op matching in initialize.
5. Implement required math functions as ops
6. Make member variable/parameters private


bnorthan:
@dietzc, @ctrueden, @LeonYang5114, I went through the comments and dealt with all the "low hanging fruit" (the easy stuff).  There are still a few more subtle issues I have to think about, and will probably be asking questions about in the next couple weeks.

dietzc:
I'll try to review the code again over the weekend if you like?

bnorthan:
@dietzc if you have time that would be great.

src/main/java/net/imagej/ops/filter/fft/FFTMethodsOpF.java
```diff
@@ -0,0 +1,157 @@
...
+
+	private UnaryComputerOp<RandomAccessibleInterval<T>, RandomAccessibleInterval<C>> fftMethodsOp;
+
+	@Override
+	@SuppressWarnings({ "unchecked", "rawtypes" })
+	public void initialize() {
+		super.initialize();
+
+		// if no type was passed in the default is ComplexFloatType
+		if (fftType == null) {
```
dietzc: shouldn't this be done in the `CreateOp`?

src/main/java/net/imagej/ops/filter/fft/FFTMethodsUtility.java
```diff
@@ -0,0 +1,52 @@
...
+import net.imglib2.Dimensions;
+import net.imglib2.FinalDimensions;
+import net.imglib2.algorithm.fft2.FFTMethods;
+
+/**
+ * Utility class that interacts with FFTMethods
+ * 
+ * @author bnorthan
+ */
+public class FFTMethodsUtility {
```
dietzc: javadoc + final method arguments (I'm very picky here ;-))

src/main/java/net/imagej/ops/filter/fftSize/ComputeFFTSize.java
```diff
@@ -30,43 +30,65 @@
...
- * Op to calculate JTransform FFT sizes.
+ * Op that calculates FFT sizes.
  * 
  * @author Brian Northan
  */
 @Plugin(type = Ops.Filter.FFTSize.class)
-public class ComputeFFTSize extends AbstractFFTSize {
+public class ComputeFFTSize extends AbstractOp implements Ops.Filter.FFTSize {
+
+	@Parameter
```
dietzc: move down, such that `BOTH` are on top.

src/main/java/net/imagej/ops/filter/ifft/IFFTComputerOp.java
```diff
@@ -71,4 +77,35 @@ public void compute1(final RandomAccessibleInterval<C> input,
 			.unpaddingIntervalCentered(input, output), 0, true, service);
 	}
 
+	/**
+	 * Make sure that the input size conforms to a supported FFT size.
+	 */
+	@Override
+	public boolean conforms() {
+
```
dietzc: what if `in()` is null?

src/main/java/net/imagej/ops/filter/pad/PadInput.java
```diff
@@ -0,0 +1,56 @@
...
+ * @param <O>
+ */
+@Plugin(type = Ops.Filter.PadInput.class, name = Ops.Filter.PadInput.NAME,
+	priority = Priority.HIGH_PRIORITY)
+public class PadInput<T extends RealType<T>, I extends RandomAccessibleInterval<T>, O extends RandomAccessibleInterval<T>>
+	extends AbstractUnaryFunctionOp<I, O> implements Ops.Filter.PadInput
+{
+
+	@Parameter
+	Dimensions paddedDimensions;
```
dietzc: private

src/main/java/net/imagej/ops/filter/pad/PadInput.java
```diff
@@ -0,0 +1,56 @@
...
+	@Override
+	@SuppressWarnings("unchecked")
+	public O compute1(final I input) {
+
+		if (obf == null) {
+			obf = new OutOfBoundsConstantValueFactory<T, RandomAccessibleInterval<T>>(
+				Util.getTypeFromInterval(input).createVariable());
+		}
+
+		Interval inputInterval = ops().filter().paddingIntervalCentered(input,
```
dietzc: `initialize`

src/main/java/net/imagej/ops/filter/pad/PadInputFFTMethods.java
```diff
@@ -0,0 +1,60 @@
...
+
+		Dimensions paddedFFTMethodsInputDimensions = FFTMethodsUtility
+			.getPaddedInputDimensionsRealToComplex(fast, paddedDimensions);
+
+		if (obf == null) {
+			obf = new OutOfBoundsConstantValueFactory<T, RandomAccessibleInterval<T>>(
+				Util.getTypeFromInterval(input).createVariable());
+		}
+
+		Interval inputInterval = ops().filter().paddingIntervalCentered(input,
```
dietzc: `initialize`

src/main/java/net/imagej/ops/filter/pad/PadShiftKernelFFTMethods.java
```diff
@@ -0,0 +1,61 @@
...
+	public O compute2(final I kernel, final Dimensions paddedDimensions) {
+
+		Dimensions paddedFFTMethodsInputDimensions = FFTMethodsUtility
+			.getPaddedInputDimensionsRealToComplex(fast, paddedDimensions);
+
+		// compute where to place the final Interval for the kernel so that the
+		// coordinate in the center
+		// of the kernel is shifted to position (0,0).
+
+		final Interval kernelConvolutionInterval = ops().filter()
```
dietzc: `initialize`

src/main/java/net/imagej/ops/filter/pad/PadShiftKernelFFTMethods.java
```diff
@@ -0,0 +1,61 @@
...
+			.paddingIntervalCentered(kernel, paddedFFTMethodsInputDimensions);
+
+		final Interval kernelConvolutionIntervalOrigin = ops().filter()
```
dietzc: `initialize`

src/main/java/net/imagej/ops/math/divide/DivideHandleZero.java
```diff
@@ -0,0 +1,64 @@
...
+			if (out() == null) return true;
+			return Intervals.equalDimensions(in1(), out()) && in1().iterationOrder()
+				.equals(out().iterationOrder());*/
+	}
+
+	@Override
+	public void compute2(final IterableInterval<T> input1,
+		final IterableInterval<T> input2, final IterableInterval<T> output)
+	{
+		final Cursor<T> cursorI1 = input1.cursor();
```
dietzc: looks like a `BinaryMap`

src/main/java/net/imagej/ops/math/multiply/ComplexConjugateMultiply.java
```diff
@@ -0,0 +1,59 @@
...
+			if (out() == null) return true;
+			return Intervals.equalDimensions(in1(), out()) && in1().iterationOrder()
+				.equals(out().iterationOrder());*/
+	}
+
+	@Override
+	public void compute2(final IterableInterval<T> input1,
+		final IterableInterval<T> input2, final IterableInterval<T> output)
+	{
+		final Cursor<T> cursorI1 = input1.cursor();
```
dietzc: `BinaryMap`?

bnorthan:
@dietzc thanks for feedback again.  I fixed the issues you mentioned.  I also did a few other tweaks.  

@kephale I changed names of filter related classes on this branch.  The low level FFT and IFFT are now "FFTMethodsOpC" and "IFFTMethodsOpC".  These two classes now use the ThreadService.  So hopefully the problem that you saw yesterday is fixed.  If you get a chance, it would be great if you could try out this branch and let me know if everything works for you. 

I also changed the default setting for the OutOfBounds strategy for convolution to "Mirror" (it was "zero").  This might change the output of your algorithm(s).  Let me know if this is a problem.   You should be able to change back to "zero" by passing in the ```OutofBoundsFactory``` to the op.

dietzc:
I will do another pass if you like?

bnorthan:
@dietzc sure, that would be great.

kephale:
@bnorthan Awesome, thanks! 

I strongly vote for zero-padding as the default. That is the default behavior in both Matlab and Numpy.

ctrueden:
@dietzc This branch is getting moldy and has conflicts now... any chance you can look things over soon?

@bnorthan I see a "fixup" commit above—should that be squashed elsewhere?

dietzc:
I will take a look on Monday.

bnorthan:
@ctrueden should I merge and squash the entire thing back to one commit??  Does that make it easier for everyone (including myself) ??

ctrueden:
@bnorthan One monster commit is better than intermediate commits which don't compile with passing tests. But multiple granular commits that are well separated are the best. That said, I understand that it costs a lot of extra time to convert a monster commit into well-separated ones. So I'll leave it to your judgment. I personally strive to at least "peel off" the easy low-hanging fruit, where there are very small commits in obviously distinct places that do things other than what is described in the "big" commit. But it's your work—so do what you need to do.

ctrueden:
OK, I assigned this to @dietzc. But @bnorthan if you want to make changes before Monday, please do so. And if you want @dietzc to hold off on the review, please assign the issue to yourself instead. Thanks guys!

bnorthan:
Hi @dietz, @ctrueden  this branch should now be compatible with master again.  Feel free to review.

src/main/java/net/imagej/ops/deconvolve/RichardsonLucyC.java
```diff
@@ -0,0 +1,231 @@
...
+		}
+
+		// create image for the estimate, this image is defined over the entire
+		// convolution interval
+		final Img<O> estimate = create.compute1(getImgConvolutionInterval());
+
+		// create image for the reblurred
+		final Img<O> reblurred = create.compute1(getImgConvolutionInterval());
+
+		setRAIExtendedEstimate(estimate);
```
dietzc: I still have a bad feeling that we here set something in the parent class which we will use in the `performIterations` of the same child class. You should take care about thread-safety then (`getIndependentInstance`) ... but somehow I think we should redesign this.

src/main/java/net/imagej/ops/deconvolve/RichardsonLucyNonCirculantC.java
```diff
@@ -0,0 +1,382 @@
...
+	priority = Priority.LOW_PRIORITY)
+public class RichardsonLucyNonCirculantC<I extends RealType<I>, O extends RealType<O>, K extends RealType<K>, C extends ComplexType<C>>
+	extends AbstractIterativeFFTFilterC<I, O, K, C> implements
+	Ops.Deconvolve.RichardsonLucy
+{
+
+	@Parameter(required = false)
+	private StatusService status;
+
+	@Parameter(required = false)
```
dietzc: I think there shouldn't be any dependency on the `UIService` in `Ops` at all.

src/main/java/net/imagej/ops/deconvolve/RichardsonLucyNonCirculantC.java
```diff
@@ -0,0 +1,382 @@
...
+			normCursor.fwd();
+			normCursor.get().setReal(1.0);
+		}
+
+		final Img<O> tempImg = create.compute1(fd);
+
+		// 3. correlate psf with the output of step 2.
+		// TODO: Discuss how to initialize this and whether FFT filters should be
+		// unary...
+		ops().run(CorrelateFFTC.class, tempImg, normalization, kernel,
```
dietzc: we certainly should initialize this somehow. calling `ops().run` is really expensive.

src/main/java/net/imagej/ops/deconvolve/RichardsonLucyTVUpdate.java
```diff
@@ -387,54 +382,29 @@ public void run() {
 									Dymb = (bijk - bjm) / hy;
 									Dzmc = (cijk - ckm) / hz;
 
-									outRandom.setPosition(new int[] { i, j, k });
-									outRandom.get().setReal(Dxma + Dymb + Dzmc);
+									// outRandom.setPosition(new int[] { i, j, k });
```
dietzc: remove commented code.

src/main/java/net/imagej/ops/filter/AbstractIterativeFFTFilterC.java
```diff
@@ -0,0 +1,137 @@
...
+	private int maxIterations;
+
+	/**
+	 * The interval to process TODO: this is probably redundant - remove
+	 */
+	@Parameter
+	private Interval imgConvolutionInterval;
+
+	/**
+	 * TODO: make this an op?? A boolean which indicates whether to perform
```
dietzc: it's an `Op` now, right? :+1:

src/main/java/net/imagej/ops/filter/AbstractIterativeFFTFilterC.java
```diff
@@ -0,0 +1,137 @@
...
+ * @param <I>
+ * @param <O>
+ * @param <K>
+ * @param <C>
+ */
+public abstract class AbstractIterativeFFTFilterC<I extends RealType<I>, O extends RealType<O>, K extends RealType<K>, C extends ComplexType<C>>
+	extends
+	AbstractFFTFilterC<RandomAccessibleInterval<I>, RandomAccessibleInterval<O>, RandomAccessibleInterval<K>, RandomAccessibleInterval<C>>
+{
+	// TODO remove required=false? Can we always expect a Status Service?
```
dietzc: If we ask for a `StatusService` it should exist.

src/main/java/net/imagej/ops/filter/convolve/ConvolveFFTF.java
```diff
@@ -51,20 +55,37 @@
...
+public class ConvolveFFTF<I extends RealType<I> & NativeType<I>, O extends RealType<O> & NativeType<O>, K extends RealType<K> & NativeType<K>, C extends ComplexType<C> & NativeType<C>>
+	extends AbstractFFTFilterF<I, O, K, C> implements Ops.Filter.Convolve
 {
 
+	@Override
+	public void initialize() {
+
+		// TODO: try and figure out if there is a better place to set the default
+		// OBF
+		if (this.getOBFInput() == null) {
```
dietzc: shouldn't this be done in the super class?

dietzc:
I added some comments. Overall it looks really good. I only found many minor issues. Thanks @bnorthan for your work! Looking forward integrating it into KNIME!

src/main/java/net/imagej/ops/deconvolve/RichardsonLucyNonCirculantC.java
ctrueden: Agreed! It is a big bag of side effects.

src/main/java/net/imagej/ops/filter/AbstractIterativeFFTFilterC.java
bnorthan: @dietzc, @ctrueden 

currently I get "IllegalArgument Required service is missing" when running a unit test for an op that uses the StatusService.  RichardsonLucy really needs status service as it is nice to (apr.) what iteration we are at. 

I can fix this 2 ways

1.  Use ```required=false``` (which I am currently doing)
2.  Or @LeonYang5114 suggested using....

``` java
@Override
	public Context createContext() {
	    return new Context(OpService.class, OpMatchingService.class,
	        CacheService.class, StatusService.class);
	} 
```

Do you prefer the second option??  Let me know it doesn't matter to me.

ctrueden: I like `required = false` more. The your op will work when someone uses it with a `Context` that has no `StatusService`. It is not the case that every context will have one—it really depends on how it was constructed. The unit test is working as intended here.

bnorthan:
Thanks @dietzc.  I should be able to address your issues by the end of the week.  More to come...

src/main/java/net/imagej/ops/deconvolve/RichardsonLucyC.java
bnorthan: addressed by af24a7025f5db191d2593a3516107025368cd5c1

that variable no longer belongs to the base class.  There is no longer a ```RichardsonLucyNonCirculantC```, instead there is only ```RichardsonLucyC```, and different implementations of RL can be achieved by composition.

src/main/java/net/imagej/ops/deconvolve/RichardsonLucyNonCirculantC.java
bnorthan: That was just left over from debugging, it should be gone now.

src/main/java/net/imagej/ops/deconvolve/RichardsonLucyNonCirculantC.java
bnorthan: That piece of code got shifted around a bit, but the correlate is now initialized in [NonCirculantNormalizationFactor](https://github.com/imagej/imagej-ops/blob/filter_refactor2/src/main/java/net/imagej/ops/deconvolve/NonCirculantNormalizationFactor.java)

bnorthan:
@dietzc, @ctrueden I addressed most of Christian's comments.  

I anticipate more iterations on the design as this code is tested on real images, and as we try to implement different filter and deconvolution algorithms.  However I think the current state is good enough to merge, then it will be easier to continue developing with smaller sets of commits.  

Let me know if you see more that should done before merging.

bnorthan:
@ctrueden one thing I just thought of.  What is the easiest way to loop and verify that every commit compiles with passing tests?  I was looking at bisect but couldn't figure out if that's what I need.

hinerm:
@bnorthan see [this script](https://github.com/scijava/scijava-scripts/blob/1386a7b0bc9e832d45f925202e1382717cf4a706/check-branch.sh)

bnorthan:
@hinerm thanks!  That's what I was looking for.

ctrueden:
Unfortunately, since Ops takes multiple minutes to compile now due to javac algorithmic complexity, that script can take a long time. Hopefully the core Java team will get the javac compile times down in the future, so we can continue to maintain the quality of our code history.

bnorthan:
@ctrueden, @dietzc all commits should have passing tests now.

ctrueden:
> I think the current state is good enough to merge, then it will be easier to continue developing with smaller sets of commits.

Sounds good! Thank you!!

PR #377, opened 2016-03-09
Refactor Hyperslice

dietzc:
* Rename `Hyperslice` to `Slices`
* Make `Crop` a `BinaryComputerOp`
* Rename `Slicewise` to `Slice`


src/main/java/net/imagej/ops/image/crop/CropRAI.java
```diff
@@ -79,4 +77,4 @@
 		IntervalView<T> res = Views.offsetInterval(input, interval);
 		return oneSizedDims ? Views.dropSingletonDimensions(res) : res;
 	}
-}
+}
```
LeonYang5114: Newline here

PR #380, opened 2016-03-18
Add integral cursor implementation

stelfrich:
The `IntegralCursor` returns specific corner values of `RectangleNeighborhood`s that can be used for computations on integral images. 

In the 2D example the values at the following positions are returned by `IntegralCursor` according to their [binary Gray code](https://en.wikipedia.org/wiki/Gray_code) for optimized speed:
* (neighMin, neighMin)
* (neighMax-1, neighMin)
* (neighMax-1, neighMax-1)
* (neighMin, neighMax-1)

See also https://github.com/imglib/imglib2-algorithm/pull/17 for previous discussions on the PR.

**Requires a new imglib2-algorithm release because of https://github.com/imglib/imglib2-algorithm/pull/19.**

### DIFF
--- a/src/main/java/net/imagej/ops/image/integral/IntegralCursor.java
+++ b/src/main/java/net/imagej/ops/image/integral/IntegralCursor.java
@@ -1,0 +1,215 @@
+/*
+ * #%L
+ * ImageJ software for multidimensional image processing and analysis.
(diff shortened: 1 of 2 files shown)


dietzc:
great!

PR #384, opened 2016-03-31
Improve type flexibility of inplace ops

LeonYang5114:
I made a rebase and clean up the history. Unfortunately, changing the framework produces a bunch of compile errors, so that I think we need to fix them in one commit, which lead to the huge commit 79b64b0 By making this rebase, the branch differs a lot from the original one, so I just create a new one. I wish I know some git tricks to add a merge history at the tail of this branch to "connect" it with the original one.

I also created the binary CI1 CFI1 CI CFI interfaces/abstract classes. We can keep or remove them. By the way, the SpecialOpMatchingTest is getting huge because of those new interfaces.

This branch also depends on the granular matching branch, which is WIP. Therefore, this branch should not be merge before that one.

@ctrueden please check it out once you have time

### DIFF
--- a/src/main/java/net/imagej/ops/OpEnvironment.java
+++ b/src/main/java/net/imagej/ops/OpEnvironment.java
@@ -444,12 +444,11 @@ public interface OpEnvironment extends Contextual {
 
 	/** Executes the "join" operation on the given arguments. */
 	@OpMethod(op = net.imagej.ops.join.DefaultJoin2Inplaces.class)
-	default <A> A join(final A arg, final UnaryInplaceOp<A> first,
-		final UnaryInplaceOp<A> second)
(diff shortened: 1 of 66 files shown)


ctrueden:
@LeonYang5114 I see that you rebased this. Thanks! Is this good to review & merge?

ctrueden:
Ha, this stuff has gotten complicated enough that once again, `javac` on the CLI dislikes the code, even if Eclipse's compiler is OK with it. Investigating now...

ctrueden:
@axtimwalde Hopefully these changes address your concerns regarding the typing of inplace ops. We will release a new ImageJ Ops later today. Please give it a spin when you have a chance.

PR #385, opened 2016-04-06
Granular matcher

LeonYang5114:
This branch adds functionality for the matcher to match ops given multiple `OpRef`s. This allows more flexible matching (for example, see #384)


ctrueden:
Thank you!

PR #386, opened 2016-04-08
Remove varargs

LeonYang5114:
varargs OpMethods (with single parameter `Object... args`) in all namespaces are removed, so that we only have type-safe signatures. See #296


ctrueden:
:+1:

PR #387, opened 2016-04-08
Integral image-backed local thresholding ops

stelfrich:
- Add `IntegralMean` and `IntegralVariance` that compute mean and variance for `RectangleNeighborhood`, respectively
- Fix an issue with `DefaultIntegralImg`/`SquaredIntegralImg` and the handling of boundary conditions
- Add versions of local thresholding ops that only require mean and standard deviation ops:
  - `LocalMean`
  - `LocalNiblack`
  - `LocalPhansalkar`
  - `LocalSauvola`
- Integral image-backed ops are not available via `ThresholdNamespace` for now


ctrueden:
This is looking pretty super awesome.

However, I have a concern about the introduction of additional `FIXME` comments. According to [ImageJ code style](http://imagej.net/Coding_style#Javadoc_and_comments):

> For code (or lack thereof) that is considered wrong or broken, and in need of repair, we add a "FIXME" comment with the relevant developer's initials, to serve as a reminder to address it as soon as time allows.

So we really, really do not want these in the code whenever possible. If it is just a matter of needing some cleanup later, please use `TODO` (with your initials) instead. However, if the code is truly "wrong" or "broken" then we should not be merging it in the first place...

ctrueden:
@dietzc Any additional thoughts or concerns about this branch? I did not do a thorough review of every commit, but did look things over quickly...

dietzc:
I will take a look on Monday.

src/main/java/net/imagej/ops/filter/AbstractCenterAwareNeighborhoodBasedFilter.java
```diff
@@ -49,7 +49,7 @@
 {
 
 	@Parameter
-	private Shape shape;
+	protected Shape shape;
```
dietzc: we could add a getter here instead of `protected` field

src/main/java/net/imagej/ops/image/integral/AbstractIntegralImg.java
```diff
@@ -90,29 +90,47 @@ public void compute1(RandomAccessibleInterval<I> input,
...
+
 		RandomAccessibleInterval<? extends RealType<?>> generalizedInput = input; // FIXME
-		
+
 		// Create integral image
 		for (int i=0; i < input.numDimensions(); ++i) {
 			// Slicewise integral addition in one direction
 			slicewiseOps[i].compute1(generalizedInput, output);
-			generalizedInput = output;
+			generalizedInput = ops().copy().rai(output);
```
dietzc: can we initialize the `copy` first? Also: Do we really need to copy? can't we just do it during the first iteration?

src/main/java/net/imagej/ops/image/integral/AbstractIntegralImg.java
```diff
@@ -90,29 +90,47 @@ public void compute1(RandomAccessibleInterval<I> input,
...
+
+		// Define the Interval on the infinite random accessibles
+		final FinalInterval interval = new FinalInterval(min, max);
+
+		RandomAccessibleInterval<RealType<?>> output = null;
+
 		// Create integral image
 		if (Util.getTypeFromInterval(input) instanceof IntegerType) {
-			return (RandomAccessibleInterval) ops().create().img(input, new LongType());
+			output = (RandomAccessibleInterval) ops().create().img(interval,
```
dietzc: initialize `create` in `initialize`

src/main/java/net/imagej/ops/stats/IntegralMean.java
```diff
@@ -0,0 +1,149 @@
...
+		final DoubleType sum = new DoubleType();
+		sum.setZero();
+
+		// Convert from input to return type
+		final Converter<I, DoubleType> conv = new RealDoubleConverter<>();
+		
+		while ( cursor.hasNext() )
+		{
+			final I value = cursor.next().get(0).copy();
+			final DoubleType valueAsDoubleType = new DoubleType();
```
dietzc: can we reuse the `DoubleType` instead creating one inside the loop?

src/main/java/net/imagej/ops/stats/IntegralMean.java
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	/**
+	 * Correct the provided {@link Interval} for the increased neighborhood size
+	 * that is necessary to work with {@link IntegralCursor}.
+	 * 
+	 * @param neighborhood
+	 * @return size-corrected {@link Interval}.
+	 */
+	public static Interval correctNeighborhoodInterval(Interval neighborhood) {
```
dietzc: There is a method `Intervals.expand` or `Intervals.extend` or something like this in `imglib2`.

src/main/java/net/imagej/ops/stats/IntegralSum.java
```diff
@@ -0,0 +1,102 @@
...
+		while ( cursor.hasNext() )
+		{
+			// Obtain the cursor position encoded as corner vector
+			int cornerInteger = cursor.getCornerRepresentation();
+			final I value = cursor.next().copy();
+			final DoubleType valueAsDoubleType = new DoubleType();
+			conv.convert(value, valueAsDoubleType);
+			
+			// Determine if the value has to be added (factor==1) or subtracted (factor==-1)
+			DoubleType factor = new DoubleType(Math.pow(-1.0d, dimensions - norm(cornerInteger)));
```
dietzc: reuse `DoubleType`

src/main/java/net/imagej/ops/stats/IntegralVariance.java
```diff
@@ -0,0 +1,118 @@
...
+		
+		// Compute \sum (-1)^{dim - ||cornerVector||_{1}} * I(x^{cornerVector})
+		final DoubleType sum2 = new DoubleType();
+		sum2.setZero();
+		
+		while ( cursorS1.hasNext() )
+		{
+			final Composite<I> compositeValue = cursorS1.next();
+			final I value1 = compositeValue.get(0).copy();
+			final DoubleType value1AsDoubleType = new DoubleType();
```
dietzc: reuse `DoubleType`

src/main/java/net/imagej/ops/stats/StatsNamespace.java
```diff
@@ -90,6 +93,44 @@
...
+	@OpMethod(op = net.imagej.ops.stats.IntegralSum.class)
+	public DoubleType integralSum(final RectangleNeighborhood in) {
+		final DoubleType result = (DoubleType) ops().run(
+			net.imagej.ops.stats.IntegralSum.class, in);
+		return result;
+	}
+
+	@SuppressWarnings("rawtypes")
+	@OpMethod(op = net.imagej.ops.stats.IntegralSum.class)
+	public DoubleType integralSum(final DoubleType out,
```
dietzc: why does `Integral` sum need its own method signature? Shouldn't the framework take care about choosing the fastest possible `Op` for a certain use-case?

src/main/java/net/imagej/ops/threshold/apply/LocalThresholdIntegral.java
```diff
@@ -0,0 +1,186 @@
...
+			listOfIntegralImages);
+		RandomAccessibleInterval<? extends Composite<RealType>> compositeRAI = Views
+			.collapse(stacked);
+		RandomAccessibleInterval<? extends Composite<RealType>> extendedCompositeRAI =
+			removeLeadingZeros(compositeRAI);
+
+		NeighborhoodsIterableInterval<? extends Composite<RealType>> neighborhoods =
+			shape.neighborhoodsSafe(extendedCompositeRAI);
+
+		if (map == null) {
```
dietzc: why can't we do this in the initialize? limitation of framework? which information would you need in the framework? @ctrueden maybe we need the smarter type matching or rather the _fake_ `RAI` etc.

src/main/java/net/imagej/ops/threshold/apply/LocalThresholdIntegral.java
```diff
@@ -0,0 +1,186 @@
...
+	public RectangleShape getShape() {
+		return shape;
+	}
+
+	/**
+	 * @return the Computer to map to all neighborhoods of input to output.
+	 */
+	protected abstract CenterAwareIntegralComputerOp<I, BitType> unaryComputer();
+
+	protected abstract int[] requiredIntegralImages();
```
dietzc: java doc

src/main/java/net/imagej/ops/threshold/localPhansalkar/LocalPhansalkarThresholdIntegral.java
```diff
@@ -0,0 +1,150 @@
...
+
+			final DoubleType stdDev = new DoubleType(Math.sqrt(variance.get()));
+
+			final DoubleType threshold = new DoubleType(mean.getRealDouble() * (1.0d +
+				p * Math.exp(-q * mean.getRealDouble()) + k * ((stdDev.getRealDouble() /
+					r) - 1.0)));
+
+			// Set value
+			final Converter<I, DoubleType> conv = new RealDoubleConverter<>();
+			final DoubleType centerPixelAsDoubleType = new DoubleType();
```
dietzc: can we reuse another `DoubleType` from above?

dietzc:
I added some comments. @stelfrich did you actually compare the runtime of `IntegralImg`s in `Ops` with the implementation in `imglib2`? What are the results of this benchmark? Could be interesting to make this faster.

src/main/java/net/imagej/ops/image/integral/AbstractIntegralImg.java
stelfrich: > can't we just do it during the first iteration?

Could you elaborate on that? I don't get the idea :worried:

src/main/java/net/imagej/ops/threshold/apply/LocalThresholdIntegral.java
stelfrich: That's actually a good question. Why would the concrete types of `out` and `in` be required (i.e. `T` in `RAI<T>`) for a `MapOp`?

src/main/java/net/imagej/ops/stats/StatsNamespace.java
stelfrich: You are right. However, the requirements for `integral*` methods are deviating from a typical `sum` method so much that I don't think the framework would have any chance to bring those together.

That is mostly due to the fact that e.g. `integralSum` cannot be fed a typical `II` to compute the sum but rather requires computation of an integral image beforehand. We could make `integralSum` behave in a more `sum`-like way by internally computing the integral image from the input `RAI`. But then you (as developer) would not be able to reuse the integral image, which is the whole point of integral image-backed implementations..

dietzc:
@stelfrich you think this is good to merge?

stelfrich:
@dietzc I have done some benchmarks for the local mean thresholding on 1000px * 1000px randomly generated images with rectangular windows of different spans:
runtimes
From that point of view we should definitely get this merged :smile:

However, I also compared runtimes of the local mean thresholding for the two implementations of integral images (the re-implementation as op with the SliceOp and the imglib2 original). The imglib2 version outperforms the re-implementation:
```
Re-implementation   614.363 ± 56.873 ms/op
Imglib2             411.373 ± 42.024 ms/op
```

Further benchmarks and profiling revealed that the re-implementation an order of magnitude slower than the original one due to the copying in [AbstractIntegralImg:L119](https://github.com/imagej/imagej-ops/pull/387/files#diff-b6f5aded641f401104304fef4a646b76R119):
```
DefaultIntegralImg   112.643 ± 7.317 ms/op
WrappedIntegralImg   13.856 ± 2.251 ms/op
```

dietzc:
I still don't understand why we need the copying. Can you explain again? Or should we meet in person and discuss it?

dietzc:
from personal discussions with @stelfrich: 

TODO:

* [ ] Fix leading zero to avoid copying

This is a hack to get around the leading zero problem. Has to be cleaned up and beautified:
```
		final long[] min = Intervals.minAsLongArray(img);
			final long[] max = Intervals.maxAsLongArray(img);

			for (int i = 0; i < max.length; i++) {
				min[i]--;
			}

			RealType tmp = (RealType) Util.getTypeFromInterval(img).copy();
			tmp.setZero();

			IntervalView<RealType> offsetInterval = Views.offsetInterval(Views.extendValue(img, tmp), min, max);
			ExtendedRandomAccessibleInterval<RealType, IntervalView<RealType>> extendBorder = Views
					.extendBorder(offsetInterval);
			img = Views.offsetInterval(extendBorder, input);
```

* [ ] Initialize `Op`s whenever possible. See methods
* [ ] Clean-up code and to make it _even_ faster: Example `IntegralMean` `compute2(...)` is a bit over designed (avoid constructor calls, avoid method calls, avoid converters, ...).
* [ ] Benchmark again.

Thanks @stelfrich for your work.

stelfrich:
New benchmarks are in:

```
Benchmark                                    Mode  Cnt   Score   Error  Units
IntegralImgBenchmark.testDefaultIntegralImg  avgt    5  62.276 ± 8.137  ms/op
IntegralImgBenchmark.testWrappedIntegralImg  avgt    5  28.234 ± 1.945  ms/op
```

Currently, I don't see any obvious ways to further improve the speed of the integral image computations, @dietzc.

> Clean-up code

Has the preferred line length in the code style/formatter preferences recently changed? When I format with Eclipse I get lines with length > 100chars..

dietzc:
@stelfrich is the op-matching part of the benchmark? if you run it 50 times and the op-matching is also run 50-times, then I'm sure I know why we are slower ;-)

stelfrich:
@dietzc The benchmark uses matching on the highest level for both implementations. Under the hood, however, the ImgLib2 implementation does not include any matching. The re-implementation uses matching for the output creation and a `Slice` op in each dimension (with all that's going on under the hood with chunking etc).

However, I just realized a stupid mistake on my side with the benchmarks: running them in parallel interferes with the parallelization of Ops and skews the results. Here are the new results when running the benchmarks serially:

```
Benchmark                                    Mode  Cnt   Score   Error  Units
IntegralImgBenchmark.testDefaultIntegralImg  avgt   10  11.015 ± 2.212  ms/op
IntegralImgBenchmark.testWrappedIntegralImg  avgt   10   9.328 ± 0.417  ms/op
```

dietzc:
same VM or different VM instances? However, the 2seconds overhead are pretty likely to come from the matching?

stelfrich:
> same VM or different VM instances?

Same VM but different forks (i.e. the JIT profile has been reset before runs).

> However, the 2seconds overhead are pretty likely to come from the matching?

That's 2 milliseconds and this might actually be parallelization overhead. Results for a 5000px*5000px:
```
Benchmark                                    Mode  Cnt    Score    Error  Units
IntegralImgBenchmark.testDefaultIntegralImg  avgt   20  226.224 ± 13.046  ms/op
IntegralImgBenchmark.testWrappedIntegralImg  avgt   20  285.777 ±  5.302  ms/op
```

dietzc:
sitting next to me and forcing me to merge doesn't count as cool kid, but it's close! Thanks for the work ;-)

PR #389, opened 2016-04-22
Multipass matching

LeonYang5114:
DefaultOpMatchingService now do a multi-pass matching instead of a
single pass.

First it search for Ops whose arguments match exactly with the
references.

If fails, it search for Ops whose arguments match by casting. Ops with
fewest total levels of casting will be returned.

If fails, it then search for Ops whose arguments match by lossless
conversion. This functionality is not yet implemented, because it needs
support from the conversion service to tell if a conversion is lossy.

If still no match is found, it does the regular search.

### DIFF
--- a/src/main/java/net/imagej/ops/DefaultOpMatchingService.java
+++ b/src/main/java/net/imagej/ops/DefaultOpMatchingService.java
@@ -36,6 +36,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.function.Predicate;
 
 import net.imagej.ops.OpCandidate.StatusCode;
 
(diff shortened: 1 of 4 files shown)


src/main/java/net/imagej/ops/DefaultOpMatchingService.java
```diff
@@ -246,6 +244,220 @@ public Module assignInputs(final Module module, final Object... args) {
...
+		for (final OpCandidate<?> candidate : candidates) {
+			final ModuleInfo info = candidate.cInfo();
+			final double p = info.getPriority();
+			if (p != priority && !matches.isEmpty()) {
+				// NB: Lower priority was reached; stop looking for any more matches.
+				break;
+			}
+			priority = p;
+
+			if (typesMatch(candidate, candidate.getArgs()) != -1) continue;
```
ctrueden: This method and `losslessMatches` and `perfectMatches` are all the same, except for this type matching test per candidate. Can we use some Java 8 functional goodness here to consolidate them all into a single method?

src/test/java/net/imagej/ops/OpMatchingServiceTest.java
```diff
@@ -38,7 +38,21 @@
 
 import java.util.List;
 
+import net.imglib2.Dimensions;
```
ctrueden: We need to be careful about depending on ImgLib2 classes at this layer. Otherwise, it will make splitting out the `scijava-ops` project much harder. (I know: I see `DoubleType` was already imported below... but we should fix that too.)

ctrueden:
If you can find a way to test the multi-pass matching without leaning on ImgLib2, I'd appreciate it.

The functional consolidation would be nice, but we can merge without that, and fix it later. Either way.

LeonYang5114:
@ctrueden the branch is now updated

ctrueden:
I still feel like there is a lot of new code here, which could be reorganized and/or streamlined. However, in the interest of pragmatism, I'm merging now. I hope to reduce the size of the `DefaultOpMatchingService`... later this year, maybe.

Issue #363, opened 2016-02-24
Make op method extensible - TODO


LeonYang5114:
See #348 

As mentioned, the filter namespace and logic namespace were not not refactored in that PR. The reason is that, the filter namespace is still under refactoring (#336), and the logic namespace could crash if such refactoring is applied.

The logic namespace has conflicts between some `PrimitiveLogic` ops and some generic logic ops (which are typed on `T`, not `T extends Type<T>`). For example:
```
ops().logic().greaterThan(10, 11);         // should it returns BoolType or boolean?
```
To solve this problem, more work might need to be done. Since I am not currently working on this project, I file this issue just as a reminder.

ctrueden:
I think we should resolve this now, at least with respect to the logic namespace. To have a clean separation of SciJava Ops, we need logic ops that work on primitives. So the example above should use `boolean` (or `Boolean`), not `BoolType`. With the right `Converter` implementations, it should be OK.

PR #392, opened 2016-04-27
Improve logic

LeonYang5114:
Add logical operations: and, or, xor, not, and their map versions. Notice that the existing framework using "Condition" is not very convenient to use and might be removed in the future.


LeonYang5114:
The branch is updated such that the logic ops working with the `Conditiona` framework are replaced with ops in the `BooleanTypeLogic` utility class.

`LogicNamespace` now uses op interfaces to do matching inside the `OpMethod`s, which solves the issue mentioned here #363

@ctrueden

LeonYang5114:
The `Ops.Logic.Bool` op is not migrated, since I think it is too simple and some converter should achieve its functionality.

ctrueden:
Thank you!

Issue #391, opened 2016-04-25
Wrap iterative flood fill as an op


ctrueden:
@axtimwalde [writes](https://gitter.im/imagej/imagej-ops/archives/2016/04/25):

> Hey ops developers, @hanslovskyp has just finished an iterative (vs recursive) generic (int terms of type and neighborhood) n-dimensional flood-fill implementation https://github.com/imglib/imglib2-algorithm/tree/master/src/main/java/net/imglib2/algorithm/fill
> we did not feel comfortable including this as an op upfront but if anybody here feels like wrapping it up we would be excited

ctrueden:
@LeonYang5114 Please take a look. I'm guessing it will be easy to integrate this into the morphology namespace.

LeonYang5114:
I think we want to copy and paste the source code and make minimum modification to make it fit into an op, right?

Also there is a few questions: why the value of the target RAI matters? so this operation is inherently in-place? Should we create a `fill` package and namespace?

axtimwalde:
@LeonYang5114 

> why the value of the target RAI matters?

because you need to know where you've been already.

> so this operation is inherently in-place? 

Nope.

dietzc:
Actually, we don't want to copy & Paste the source. rather wrap the original implementation. If this is not possible, we have to understand why and then solve the problem.

tibuch:
Imglib2 implementation is the new DefaultFloodFill.

PR #394, opened 2016-04-29
Create FloodFill op

lnyng:
see #391 

This branch create a wrapper op for the imglib2-algorithm.fill.FloodFill class. FillNamespace is also created for this op.

This PR is for review purpose. I feel that the `Filter<Pair<T, U>, Pair<T, U>>` and the `Writer<U>` are not looking very nice. I hope that they could be some kind of ops.

I am not sure if we also want some default `Filter`s. And do we need to write unit tests? since the op is just a wrapper.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -204,6 +204,7 @@
 		<dependency>
 			<groupId>net.imglib2</groupId>
 			<artifactId>imglib2-algorithm</artifactId>
+			<version>0.5.1-SNAPSHOT</version>
 		</dependency>
 		<dependency>
 			<groupId>net.imglib2</groupId>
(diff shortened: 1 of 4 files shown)


ctrueden:
Just a quick note: can't merge this until `imglib2-algorithm` is pinned to a release version.

dietzc:
@tibuch any news on the integration of the `Imglib2-algorithm` `FloodFill`?

tibuch:
I replaced the DefaultFloodFill with the imglib2 implementation. 

DefaultFloodFill is part of the morphology namespace and I haven't changed that. If you think a FillNamespace is needed, I can add it.

pom.xml
```diff
@@ -204,6 +204,7 @@
 		<dependency>
 			<groupId>net.imglib2</groupId>
 			<artifactId>imglib2-algorithm</artifactId>
+			<version>0.5.1-SNAPSHOT</version>
```
dietzc: We have to wait until imglib2-algorithm was released or change this back if it has already been released.

dietzc:
I don't think we need a namespace here. The existing namespace should be fine. I'm closing this issue. Feel free to open it again if you disagree.

PR #395, opened 2016-04-29
Array img opt neighborhoods-Leon's branch

LeonYang5114:
This is a branch built on top of @Squareys and @ctrueden 's work. See #191 

The first 4 commits modify the existing interfaces of `MapNeighborhood`, so that the following commits could work. Basically it makes the `Shape` parameter generic, and add an optional `OOBF` to both default and center-aware implementations.

Some modifications are rebased into the middle commits, so that the `OpEnvironment` always compile.

My last four commits change the structure of @Squareys 's implementation a little bit, so that those `NativeType` versions are actually `MapNeighborhood` ops, instead just `MapOp`s. I also made some small fixes. See the commit message for more information.

This branch is not ready for merging, because the history is messy (although all the tests are passing now). @Squareys @ctrueden Please have a look and clean up the history as you want.


dietzc:
@Squareys can you have a look please?

One remark which I also discussed with @Squareys back then: We have to make sure that we don't introduce any new behaviour. I mean: we only want to merge this branch if the behaviour of `MapNeighborhood` is exactly the same as before, just faster. We don't want to introduce a new `OutOfBoundsStrategy` like _ignore if out of bounds_. I short: the optimization should work with any `OOBStrategy`.

Squareys:
@dietzc 
> We have to make sure that we don't introduce any new behaviour. I mean: we only want to merge this branch if the behaviour of MapNeighborhood is exactly the same as before, just faster.

The `MapNeighborhoodNativeTypeExtended` is a fast implementation for NativeType ArrayImg and is compatible with the `oobFactory` parameter, but will only use it when it needs to (the border neighborhoods.) This implementation can make use of super optimized Neighborhood implementations which do not require OOB (one follows).

The `MapNeighborhoodNativeType` is an implementation *without* out of bounds factories on NativeType ArrayImg. It assumes there is no out of bounds needed (therefore, the region iterated over is sufficiently smaller than the image itself.) This implementation is used by the other implementation to speed up the center interval which does not require out of bounds checks.

In short: This complies with existing behaviour.

I hope this clears some confusion.

@LeonYang5114 The changes look great, thank you for taking care of this! There are a few formatting commits mixed with actual changes in the WIP commits, but I guess that will be taken care of once the history is rewritten.

LeonYang5114:
Since I am not very familiar with the original commits, I made two big rebases to rewrite the history, one for implementation and one for test. @Squareys if you have time, feel free to split the commit and add more detailed commit message. If you think it is OK, then this branch is ready for merging.

PR #401, opened 2016-06-01
Mean filter types

bnorthan:
The mean filter op can support different input and output types.  This pull request changes the function in the namespace so that it also supports different input and output types.


ctrueden:
Nice catch, thanks @bnorthan!

PR #407, opened 2016-06-14
Refactor geom ops

tibuch:
I refactored all geometric ops from ...FunctionOp to ...HybridCF. 

I did this because I need Computers to create ComputerSets: https://github.com/imagej/imagej-ops/tree/feature-sets-table

@dietzc could you take a look?


dietzc:
to me this looks good. tests are still passing?

tibuch:
Tests are passing.

ctrueden:
Since all the ops return the concrete type `DoubleType`, using `HybridCF` here is awesome. Thanks!

Issue #404, opened 2016-06-09
CenterAwareComputerOps that can operate without OutOfBoundsFactory


stelfrich:
`CenterAwareComputerOps` require an `OutOfBounds` strategy to access pixels in neighborhoods that do not completely cover the image area (/volume, I will stick to the 2D case here). There are, however, filters that could compute exact values for those neighborhoods by adapting their computation scheme.

For example, `MeanFilterOp` could compute the mean value based on the available image pixels, effectively ignoring the out of bounds strategy. I don't know how those cases are handled in other libraries/frameworks/etc and if we should stick to the current behavior of `CenterAwareComputerOps` when no `OOBF` is provided by the user (i.e. assuming an `OOBF` that is used for the computations).

What do you think about that @dietzc, @bnorthan, anyone?

dietzc:
Two thoughts on that:

* Actually, we shouldn't force the user to provide a `OOB` ever, as in some cases, the `OOB` is not required at all (e.g. if the `RandomAccess` is big enough, so we don't need an `OOB`). However, this case should be handled and optimized by `ImgLib2`. (would be nice if we could test if it's true or if we make something wrong on ops side here):

*  The `exact calculation` behavior if no `OOB` is defined doesn't work for all algorithms. If we want to support this, then we have to put the `OOB` field explicitly into the implementation of the `algorithm` rather than in the `CenterAware...` or any other `AbstractNeighborhoodOp`. Then we could have the following for e.g. the `MeanFilter`: If no `OOB` is provided, we use the exact calculation (integral image based). If an `OOB` is provided we could use  (a) the original values of the `RandomAccess` if defined or (b) the values generated by the `OOB`. I think using the original values should be favoured (and as mentioned before this should be optimized by imglib2).

Does this make sense?

stelfrich:
> Does this make sense?

It does to me. We will just have to document the behavior such that the user is not confused with the computed results from exact calculations near the borders. Usually it's not a big deal, but depending on the size ratio of the neighborhood to the image you can have a pretty large region of influence of the `OOB`. 

The Image Processing Toolbox of MATLAB for example provides several `OOB`s and sets one by default based on what they assume to be the best default for the respective filter (as far as I can tell). This is what is currently implemented in Ops.

bnorthan:
[...]

PR #409, opened 2016-06-28
Integral image-based local thresholdings use provided OOBF

### LINKED ISSUE #403: Heuristics in ThresholdNamespace methods
With the current implementations in `ThresholdNamespace` the local thresholding ops that are backed by integral images are not actually used due to the higher priority of their "normal" counterparts (see [ThresholdNamespace.java#L431](https://github.com/imagej/imagej-ops/blob/62256de7e5f9f5a85bb51970735cb16fdfa98d60/src/main/java/net/imagej/ops/threshold/ThresholdNamespace.java#L431)).

In this particular case we cannot decide which implementation to use solely based on the parameters and the matching. What's the opinion on introducing some heuristics in those methods that direct the calls to specific implementations based on properties of the parameters (e.g. if `Shape.getSpan() >= 3` use the integral version).

I feel that this could make life a lot easier when working with the namespace methods from scripts as long as the outputs are guaranteed to be the same for both implementations (I have addressed that issue on the [`local-threshold-oobf-refactoring`](https://github.com/imagej/imagej-ops/tree/local-threshold-oobf-refactoring) branch).

stelfrich:
This PR changes the behavior of integral image-based local thresholding techniques such that the user provided OOBF is properly incorporated. This is achieved by increasing the size of the input image by the span of the local window/kernel and computing the integral image from this version instead of the smaller input. Hence, the provided OOBF will be used to construct the integral image. The output interval is set back to the one of the input image. 

The integral image-based implementations are now chosen based on the span as well as the image size using `Contingent`.

Closes #403 and is related to #404 (this PR removes the exact computation that is discussed in there).

### DIFF
--- a/src/main/java/net/imagej/ops/stats/IntegralMean.java
+++ b/src/main/java/net/imagej/ops/stats/IntegralMean.java
@@ -33,8 +33,7 @@ package net.imagej.ops.stats;
 import net.imagej.ops.Op;
 import net.imagej.ops.Ops;
 import net.imagej.ops.image.integral.IntegralCursor;
-import net.imagej.ops.special.computer.AbstractBinaryComputerOp;
-import net.imglib2.Interval;
+import net.imagej.ops.special.computer.AbstractUnaryComputerOp;
 import net.imglib2.algorithm.neighborhood.RectangleNeighborhood;
 import net.imglib2.converter.Converter;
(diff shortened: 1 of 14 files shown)


dietzc:
great. I think consistency with the other methods is really something we want here!

PR #412, opened 2016-07-08
HoG can handle images with 3 channels (e.g. RGB)

SimonSchmid:


src/main/java/net/imagej/ops/features/hog/HistogramOfOrientedGradients2D.java
```diff
@@ -124,33 +144,73 @@ public void convert(final T arg0, final FloatType arg1) {
 
 	@Override
 	public boolean conforms() {
-		return in().numDimensions() == 2;
+		return ((in().numDimensions() == 2) || ((in().numDimensions() == 3) && (in().dimension(2) == 3)));
```
dietzc: is this really a requirement that `in().dimension(2)==3`? actually it should work for arbitrary dims, right?

SimonSchmid: Yes you're right. I'll fix that.

src/main/java/net/imagej/ops/features/hog/HistogramOfOrientedGradients2D.java
```diff
@@ -69,8 +69,8 @@
  * pixel a histogram of gradient directions by summing up the magnitudes of the
  * neighbored (@param spanOfNeighborhood) pixels. The directions are divided in
  * (@param numOrientations) bins. The output is 3d: for each bin an own channel.
- * Input can be either a grayscale image or a color image with three channels
- * (e.g. RGB, LAB, ...).
+ * Input can be either a grayscale image or a color image with several color
```
dietzc: maybe more like "Input can be either a 2d or 3d image, where the third dimension is interpreted as color (see paper...)"?

dietzc:
great. thanks!

PR #415, opened 2016-07-10
Topology namespace and ops

rimadoma:
Adds a `Topology` namespace and first topological ops:
- `EulerCharacteristic26N` - calculates the Euler characteristic of 3D binary images
- `EulerCharacteristic26NFloating` - same as above, but voxels touching the edges of interval do not affect the result
- `EulerCorrection` - approximates the interval's contribution to the Euler characteristic of the whole image (assuming the object in the interval is a part of a larger object)

### DIFF
--- a/src/main/java/net/imagej/ops/topology/TopologyNamespace.java
+++ b/src/main/java/net/imagej/ops/topology/TopologyNamespace.java
@@ -1,0 +1,38 @@
+package net.imagej.ops.topology;
+
+import net.imagej.ops.AbstractNamespace;
+import net.imagej.ops.Namespace;
+import net.imagej.ops.OpMethod;
+import net.imagej.ops.topology.eulerCharacteristic.EulerCorrection;
+import net.imglib2.RandomAccessibleInterval;
+import net.imglib2.type.BooleanType;
+import org.scijava.plugin.Plugin;
+
+/**
+ * The topology namespace contains operations for calculating topology characteristics
+ *
+ * @author Richard Domander (Royal Veterinary College, London)
+ */
+@Plugin(type = Namespace.class)
+public class TopologyNamespace extends AbstractNamespace {
+
+    @Override
+    public String getName() {
+        return "topology";
+    }
+
+    @OpMethod(op = net.imagej.ops.topology.eulerCharacteristic.EulerCharacteristic26N.class)
+    public <B extends BooleanType<B>> Double eulerCharacteristic26N(final RandomAccessibleInterval<B> in) {
+        return (Double) ops().run(net.imagej.ops.Ops.Topology.EulerCharacteristic26N.class, in);
+    }
+
+    @OpMethod(op = net.imagej.ops.topology.eulerCharacteristic.EulerCharacteristic26NFloating.class)
+    public <B extends BooleanType<B>> Double eulerCharacteristic26NFloating(final RandomAccessibleInterval<B> in) {
+        return (Double) ops().run(net.imagej.ops.Ops.Topology.EulerCharacteristic26NFloating.class, in);
+    }
+
+    @OpMethod(op = EulerCorrection.class)
+    public <B extends BooleanType<B>> Double eulerCorrection(final RandomAccessibleInterval<B> in) {
+        return (Double) ops().run(net.imagej.ops.Ops.Topology.EulerCorrection.class, in);
+    }
+}
--- a/src/main/java/net/imagej/ops/topology/eulerCharacteristic/EulerCharacteristic26N.java
+++ b/src/main/java/net/imagej/ops/topology/eulerCharacteristic/EulerCharacteristic26N.java
@@ -1,0 +1,116 @@
+package net.imagej.ops.topology.eulerCharacteristic;
+
+import net.imagej.ops.Contingent;
+import net.imagej.ops.Ops;
+import net.imagej.ops.special.function.AbstractUnaryFunctionOp;
+import net.imglib2.RandomAccess;
+import net.imglib2.RandomAccessibleInterval;
+import net.imglib2.type.BooleanType;
+import org.scijava.plugin.Plugin;
+
+/**
+ * An Op which calculates the euler characteristic (χ) of the given 3D binary image.
+ * Here Euler characteristic is defined as χ = β_0 - β_1 + β_2, where β_i are so called Betti numbers.
+ * β_0 = number of separate particles
+ * β_1 = number of handles
+ * β_2 = number enclosed cavities
+ * <p>
+ * The Op calculates χ by using the triangulation algorithm described by Toriwaki & Yonekura (see below).
+ * There it's calculated X = ∑Δχ(V), where V is a 2x2x2 neighborhood around each point in the 3D space.
+ * We are using the 26-neighborhood version of the algorithm. The Δχ(V) values here are predetermined.
+ * <p>
+ * For the algorithm see
+ * Toriwaki J, Yonekura T (2002) Euler Number and Connectivity Indexes of a Three Dimensional Digital Picture.
+ * Forma 17: 183-209.
+ * <a href="http://www.scipress.org/journals/forma/abstract/1703/17030183.html">
+ * http://www.scipress.org/journals/forma/abstract/1703/17030183.html</a>
+ * <p>
+ * For the Betti number definition of Euler characteristic see
+ * Odgaard A, Gundersen HJG (1993) Quantification of connectivity in cancellous bone,
+ * with special emphasis on 3-D reconstructions.
+ * Bone 14: 173-182.
+ * <a href="http://dx.doi.org/10.1016/8756-3282(93)90245-6">doi:10.1016/8756-3282(93)90245-6</a>
+ *
+ * @author Richard Domander (Royal Veterinary College, London)
+ * @author David Legland  - original MatLab implementation
+ */
+@Plugin(type = Ops.Topology.EulerCharacteristic26N.class)
+public class EulerCharacteristic26N<B extends BooleanType<B>>
+        extends AbstractUnaryFunctionOp<RandomAccessibleInterval<B>, Double>
+        implements Ops.Topology.EulerCharacteristic26N, Contingent {
+    /** Δχ(v) for all configurations of a 2x2x2 voxel neighborhood */
+    private static final int[] EULER_LUT = {
+             0,  1,  1,  0,  1,  0, -2, -1,  1, -2,  0, -1,  0, -1, -1,  0,
+             1,  0, -2, -1, -2, -1, -1, -2, -6, -3, -3, -2, -3, -2,  0, -1,
+             1, -2,  0, -1, -6, -3, -3, -2, -2, -1, -1, -2, -3,  0, -2, -1,
+             0, -1, -1,  0, -3, -2,  0, -1, -3,  0, -2, -1,  0,  1,  1,  0,
+             1, -2, -6, -3,  0, -1, -3, -2, -2, -1, -3,  0, -1, -2, -2, -1,
+             0, -1, -3, -2, -1,  0,  0, -1, -3,  0,  0,  1, -2, -1,  1,  0,
+            -2, -1, -3,  0, -3,  0,  0,  1, -1,  4,  0,  3,  0,  3,  1,  2,
+            -1, -2, -2, -1, -2, -1,  1,  0,  0,  3,  1,  2,  1,  2,  2,  1,
+             1, -6, -2, -3, -2, -3, -1,  0,  0, -3, -1, -2, -1, -2, -2, -1,
+            -2, -3, -1,  0, -1,  0,  4,  3, -3,  0,  0,  1,  0,  1,  3,  2,
+             0, -3, -1, -2, -3,  0,  0,  1, -1,  0,  0, -1, -2,  1, -1,  0,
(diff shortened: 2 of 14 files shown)


src/main/java/net/imagej/ops/topology/EulerCharacteristic26N.java
```diff
@@ -0,0 +1,112 @@
...
+ * with special emphasis on 3-D reconstructions.
+ * Bone 14: 173-182.
+ * <a href="http://dx.doi.org/10.1016/8756-3282(93)90245-6">doi:10.1016/8756-3282(93)90245-6</a>
+ *
+ * @author Richard Domander
+ * @author David Legland (original MatLab implementation)
+ */
+@Plugin(type = Op.class, name = "eulerCharacteristic26N")
+public class EulerCharacteristic26N<B extends BooleanType>
+        extends AbstractUnaryFunctionOp<RandomAccessibleInterval<B>, Double> implements Contingent {
```
ctrueden: We should consider carefully what return type to use.

`Double` is immutable, and therefore unsuitable for use with `ComputerOp`s.

If you look at the stats ops such as stats.mean (e.g., `DefaultMean`), you'll see that they are hybrid ops which produce an `O extends RealType<O>`. Unfortunately, this actually causes a problem with generics. Really the best structure is to use two ops:

* `DefaultMeanFunction<I extends RealType<I>, O extends RealType<O>> extends UFViaUC<I, DoubleType, I, O>`
* `DefaultMeanComputer<I extends RealType<I>, O extends RealType<O>> implements AbstractUnaryComputerOp<I, O>`

That way, you end up with two method signatures:

* `<I extends RealType<I>> DoubleType mean(I in)`
* `<I extends RealType<I>, O extends RealType<O>> O mean(I in, O out)`

The `UFViaUC` is a "unary function via unary computer" convenience class where the function computation delegates to a computer. However, if you study the generics above carefully, they are still a bit funny. I may need to rethink this.

Anyway, I think we want to follow a similar pattern here. Of course, I am open to any ideas to make this pattern as painless as possible. (In particular, it sucks needing to create the function wrapper version of such ops every time.)

src/main/java/net/imagej/ops/topology/EulerCharacteristic26N.java
```diff
@@ -36,7 +36,7 @@
  * @author David Legland (original MatLab implementation)
  */
 @Plugin(type = Op.class, name = "eulerCharacteristic26N")
-public class EulerCharacteristic26N<B extends BooleanType>
+public class EulerCharacteristic26N<B extends BooleanType<B>>
```
ctrueden: It would be cleaner to squash this into the previous commit.

src/main/java/net/imagej/ops/topology/EulerCharacteristic26N.java
```diff
@@ -0,0 +1,112 @@
...
+ * For the Betti number definition of Euler characteristic see
+ * Odgaard A, Gundersen HJG (1993) Quantification of connectivity in cancellous bone,
+ * with special emphasis on 3-D reconstructions.
+ * Bone 14: 173-182.
+ * <a href="http://dx.doi.org/10.1016/8756-3282(93)90245-6">doi:10.1016/8756-3282(93)90245-6</a>
+ *
+ * @author Richard Domander
+ * @author David Legland (original MatLab implementation)
+ */
+@Plugin(type = Op.class, name = "eulerCharacteristic26N")
```
ctrueden: The goal is for every commit to compile with passing tests, as described in the [Reproducible builds](http://imagej.net/Architecture#Reproducible_builds) section of the web site.

So, best would be to squash commits that fix the namespace into this initial commit.

src/main/java/net/imagej/ops/topology/Octant.java
```diff
@@ -0,0 +1,84 @@
...
+import net.imglib2.type.BooleanType;
+import net.imglib2.view.Views;
+
+import java.util.Arrays;
+
+/**
+ * A convenience class for storing a 2x2x2 voxel neighborhood in an image
+ *
+ * @author Richard Domander
+ * @author Mark Hiner
```
ctrueden: Where'd this come from? Why is Mark an author? Just curious. :-)

src/test/java/net/imagej/ops/topology/EulerCharacteristic26NFloatingTest.java
```diff
@@ -0,0 +1,127 @@
...
+import net.imglib2.RandomAccess;
+import net.imglib2.img.Img;
+import net.imglib2.img.array.ArrayImgs;
+import net.imglib2.type.logic.BitType;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+/**
+ * Tests for the {@link EulerCharacteristic26NFloating EulerCharacteristic26NFloating} class
```
ctrueden: No need to replicate the class name twice here.

src/test/java/net/imagej/ops/topology/EulerCharacteristic26NFloatingTest.java
```diff
@@ -0,0 +1,127 @@
...
+    @Test(expected = IllegalArgumentException.class)
+    public void testConforms() throws AssertionError {
+        final Img<BitType> img = ArrayImgs.bits(3, 3);
+
+        ops.run(EulerCharacteristic26NFloating.class, img);
+    }
+
+    /**
+     * Test with a single voxel (=solid cube) that floats in the middle of a 3x3x3 space
+     * <p>
```
ctrueden: Probably want to close your paragraph with `</p>` at the end, although javadoc parser was historically lax about this.

src/test/java/net/imagej/ops/topology/EulerCharacteristic26NFloatingTest.java
```diff
@@ -0,0 +1,127 @@
...
+     * Here χ = β_0 - β_1 + β_2 = 1 - 0 + 0 = 1.
+     * The formula χ = vertices - edges + faces for surfaces of polyhedra doesn't apply because the cuboid is solid.
```
ctrueden: Always previous your javadoc in your IDE; as written, the sentence starting with "The formula" will _not_ begin on a new line.

src/test/java/net/imagej/ops/topology/OctantTest.java
```diff
@@ -0,0 +1,55 @@
...
+import net.imglib2.RandomAccess;
+import net.imglib2.img.Img;
+import net.imglib2.img.array.ArrayImgs;
+import net.imglib2.type.logic.BitType;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+/**
+ * Unit tests for the {@link Octant Octant} convenience class
```
ctrueden: Again, no need to duplicate link text twice.

src/test/java/net/imagej/ops/topology/OctantTest.java
```diff
@@ -0,0 +1,55 @@
...
+    @Test
+    public void testIsNeighborhoodEmpty() throws Exception {
+        final Img<BitType> img = ArrayImgs.bits(2, 2, 2);
+        Octant<BitType> octant = new Octant<>(img);
+
+        octant.setNeighborhood(1, 1, 1);
+
+        assertTrue("Neighborhood should be empty", octant.isNeighborhoodEmpty());
+
+        img.forEach(BitType::setOne);
```
ctrueden: Yay for Java 8!

src/main/java/net/imagej/ops/topology/EulerCharacteristic26N.java
```diff
@@ -89,7 +89,8 @@ public Double compute1(RandomAccessibleInterval<B> interval) {
      * @param z         Location of the neighborhood in the 3rd spatial dimension (z)
      * @return the index of the Δχ value for this configuration of voxels
      */
-    private int neighborhoodEulerIndex(final RandomAccess<B> access, final long x, final long y, final long z) {
+    public static <B extends BooleanType<B>> int neighborhoodEulerIndex(final RandomAccess<B> access, final long x,
+                                                                        final long y, final long z) {
```
ctrueden: Note that this does not conform to ImageJ code style. We should probably create an approximation of the code style rules for IntelliJ at some point.

src/test/java/net/imagej/ops/topology/EulerCharacteristic26NTest.java
```diff
@@ -22,6 +22,28 @@ public void testConforms() throws AssertionError {
         ops.run(EulerCharacteristic26N.class, img);
     }
 
+    @Test
+    public void testNeighborhoodEulerIndex() throws Exception {
+        final Img<BitType> bits = ArrayImgs.bits(2, 2, 2);
+        final RandomAccess<BitType> neighborhood = bits.randomAccess();
+
+        // Create and test all 256 configurations of a 2x2x2 neighborhood
+        for (int n = 0; n < 256; n++) {
```
ctrueden: This is so elegant!

src/main/java/net/imagej/ops/topology/eulerCharacteristic/EulerCharacteristic26N.java
```diff
@@ -1,4 +1,4 @@
-package net.imagej.ops.topology;
+package net.imagej.ops.topology.eulerCharacteristic;
```
ctrueden: Should probably squash this whole commit into the first one.

src/test/java/net/imagej/ops/topology/eulerCharacteristic/EulerCharacteristic26NFloatingTest.java
```diff
@@ -103,6 +99,7 @@ public void testHandleCube() throws Exception {
         assertEquals("Euler characteristic (χ) is incorrect", 0.0, result, 1e-12);
     }
 
+    //region -- Helper methods --
```
ctrueden: Is this "region" annotation an IntelliJ thing?

src/main/java/net/imagej/ops/topology/eulerCharacteristic/EulerCharacteristic26N.java
```diff
@@ -31,12 +31,13 @@
  * Bone 14: 173-182.
  * <a href="http://dx.doi.org/10.1016/8756-3282(93)90245-6">doi:10.1016/8756-3282(93)90245-6</a>
  *
- * @author Richard Domander
- * @author David Legland (original MatLab implementation)
+ * @author Richard Domander (Royal Veterinary College, London)
+ * @author David Legland  - original MatLab implementation
```
ctrueden: It somehow makes me really happy that you changed this to a dash, so that you wouldn't have two inconsistent meanings for parenthetical author notes (one meaning "affiliated with this place" and another meaning more generally "additional details of authorship"). 👍

src/main/java/net/imagej/ops/topology/eulerCharacteristic/EulerCharacteristic26NFloating.java
```diff
@@ -32,13 +33,13 @@
  * Bone 14: 173-182.
  * <a href="http://dx.doi.org/10.1016/8756-3282(93)90245-6">doi:10.1016/8756-3282(93)90245-6</a>
  *
- * @author Richard Domander
- * @author Michael Doube
+ * @author Richard Domander (Royal Veterinary College, London)
```
ctrueden: BTW, I really don't care if you include your affiliation or not. As you probably noticed, all the U. Konstanz team does, but the UW-Madison team does not. As long as it is consistent for each organization I am fine with whatever.

src/main/java/net/imagej/ops/topology/TopologyNamespace.java
```diff
@@ -3,6 +3,7 @@
 import net.imagej.ops.AbstractNamespace;
 import net.imagej.ops.Namespace;
 import net.imagej.ops.OpMethod;
+import net.imagej.ops.topology.eulerCharacteristic.EulerCorrection;
```
ctrueden: Question for you: are there Euler-related ops that are not "Euler characteristics"? Because if not, I would vote for `topology.euler` over `topology.eulerCharacteristic`.

Actually though, both of these packages violate the naming convention. From the [Naming guide](https://github.com/imagej/imagej-ops/wiki/Naming):
> Each kind of operation lives in an associated package matching its namespace and op name. For example, ops named math.add are declared in the net.imagej.ops.math.add package.

src/test/java/net/imagej/ops/topology/eulerCharacteristic/EulerCharacteristic26NFloatingTest.java
```diff
@@ -19,7 +19,7 @@
     public void testConforms() throws AssertionError {
         final Img<BitType> img = ArrayImgs.bits(3, 3);
 
-        ops.run(EulerCharacteristic26NFloating.class, img);
+        ops.topology().eulerCharacteristic26NFloating(img);
```
ctrueden: Actually, since the built-in method signatures do _not_ guarantee that your exact op is called, there is some question about whether to use these sigs, or to use `run`. Maybe worth a quick discussion. This is fine for the moment though since I think the test codebase is pretty inconsistent about it right now.

ctrueden:
Thanks so much! This is awesome!

PR #416, opened 2016-07-11
Added euler-char review suggestions

rimadoma:
* Make ops unaryHybridCF
* Fix Javadoc formatting


ctrueden:
Thanks!

ctrueden:
You could refactor the Euler hybrid ops to have a common base class, to slightly reduce boilerplate code. But it is not a big deal.

Issue #112, opened 2015-02-23
Create Tukey threshold method


ctrueden:
A ThresholdMethod that uses Tukey's rule to estimate a minimum/maximum range that excludes outliers would be useful.

http://www.edgarstat.com/tukeys_outliers_help.cfm

This would definitely work well on FLIM images and perhaps on other images.

Migrated-From: http://trac.imagej.net/ticket/1970

Issue #109, opened 2015-01-29
Separate OPS framework from built-in operations


ctrueden:
I know exactly how we can achieve a separation of concerns without burdening downstream consumers too much!

1. Create (again) `scijava-ops` with the core OPS framework.
    * No dependencies on image-specific things (e.g., `imagej-common` or `imglib2`) of course.
    * Remove the image-specific built-in op signatures from the `OpService` interface.
    * Put the code generation layer and arithmetic operations, expression parser, etc., which are not specific to images, all in the `scijava-ops` component.
2. Keep `imagej-ops` as the place for all the image processing goodies.
    * Provide a subinterface called `ImageOpService` which extends `OpService` and adds all the built-ins we want to have.
    * Narrow the return type of the `ImageJ` gateway's `op()` method (in the `net.imagej:imagej` component) to return an `ImageOpService` instead of just an `OpService`. So anyone using an ImageJ gateway will have access to these image processing routines.

In this way, we will have the powerful SciJava OPS framework available to the world, including an awesome, extensible expression parser and evaluator. And the ImageJ OPS project as the flagship image processing routines built on top of it!

ctrueden:
Just to clarify: we will not do this right away. Not until the core framework has stabilized and we are approach `1.0.0` release.

dietzc:
absolutely makes sense from my point of view. there are several other areas which could make use of the mechanism introduced in the ops framework.

PR #417, opened 2016-07-11
Add thresholds ops

rimadoma:
- Add an op that creates a binary mask based on thresholds
- Create an op that filters elements that fall within thresholds (this should go to the SciJava ops module).
  Elements and thresholds can be any type as long as they are comparable.

I don't know if the Thresholds name space is the best fit for these, but at least conceptually it seemed like the place to add these

### DIFF
--- a/src/main/java/net/imagej/ops/mask/maskRange/DefaultMaskRange.java
+++ b/src/main/java/net/imagej/ops/mask/maskRange/DefaultMaskRange.java
@@ -8,6 +8,7 @@ import java.util.stream.StreamSupport;
 import net.imagej.ops.Ops;
 import net.imagej.ops.special.function.AbstractBinaryFunctionOp;
 import net.imglib2.type.Type;
+import net.imglib2.util.Pair;
 
 import org.scijava.plugin.Plugin;
 
@@ -18,36 +19,18 @@ import org.scijava.plugin.Plugin;
  * @author Richard Domander (Royal Veterinary College, London)
  */
 @Plugin(type = Ops.Mask.MaskRange.class)
-public class DefaultMaskRange<T extends Type<T>, S extends Comparable<T>>
-	extends
-	AbstractBinaryFunctionOp<Iterable<T>, DefaultMaskRange.Range<S>, Iterable<T>>
+public class DefaultMaskRange<T extends Type<T>, U extends Comparable<T>>
+	extends AbstractBinaryFunctionOp<Iterable<T>, Pair<U, U>, Iterable<T>>
 	implements Ops.Mask.MaskRange
 {
 
 	@Override
-	public Iterable<T> compute2(Iterable<T> iterable, Range<S> range) {
+	public Iterable<T> compute2(Iterable<T> iterable, Pair<U, U> range) {
 		final Stream<T> inStream = StreamSupport.stream(iterable.spliterator(),
 			false);
-		final Stream<T> filtered = inStream.filter(e -> range.min.compareTo(
-			e) <= 0 && range.max.compareTo(e) >= 0);
+		final Stream<T> filtered = inStream.filter(e -> range.getA().compareTo(
+			e) <= 0 && range.getB().compareTo(e) >= 0);
 
 		return filtered.map(Type::copy).collect(Collectors.toList());
-	}
-
-	public static final class Range<S> {
-
-		/**
-		 * Minimum value for elements within the range
-		 */
-		public final S min;
-		/**
-		 * Maximum value for elements within the range
-		 */
-		public final S max;
-
-		public Range(S min, S max) {
-			this.min = min;
-			this.max = max;
-		}
 	}
 }
--- a/src/main/java/net/imagej/ops/mask/maskRange/DefaultMaskRange.java
+++ b/src/main/java/net/imagej/ops/mask/maskRange/DefaultMaskRange.java
@@ -1,0 +1,53 @@
+
+package net.imagej.ops.mask.maskRange;
+
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+import net.imagej.ops.Ops;
+import net.imagej.ops.special.function.AbstractBinaryFunctionOp;
+import net.imglib2.type.Type;
+
+import org.scijava.plugin.Plugin;
+
+/**
+ * Creates an Iterable from all the elements in the input that are in the given
+ * range of values
+ *
+ * @author Richard Domander (Royal Veterinary College, London)
+ */
+@Plugin(type = Ops.Mask.MaskRange.class)
+public class DefaultMaskRange<T extends Type<T>, S extends Comparable<T>>
+	extends
+	AbstractBinaryFunctionOp<Iterable<T>, DefaultMaskRange.Range<S>, Iterable<T>>
+	implements Ops.Mask.MaskRange
+{
+
+	@Override
+	public Iterable<T> compute2(Iterable<T> iterable, Range<S> range) {
+		final Stream<T> inStream = StreamSupport.stream(iterable.spliterator(),
+			false);
+		final Stream<T> filtered = inStream.filter(e -> range.min.compareTo(
+			e) <= 0 && range.max.compareTo(e) >= 0);
+
+		return filtered.map(Type::copy).collect(Collectors.toList());
+	}
+
+	public static final class Range<S> {
+
+		/**
+		 * Minimum value for elements within the range
+		 */
+		public final S min;
+		/**
+		 * Maximum value for elements within the range
+		 */
+		public final S max;
+
+		public Range(S min, S max) {
+			this.min = min;
+			this.max = max;
+		}
+	}
+}
--- a/src/main/templates/net/imagej/ops/Ops.list
+++ b/src/main/templates/net/imagej/ops/Ops.list
@@ -226,6 +226,11 @@ namespaces = ```
 		[name: "or",                             iface: "Or"],
 		[name: "xor",                            iface: "Xor"],
 	]],
+	[name: "mask", iface: "Mask", ops: [
+		[name: "maskHigh",                       iface: "MaskHigh"],
+		[name: "maskLow",                        iface: "MaskLow"],
+		[name: "maskRange",                      iface: "MaskRange"],
+	]],
 	[name: "math", iface: "Math", ops: [
 		[name: "abs",                            iface: "Abs"],
 		[name: "add",                            iface: "Add",                 aliases: ["sum"]],
--- a/src/main/java/net/imagej/ops/mask/MaskNamespace.java
+++ b/src/main/java/net/imagej/ops/mask/MaskNamespace.java
@@ -1,0 +1,63 @@
+
+package net.imagej.ops.mask;
+
+import net.imagej.ops.AbstractNamespace;
+import net.imagej.ops.Namespace;
+import net.imagej.ops.OpMethod;
+import net.imagej.ops.mask.maskHigh.DefaultMaskHigh;
+import net.imagej.ops.mask.maskLow.DefaultMaskLow;
+import net.imagej.ops.mask.maskRange.DefaultMaskRange;
+
+import org.scijava.plugin.Plugin;
+
+/**
+ * The mask namespace contains ops for creating iterables where certain values
+ * have been masked out
+ *
+ * @author Richard Domander (Royal Veterinary College, London)
+ */
+@Plugin(type = Namespace.class)
+public class MaskNamespace extends AbstractNamespace {
+
+	// -- Mask high --
+	@OpMethod(ops = DefaultMaskHigh.class)
+	public <T, S extends Comparable<T>> Iterable<T> maskHigh(Iterable<T> in,
+		S value)
+	{
+		@SuppressWarnings("unchecked")
+		Iterable<T> result = (Iterable<T>) ops().run(DefaultMaskHigh.class, in,
+			value);
+
+		return result;
+	}
+
+	// -- Mask low --
+	@OpMethod(ops = DefaultMaskLow.class)
+	public <T, S extends Comparable<T>> Iterable<T> maskLow(Iterable<T> in,
+		S value)
+	{
+		@SuppressWarnings("unchecked")
+		Iterable<T> result = (Iterable<T>) ops().run(DefaultMaskLow.class, in,
+			value);
+
+		return result;
+	}
+
+	// -- Mask range --
+	@OpMethod(ops = DefaultMaskRange.class)
+	public <T, S extends Comparable<T>> Iterable<T> maskRange(Iterable<T> in,
+		DefaultMaskRange.Range<S> range)
+	{
+		@SuppressWarnings("unchecked")
+		Iterable<T> result = (Iterable<T>) ops().run(DefaultMaskRange.class, in,
+			range);
+
+		return result;
+	}
+
+	// -- Namespace methods --
+	@Override
+	public String getName() {
+		return "mask";
+	}
+}
--- a/src/main/java/net/imagej/ops/mask/MaskNamespace.java
+++ b/src/main/java/net/imagej/ops/mask/MaskNamespace.java
@@ -7,6 +7,7 @@ import net.imagej.ops.OpMethod;
 import net.imagej.ops.mask.maskHigh.DefaultMaskHigh;
 import net.imagej.ops.mask.maskLow.DefaultMaskLow;
 import net.imagej.ops.mask.maskRange.DefaultMaskRange;
+import net.imglib2.util.Pair;
 
 import org.scijava.plugin.Plugin;
 
@@ -45,8 +46,8 @@ public class MaskNamespace extends AbstractNamespace {
 
 	// -- Mask range --
 	@OpMethod(ops = DefaultMaskRange.class)
-	public <T, S extends Comparable<T>> Iterable<T> maskRange(Iterable<T> in,
-		DefaultMaskRange.Range<S> range)
+	public <T, U extends Comparable<T>> Iterable<T> maskRange(Iterable<T> in,
+		Pair<U, U> range)
 	{
 		@SuppressWarnings("unchecked")
 		Iterable<T> result = (Iterable<T>) ops().run(DefaultMaskRange.class, in,
--- a/src/main/java/net/imagej/ops/mask/maskHigh/DefaultMaskHigh.java
+++ b/src/main/java/net/imagej/ops/mask/maskHigh/DefaultMaskHigh.java
@@ -1,0 +1,30 @@
+
+package net.imagej.ops.mask.maskHigh;
+
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+import net.imagej.ops.Ops;
+import net.imagej.ops.special.function.AbstractBinaryFunctionOp;
+import net.imglib2.type.Type;
+
+import org.scijava.plugin.Plugin;
+
+/**
+ * Creates an Iterable from all the elements in the input that are greater than
+ * the given value
+ *
+ * @author Richard Domander (Royal Veterinary College, London)
+ */
+@Plugin(type = Ops.Mask.MaskHigh.class)
+public class DefaultMaskHigh<T extends Type<T>, S extends Comparable<T>>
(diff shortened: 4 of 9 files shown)


rimadoma:
Disregard: tests don't pass

rimadoma:
BinaryMask could be made HybridCFI

src/main/java/net/imagej/ops/threshold/ThresholdNamespace.java
```diff
@@ -122,6 +126,24 @@
 		return result;
 	}
 
+	// -- binary mask --
+
+	@OpMethod(op = net.imagej.ops.threshold.binaryMask.DefaultBinaryMask.class)
```
dietzc: there already exist threshold ops in ops. see https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/threshold/ThresholdNamespace.java. However, we could / should modify these ops to also get a `Condition` (`lt`, `gt`, ... etc) as input parameter. Then we could cover all the cases you implemented here. @stelfrich are you still refactoring threshold ops?

src/main/java/net/imagej/ops/threshold/binaryMask/DefaultBinaryMask.java
```diff
@@ -0,0 +1,108 @@
...
+            final Img<BitType> output) {
+        final Cursor<BitType> cursor = output.localizingCursor();
+        final RandomAccess<T> intervalAccess = interval.randomAccess();
+        final long[] maskPosition = new long[3];
+        final long[] intervalPosition = new long[interval.numDimensions()];
+
+        while (cursor.hasNext()) {
+            cursor.fwd();
+            cursor.localize(maskPosition);
+            System.arraycopy(maskPosition, 0, intervalPosition, 0, 3);
```
dietzc: you could call `intervalAccess.setPosition(cursor)` directly

src/main/java/net/imagej/ops/threshold/binaryMask/DefaultBinaryMask.java
```diff
@@ -0,0 +1,108 @@
...
+@Plugin(type = Ops.Threshold.BinaryMask.class)
+public class DefaultBinaryMask<T extends RealType<T> & NativeType<T>>
+        extends AbstractBinaryHybridCF<RandomAccessibleInterval<T>, DefaultBinaryMask.Thresholds<T>, Img<BitType>>
+        implements Ops.Threshold.BinaryMask, Contingent {
+    @Override
+    public void compute2(final RandomAccessibleInterval<T> interval, final Thresholds<T> thresholds,
+            final Img<BitType> output) {
+        final Cursor<BitType> cursor = output.localizingCursor();
+        final RandomAccess<T> intervalAccess = interval.randomAccess();
+        final long[] maskPosition = new long[3];
```
dietzc: do you want to rely on `3d` images? you could use `new long[interval.numDimensions()]`

src/main/java/net/imagej/ops/threshold/binaryMask/DefaultBinaryMask.java
```diff
@@ -0,0 +1,108 @@
...
+                cursor.get().setOne();
+            }
+        }
+    }
+
+    @Override
+    public Img<BitType> createOutput(final RandomAccessibleInterval<T> interval, final Thresholds<T> thresholds) {
+        final long[] dimensions = new long[interval.numDimensions()];
+        interval.dimensions(dimensions);
+        return ArrayImgs.bits(dimensions);
```
dietzc: `ops().create().img(...)`

dietzc: also I would initialize the `create` using `Functions.create(...)`. However, @ctrueden will change the way how one creates `Function`s and `Computer`s and `Hybrid`s.

src/main/java/net/imagej/ops/threshold/ThresholdNamespace.java
```diff
@@ -1346,6 +1347,14 @@
 		return result;
 	}
 
+	@OpMethod(op = net.imagej.ops.threshold.thresholdElements.DefaultThresholdElements.class)
```
dietzc: I think this op is really interesting. I would rather call it `FilteringOp` or something like this which takes some `Iterable<O>` and return some other `Iterable<O>` where all `O` fulfill some `Condition<O>`.

src/main/java/net/imagej/ops/threshold/ThresholdNamespace.java
ctrueden: @dietzc Note that @LeonYang5114 and I removed the `Condition` interface and related ops. I suppose we could reintroduce it, if you see a use case for it here.

src/main/java/net/imagej/ops/threshold/binaryMask/DefaultBinaryMask.java
ctrueden: Don't let me hold you up, though: I will mass convert everything on the master branch to the new approach once it is fully working. Get your stuff onto master soon, to avoid needing to do the conversion yourself during a later rebase! :wink:

src/main/java/net/imagej/ops/threshold/ThresholdNamespace.java
stelfrich: @dietzc I haven't had time to create a PR from it. I find the idea of having an upper and a lower threshold pretty intriguing! I'd like to wait for a clean up of `ThresholdNamespace` till the discussion on nested namespaces has progressed (after talking with @ctrueden about that). So I'd go ahead and merge this and we might get back to the idea while cleaning up?

dietzc: I'd not merge this, because it introduces redundant concepts to existing functionality.

stelfrich:
@rimadoma @dietzc I had worked on something similar some weeks ago (but did not create a PR from it) when working on #112.

You can find my implementation on the [`multiple-thresholds`](https://github.com/imagej/imagej-ops/tree/multiple-thresholds) branch. The idea is to have a `ApplyThresholdCollection` op that applies several threshold methods to a value and combines them with a logic `AND`. The `ApplyConstantThresholdPair` op is still a lot of duplicated code that applies the `ApplyThresholdCollection` pixel-wise to an input image while creating the respective ops for the minimum threshold and the maximum threshold (i.e. `output = thresholdMin < I && I < thresholdMax`, with `I` being the intensity value of the pixel).

Is that what you have implemented here, @rimadoma?

src/main/java/net/imagej/ops/threshold/binaryMask/DefaultBinaryMask.java
rimadoma: Could that cause an exception if `interval.numDimensions() != output.numDimensions()`?

rimadoma:
@stelfrich Yes, I think your branch / PR implements nicely what I had in mind. I'll drop `DefaultBinaryMask` from this PR, and just add the general "Filter" op. You may need to show how to use your op at some point though ;)

stelfrich:
> You may need to show how to use your op at some point though ;)

```java
final UnsignedShortType thresholdMin = new UnsignedShortType(30000);
final UnsignedShortType thresholdMax = new UnsignedShortType(40000);
ThresholdPair<UnsignedShortType> thresholdPair =
    new ThresholdPair<>(thresholdMin, thresholdMax);
ops.run(ApplyConstantThresholdPair.class, out, in, thresholdPair); // option 1
ops.threshold().apply(out, in, thresholdPair); // option 2
```

About the `DefaultFilter`: it somehow violates the implicit definition of a threshold that I have built up in my head which is `T -> BitType` and/or `RAI/II<T> -> RAI/II<BitType>`. Also, it's not really a `Filter` operation like the ones in the `FilterNamespace` either. Does that make sense?

rimadoma:
@dietzc @stelfrich @ctrueden
Okay so after some mulling over I've arrived at the following solution: instead of the `Threshold` namespace create generic filtering ops "low pass", "high pass" and "band pass". They are quite similar to thresholds, but instead of `RAI<T> -> RAI<BitType>` they are `Iterable<T> -> Iterable<T>`.

In these filtering ops the generic type `T` is bound to `<T extends Type<T>>`. With the super generic `T` I painted myself into a corner: when you create a `List` from the filtered values when `T extends RealType<T>`, all the elements have the same value due some kind of shallow copying issues (see [code](https://github.com/imagej/imagej-ops/pull/417/files#diff-5e34f39588074a2423f0d4a31c6ea3b8R20) and [test](https://github.com/imagej/imagej-ops/pull/417/files#diff-1d82d90955c108828e510d66052c2d76R39)). That is, I need to be able to clone the elements of the `Iterable`, which `Type.copy()` provides nicely.

@dietzc Do you mean this [Condition](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/Condition.html)? I don't see right away how it might be used for this purpose.

imagejan:
> Also, it's not really a `Filter` operation like the ones in the `FilterNamespace` either.

@stelfrich I agree, the [`filter`](https://github.com/imagej/imagej-ops/blob/d86cff8/src/main/templates/net/imagej/ops/Ops.list#L84-L107) namespace contains mostly what's known as filters in _image processing_ (i.e. convolution filters etc.), while the [_signal processing_ definition of Filter](https://en.wikipedia.org/wiki/Filter_(signal_processing)) is a little broader (or less concise).

> They are quite similar to thresholds, but instead of RAI&lt;T&gt; -> RAI&lt;BitType&gt; they are Iterable&lt;T&gt; -> Iterable&lt;T&gt;.

@rimadoma how about naming these _masking_ operations, as they return a masked version of the original inputs?

rimadoma:
@stelfrich Sorry I missed your comment about the ambiguity of the term "filter"

@imagejan I decided to with the term "sieve". I feel like "masking" can be ambiguous too in image processing context

src/main/java/net/imagej/ops/sieve/highSieve/DefaultHighSieve.java
```diff
@@ -0,0 +1,25 @@
+package net.imagej.ops.sieve.highSieve;
+
+import net.imagej.ops.Ops;
```
stelfrich: The order of the imports is non-standard. If you are using Eclipse @rimadoma, you could _Organize Imports_ with the latest [ImageJ Eclipse preferences](https://github.com/imagej/imagej/blob/master/config/eclipse-preferences.epf) active to do that.

src/main/java/net/imagej/ops/sieve/highSieve/DefaultHighSieve.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.stream.StreamSupport;
+
+/**
+ * Creates an Iterable from all the elements in the input that are greater than the given value
+ *
+ * @author Richard Domander (Royal Veterinary College, London)
+ */
+@Plugin(type = Ops.Sieve.HighSieve.class)
+public class DefaultHighSieve<T extends Type<T>, S extends Comparable<T>>
+        extends AbstractBinaryFunctionOp<Iterable<T>, S, Iterable<T>>
```
stelfrich: Ops uses tabs instead of spaces according to the [Coding Style](http://imagej.net/Coding_style).

src/main/templates/net/imagej/ops/Ops.list
```diff
@@ -307,6 +307,11 @@ namespaces = ```
 		[name: "thin",                           iface: "Thin"],
 		[name: "topHat",                         iface: "TopHat"],
 	]],
+	[name: "sieve", iface: "Sieve", ops: [
+    	[name: "highSieve",                      iface: "HighSieve"],
```
stelfrich: Spaces and tabs mixed

src/test/java/net/imagej/ops/sieve/SieveNamespaceTest.java
```diff
@@ -0,0 +1,23 @@
...
+     * Tests that the ops of the {@code sieve} namespace have corresponding
+     * type-safe Java method signatures declared in the {@link SieveNamespace}
+     * class.
+     */
+    @Test
+    public void testCompleteness() {
+        assertComplete("sieve", SieveNamespace.class);
+    }
+
+}
```
stelfrich: Newline at the end of file missing

stelfrich:
> Sorry I missed your comment about the ambiguity of the term "filter"

No problem! :smile: 

I have added some comments that you could easily address by doing a [clean up](http://imagej.net/Coding_style#Eclipse_code_style_profiles) according to the coding style.

I am fine with the `SieveNamespace` which doesn't have any obvious other meanings that I am aware of :+1: 

Also, I like that Ops is seeing some serious Java 8 stuff going on :clap:

imagejan:
> I am fine with the SieveNamespace

I think introducing yet another terminology is not a good idea here.

Masking is a well-defined term, and e.g. the modules MaskImage and MaskObjects in CellProfiler do a similar thing, i.e. ignoring the outside of a given area, the mask.

IIUC, the difference of your ops (with regard to other masking operations in image processing) is that you skip the creation of a binary mask that serves as the input of the masking operation, but instead use the intensity information of the pixels to implicitly serve as mask. Nevertheless, I could imagine these ops living in a `mask` namespace together with an op taking a mask and an image as input, an returning a masked image. What do you think?

dietzc:
I'll also review it next week when I'm back from holiday. I agree with @imagejan, we should understand better if these concepts fit in some other namespace. I would like to avoid too many concepts and namings etc.

rimadoma:
@stelfrich Okay I think I finally managed to get Eclipse Code Formatter plugin working in IntelliJ. At least indents are now tabs, and imports are organised correctly. 

@imagejan @dietzc The ops are now in `MaskNamespace`. It didn't exist previously, so I created it.

ctrueden:
I also think we shoulf avoid the name "sieve" here (naming is hard!). In
math it refers to a certain form of numerical algorithm for discerning
numbers such as primes.

I can also review more thoroughly as needed; let me know if this would help
(on my phone right now).

dietzc:
To be still, this is more or less the same as thresholding or maybe a more general thing. It's actually applying some model per pixel (as thresholding is, too). I have the feeling that both concept should be unified. I will think more about this and write more next week!

rimadoma:
@ctrueden @dietzc 
Fair enough, I'll put this PR on hold for now, until some kind of consensus is achieved on naming. The ops _are_ very similar to threshold ops, but as @stelfrich pointed out, they do not operate on `RAI/II<T> -> RAI/II<BitType>`. Originally I wanted to make these ops super general, but ran into shallow copying issues. Thus I settled for `Iterable<T extends Type<T>>`, since the class provides `Type.copy()`  which can be used to clone without knowing the runtime type of `T`.

ctrueden:
The best way to duplicate images of arbitrary type in ops is to use op
chaining: call an op in the ops.copy namespace. It is intelligent about how
to do it. IIRC, there are several existing ops that already work this way.

On Aug 11, 2016 9:50 AM, "Richard Domander" <notifications@github.com>
wrote:
>
> @ctrueden @dietzc
> Fair enough, I'll put this PR on hold for now, until some kind of
consensus is achieved on naming. The ops are very similar to threshold ops,
but as @stelfrich pointed out, they do not operate on RAI/II<T> ->
RAI/II<BitType>. Originally I wanted to make these ops super general, but
ran into shallow copying issues. Thus I settled for Iterable<T extends
Type<T>>, since the class provides Type.copy() which can be used to clone
without knowing the runtime type of T.

rimadoma:
@ctrueden 
Specifically the problem I was facing was with:
```java
@Override
public Iterable<T> compute2(Iterable<T> iterable, Range<S> range) {
	final Stream<T> inStream = StreamSupport.stream(iterable.spliterator(), false);
	final Stream<T> filtered = inStream.filter(e -> range.min.compareTo(e) <= 0 && range.max.compareTo(e) >= 0);

	return filtered.map(Type::copy).collect(Collectors.toList());
}
```

Now if `T` was super generic (no `extends` or anything), I could not call `map()`, because `T` didn't have any means to clone the object. Thus in the case of say `Iterable<LongType>` I ended up with an `Iterable` full of shallow copies of the same object. `T extends Cloneable` didn't help, because `Cloneable` is just a naming interface, i.e. no methods.

src/main/java/net/imagej/ops/mask/maskRange/DefaultMaskRange.java
```diff
@@ -0,0 +1,53 @@
...
+	public Iterable<T> compute2(Iterable<T> iterable, Range<S> range) {
+		final Stream<T> inStream = StreamSupport.stream(iterable.spliterator(),
+			false);
+		final Stream<T> filtered = inStream.filter(e -> range.min.compareTo(
+			e) <= 0 && range.max.compareTo(e) >= 0);
+
+		return filtered.map(Type::copy).collect(Collectors.toList());
+	}
+
+	public static final class Range<S> {
```
dietzc: why not simply min, max as parameters? why do we need a specialized class here?

rimadoma: To keep the op binary

dietzc:
besides the `Iterable` to `Iterable` implementation instead of `RAI` to `RAI` : how are these ops different from: https://github.com/imagej/imagej-ops/pull/435?

Any `RAI` can be transformed to an `Iterable`, so maybe we should add specialized `Ops` for thresholding which work on `Iterable<T>` (but need to retain the ordering of objects)?

src/main/java/net/imagej/ops/mask/maskRange/DefaultMaskRange.java
dietzc: I see. One could use `Pair` here, then we don't need to introduce another class.

stelfrich:
To clarify that, @dietzc: @rimadoma's ops drop values from an `Iterable`, hence, the size of `input` and `output` might be different. The threshold ops thus far produce `output`s of the same size as the respective `input` (plus, they output `BooleanTypes`).

That might really be a stupid question at this point of the discussion but what is your use case @rimadoma? Are the `Iterable`s you are working on something other than images? If so, maybe we should open the discussion up into the direction of where to put really general ops (e.g. something like a `UtilNamespace`)..

dietzc:
I see. But then this is not really masking?!

stelfrich:
> I see. But then this is not really masking?!

Not really. It is more like a general `filterElements` / `keepElements` / `removeElements` op in a very general context because the `ouput` really is "just" something that implements `List`.

rimadoma:
@stelfrich My actual use case _is_ images, but in my thinking making the op more general would have made it more useful. Maybe this is a case of YAGNI? What I actually need to do can be done with your multiple threshold ops / threshold namespace, so maybe just close this PR?

ctrueden:
> maybe we should open the discussion up into the direction of where to put really general ops (e.g. something like a `UtilNamespace`)..

Well, non-image-specific ops will live in `scijava-ops`, whereas image-specific ones will stay in `imagej-ops`. See #109.

stelfrich:
> But of course, we still have to name things.

If these ops live in `scijava-ops`, i.e. they are completely image agnostic, then shouldn't we call them `filter` in style of Java 8 streams? That's what they are doing after all and what people should be used to anyway in the near future..

ctrueden:
I also like the name `filter`. There is a question of whether ImageJ Ops should add things into the `filter` namespace then, or should put those operations somewhere else. SciKit Image [uses the namespace `filters` for image filter operations](http://scikit-image.org/docs/dev/api/skimage.html), but of course having both `filter` and `filters` would be horribly confusing. We'll have to consider carefully what to do.

rimadoma:
I'd say the code is pretty much ready to go now once we've decided on naming / packaging

ctrueden:
I wanna finalize this during this Dresden hackathon. We should be able to hash out the naming in person, @rimadoma. OK?

ctrueden:
@rimadoma and I discussed in person at the Dec 2017 hackathon, and agreed to close this without merging in favor of @stelfrich's work on the `multiple-thresholds` branch.

ctrueden:
Regarding naming: in future, maybe SciJava Ops (i.e., non-image-specific) can have a `stream` namespace with things like `filter` and `map`? This would match Java, and be distinct from image filtering.

PR #420, opened 2016-07-12
Add CentroidTuple3d op

rimadoma:
Add an op which calculates the centroid of a collection of vectors (`? extends org.scijava.vecmath.Tuple3d`)


ctrueden:
Thanks @rimadoma!

Before we merge this, I think we really need to decide how we are going to handle vectors, matrices, etc., as core data structures. I am quite strongly against adding a vecmath dependency to ImageJ Ops core—although it is actually pretty lightweight, so could be convinced.

This op might belong in its own `imagej-ops-vecmath` project or some such. What do others think?

See also fiji/fiji#135

ctrueden:
Let's finally decide on matrix libraries at this 2017-12 hackathon! Then this PR can finally be resolved.

rimadoma:
Yes, this thing could go with my other linear algebra ops.

rimadoma:
I resurrected this PR by updating it to use JOML, which is now the consensus linear algebra library (?). Suggestions about for the namespace of this op are welcome. Could start `math/lin-alg/`? I have at least one other op that would fit there. In any case, I expect eventually this'll better fit into `scijava-ops`.

rimadoma:
If I wanted to create the same op method for a collection of `Vector3d` and `Vector3f` what would be the best way? Just `centroid(Iterable<Vector3d>)` will have the same erasure for any type. Also the _Vector_ classes have no other common supertype than `Externalizable`.

PR #424, opened 2016-07-18
Bugfix in DefaultContour Op

haesleinhuepf:
Hey guys,

I think, I found the bug reported in the imagej forum:

http://forum.imagej.net/t/issue-with-polygons-in-ops/2227

Furthermore, I added two tests for using the DefaultContour stuff with BitType and BoolType images. Both in principle work, but deliver different (and both valid) results. However, the fact that both polygons are different, suggests that still something is wrong in the DefaultContour implementation... 

However, with this BugFix, I can start using the Polygons....

If you have any questions, just let me know :)

Cheers,
Robert

### DIFF
--- a/src/main/java/net/imagej/ops/geom/geom2d/DefaultContour.java
+++ b/src/main/java/net/imagej/ops/geom/geom2d/DefaultContour.java
@@ -213,7 +213,7 @@ public class DefaultContour<B extends BooleanType<B>> extends
 		// find first black pixel
 		while (cInput.hasNext()) {
 			// we are looking for a black pixel
-			if (cInput.next().get() == isInverted) {
+			if (cInput.next().get() == var.get()) {
 				raInput.setPosition(cInput);
 				raInput.localize(startPos);
 
@@ -226,7 +226,7 @@ public class DefaultContour<B extends BooleanType<B>> extends
 				cNeigh.reset();
 
 				while (cNeigh.hasNext()) {
-					if (cNeigh.next().get() == isInverted) {
+					if (cNeigh.next().get() == var.get()) {
 
 						boolean specialBacktrack = false;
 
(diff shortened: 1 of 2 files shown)


dietzc:
great, thanks! So there is still something wrong in the implementation? If so, could you open a separate issue for that. Then we can assign and fix it (hopefully).

One more question: what are you exactly fixing here? I'm not sure what was wrong until now? Can you explain?

Thanks for your efforts!
Christian

haesleinhuepf:
I just found a bug in the bugfix. So stay tuned ;-)

haesleinhuepf:
Ok, I take the pull-request back. It was obviously just a misunderstanding in parameters of the contour function. "isInverted" may be renamed, and the documentation clearified as @tibuch suggested in the forum.

dietzc:
maybe it would be a good idea to simply remove the `isInverted` and write in the doc: `assumption: white foreground, black background`.

haesleinhuepf:
What is white? ;-)
I suggest talking in values. I would say true for foreground, false for background. Thus, the parameter "isInverted" could be renamed to something like "foregroundValue" and have default value true...

tibuch:
Since DefaultContour needs a [BooleanType](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/geom/geom2d/DefaultContour.java#L60) I would also suggest as default value `True` and remove the isInverted parameter. If someone wants to calculate the contour of the inverted label, the label could be inverted with a different op.

ctrueden:
> I would also suggest as default value True and remove the isInverted parameter.

Thanks @tibuch; I strongly agree. Otherwise, every binary op needs this extra parameter. Better to assume True=included, and then create something like Views.invert for flipping boolean data structures on the fly.

PR #425, opened 2016-07-18
DefaultContour remove isInverted parameter

tibuch:
The isInverted parameter is a bit confusing.

I removed the parameter and updated all tests.

This is a possible solution to this discussion:
https://github.com/imagej/imagej-ops/pull/424#issuecomment-233307097

### DIFF
--- a/src/main/java/net/imagej/ops/geom/geom2d/DefaultContour.java
+++ b/src/main/java/net/imagej/ops/geom/geom2d/DefaultContour.java
@@ -53,6 +53,9 @@ import org.scijava.plugin.Plugin;
 /**
  * Generic implementation of {@code geom.contour}.
  * 
+ * This implementation assumes that foreground-pixels are True and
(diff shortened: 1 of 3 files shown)


dietzc:
@tibuch unit tests still pass? 

In general I like removing this confusing parameter. Maybe you can add the assumption about `true = foreground` to the javadoc?

tibuch:
@dietzc all tests still pass.

src/main/java/net/imagej/ops/geom/geom2d/DefaultContour.java
```diff
@@ -53,6 +53,9 @@
 /**
  * Generic implementation of {@code geom.contour}.
  * 
+ * This implementation assumes that foreground-pixels are True and
```
ctrueden: I am strongly in favor of _all_ binary ops working this way, always.

PR #427, opened 2016-07-19
Fix Javadocs

### LINKED ISSUE #426: Fix Javadoc rendering of generated Ops interfaces
The javadoc for the autogenerated Ops classes is partially broken. For example, the [`Ops.Convert.Bit` javadoc](http://javadoc.imagej.net/ImageJ/net/imagej/ops/Ops.Convert.Bit.html) contains the following (invalid) HTML:

```html
<div class="block">Base interface for "bit" operations.
 <p>
 Implementing classes should be annotated with:
 </p>

 <pre></div>
```
which is an incomplete rendering of the code in the [`Ops.vm` template](https://github.com/imagej/imagej-ops/blob/master/src/main/templates/net/imagej/ops/Ops.vm#L65-L67):

```java
	/**
	 * Base interface for "$op.name" operations.
	 * <p>
	 * Implementing classes should be annotated with:
	 * </p>
	 *
	 * <pre>
	 * @Plugin(type = Ops.${opClass}.class)
	 * </pre>
	*/
```

stelfrich:
Fixes errors in Javadoc that to the most extent were not considered errors before Java 8. With these fixes `mvn javadoc:javadoc` succeeds with doclinting for Java 8 enabled.

Closes #426.


src/main/java/net/imagej/ops/threshold/huang/ComputeHuangThreshold.java
```diff
@@ -42,7 +42,7 @@
 //plugin found in Fiji (version 1.14).
 
 /**
- * Implements Huang's threshold method by Huang & Wang.
+ * Implements Huang's threshold method by Huang {@literal &} Wang.
```
ctrueden: You can also write `&amp;` instead; slightly shorter in this case.

ctrueden:
Thank you very much!

PR #430, opened 2016-07-21
Fix geom tests.

tibuch:
@ctrueden  and @dietzc

I looked really carefully at the geom tests:

- testCentroidMesh(): MarchingCubes stored the vertices in clockwise order instead of counter-clockwise. This was fixed and leads now to positive centroid coordinates, which makes sense because the whole structure was only in positive coordinates.
- testBoundaryPixelCountConvexHull(): The change is again based on the MarchingCubes fix. Now all vertices are stored in a different order, which leads to different point selections during the convex hull computation. I tested the new convex hull for correctness and everything looks fine. Just the number of points changed, because of the selection. 

(All tests are passing now.)


ctrueden:
Thank you @tibuch !

PR #432, opened 2016-07-27
Add local versions for global thresholds

stelfrich:
This PR establishes an abstract class `LocalThresholdMethodHistogram` and auto-generated sub-classes that apply the available global threshold methods (e.g. Huang, Otsu, etc) to each neighborhood of the input image individually.

### DIFF
--- a/src/main/java/net/imagej/ops/threshold/LocalThresholdMethodHistogram.java
+++ b/src/main/java/net/imagej/ops/threshold/LocalThresholdMethodHistogram.java
@@ -69,16 +69,17 @@ public abstract class LocalThresholdMethodHistogram<T extends RealType<T>>
 	public void compute2(final Iterable<T> neighborhood, final T center,
 		final BitType output)
 	{
+		// TODO Move to initialize when NIL objects are available
 		if (applyThreshold == null) {
 			applyThreshold = Computers.binary(ops(), Ops.Threshold.Apply.class, output,
 				center, center);
 		}
-		
+
 		// Compute histogram for neighborhood
 		final Histogram1d<T> hist = histCreator.compute1(neighborhood);
 
 		// Compute threshold
-		T computedThreshold = center.createVariable();
+		final T computedThreshold = center.createVariable();
 		thresholdComputer.compute1(hist, computedThreshold);
 
 		// Apply threshold
--- a/src/main/java/net/imagej/ops/threshold/LocalThresholdMethodHistogram.java
+++ b/src/main/java/net/imagej/ops/threshold/LocalThresholdMethodHistogram.java
@@ -1,0 +1,90 @@
+/*
+ * #%L
+ * ImageJ software for multidimensional image processing and analysis.
+ * %%
+ * Copyright (C) 2014 - 2016 Board of Regents of the University of
+ * Wisconsin-Madison, University of Konstanz and Brian Northan.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * 
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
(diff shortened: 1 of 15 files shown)


src/main/java/net/imagej/ops/threshold/LocalThresholdMethodHistogram.java
```diff
@@ -0,0 +1,90 @@
...
+			Histogram1d.class, in1() == null ? Iterable.class : in1());
+
+		thresholdComputer = getThresholdComputer();
+	}
+
+	@Override
+	public void compute2(final Iterable<T> neighborhood, final T center,
+		final BitType output)
+	{
+		if (applyThreshold == null) {
```
dietzc: initialization not possible in initialize? if not: why? and could we add a `FIXME`. Maybe we can use `NIL` later.

src/main/java/net/imagej/ops/threshold/LocalThresholdMethodHistogram.java
```diff
@@ -0,0 +1,90 @@
...
+			applyThreshold = Computers.binary(ops(), Ops.Threshold.Apply.class, output,
+				center, center);
+		}
+		
+		// Compute histogram for neighborhood
+		final Histogram1d<T> hist = histCreator.compute1(neighborhood);
+
+		// Compute threshold
+		T computedThreshold = center.createVariable();
```
dietzc: can be final

src/main/templates/net/imagej/ops/threshold/ApplyThresholdMethodLocal.vm
```diff
@@ -0,0 +1,91 @@
...
+import net.imagej.ops.threshold.apply.LocalThreshold;
+import net.imglib2.histogram.Histogram1d;
+import net.imglib2.img.Img;
+import net.imglib2.type.logic.BitType;
+import net.imglib2.type.numeric.RealType;
+
+import org.scijava.plugin.Plugin;
+
+/**
+ * Ops that apply a global threshold locally to an {@link Img}.
```
dietzc: why `Img`?

src/main/java/net/imagej/ops/threshold/LocalThresholdMethodHistogram.java
```diff
@@ -0,0 +1,90 @@
...
+import net.imglib2.type.logic.BitType;
+import net.imglib2.type.numeric.RealType;
+
+/**
+ * {@link AbstractCenterAwareComputerOp} for use in {@link LocalThreshold}s.
+ * 
+ * @author Stefan Helfrich (University of Konstanz)
+ */
+public abstract class LocalThresholdMethodHistogram<T extends RealType<T>>
+	extends AbstractCenterAwareComputerOp<T, BitType>
```
dietzc: as it is a computer, we could be more general and say `? extends BooleanType`? Like this we could also pass `BoolType` objects.

src/main/java/net/imagej/ops/threshold/LocalThresholdMethodHistogram.java
stelfrich: I could move it to initialize with something like this

```applyThreshold = (BinaryComputerOp) Computers.binary(ops(), Ops.Threshold.Apply.class, out()==null?BitType.class:out(), in2()==null?RealType.class:in2(), in2()==null?RealType.class:in2());```

which works but feels really hacky. I have therefore added a `TODO` for moving this code to initialize, once the `NIL` stuff is implemented.

dietzc: great, thanks!

src/main/java/net/imagej/ops/threshold/LocalThresholdMethodHistogram.java
stelfrich: I think it is doable but we currently don't have any obvious ways to create `Img/RAI/...<BoolType>`, or do we @dietzc? So there's not really a chance that a user passes something like that to the `Computer`s, as far as I can tell. If you could tell me how to create an `Img<BoolType>` I could at least write a test to verify that the proposed changes will work out as expected...

dietzc: you are right, but this doesn't mean that this won't be possible in the future and there is no disadvantage when we support it, even if we don't test it (correct me if I'm wrong).

stelfrich: That's true. And I have implemented some of the changes here, which are compiling. But without being able to create an `Img<BoolType>` in a test, to verify that the implementation really is future-proof, I feel that we should rather not open up the possibility?

dietzc: If you implement against `<BooleanType>` and test it against `BitType` that's fine. Then it has to work for `BoolType` as well, right? You also don't test every implementation of `RealType` against each and every `RealType`...

stelfrich:
- [ ] Composition instead of inheritance?

ctrueden:
The merge of this PR broke the `LocalThresholdTest`. At first glance, it seems like the merge itself broke it, as opposed to any commit on the branch—which means it is the interaction of recent changes on master plus this topic branch?

@stelfrich Do you have time to investigate and fix?

stelfrich:
It wasn't your bisect-fu, @ctrueden, the test were passing prior to the merge -- I double checked.

I will investigate immediately.

ctrueden:
Thank you very much!

stelfrich:
@ctrueden I pushed a "fix" to master. We will, however, at some point have to think about how to handle invalid thresholds (which was the reason for the failing tests). I have opened another issue on the topic (#445).

PR #433, opened 2016-08-01
Transform.Scale on RAI

stelfrich:
The implementation of `Ops.Transform.Scale` has two major drawbacks

1. It operates exclusively on `Img`
2. It explicitly copies the content from the input to the output `Img`

I have changed the implementation to operate on `RandomAccessibleInterval` (addressing _1._) where the returned `RAI` is actually a view on the input, i.e. there is no copying (addressing _2._). I have also renamed it from `ScaleImg` to `DefaultScaleView` to better fit into the naming scheme of the transformations.

### DIFF
--- a/src/main/java/net/imagej/ops/transform/scale/DefaultScaleView.java
+++ b/src/main/java/net/imagej/ops/transform/scale/DefaultScaleView.java
@@ -55,10 +55,10 @@ import org.scijava.plugin.Plugin;
  * @author Martin Horn (University of Konstanz)
  * @author Stefan Helfrich (University of Konstanz)
  */
-@Plugin(type = Ops.Transform.Scale.class, priority = Priority.HIGH_PRIORITY)
-public class ScaleImg<T extends RealType<T>> extends
+@Plugin(type = Ops.Transform.ScaleView.class, priority = Priority.HIGH_PRIORITY)
+public class DefaultScaleView<T extends RealType<T>> extends
 	AbstractUnaryFunctionOp<RandomAccessibleInterval<T>, RandomAccessibleInterval<T>>
-	implements Ops.Transform.Scale, Contingent
+	implements Ops.Transform.ScaleView, Contingent
 {
 
 	/**
--- a/src/main/java/net/imagej/ops/transform/scaleView/DefaultScaleView.java
+++ b/src/main/java/net/imagej/ops/transform/scaleView/DefaultScaleView.java
@@ -106,15 +106,14 @@ public class DefaultScaleView<T> extends
 
 	@Override
 	public RandomAccessibleInterval<T> compute1(RandomAccessibleInterval<T> input) {
-		RandomAccessible<T> extended = RAIs.extend(input, outOfBoundsFactory);
-		RealRandomAccessible<T> interpolated = interpolateOp.compute1(extended);
-		AffineRealRandomAccessible<T, AffineGet> transformed = RealViews.affineReal(
-			interpolated, new Scale(scaleFactors));
-		RandomAccessibleOnRealRandomAccessible<T> rasterized = rasterOp.compute1(
-			transformed);
-		IntervalView<T> intervaled = intervalOp.compute1(rasterized);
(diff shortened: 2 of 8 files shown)


ctrueden:
Thanks! But the op needs to be called `scaleView` please.

stelfrich:
> Thanks! But the op needs to be called `scaleView` please.

Done. (It's not that I hadn't thought about that name, but it contains some not-`Views` logic which made me not call it `ScaleView`)

I have opened a related issue (see #434) on keeping/implementing non-`Views` variants of the transform ops.

src/main/templates/net/imagej/ops/Ops.list
```diff
@@ -403,7 +403,7 @@ namespaces = ```
 		[name: "project",                        iface: "Project"],
 		[name: "rasterView",                     iface: "RasterView"],
 		[name: "rotateView",                     iface: "RotateView"],
-		[name: "scale",                          iface: "Scale",                          aliases: ["resize"]],
+		[name: "scaleView",                      iface: "ScaleView",                      aliases: ["resize"]],
```
ctrueden: Please change the alias also to `resizeView`.

src/main/java/net/imagej/ops/transform/scale/DefaultScaleView.java
```diff
@@ -55,10 +55,10 @@
  * @author Martin Horn (University of Konstanz)
  * @author Stefan Helfrich (University of Konstanz)
  */
-@Plugin(type = Ops.Transform.Scale.class, priority = Priority.HIGH_PRIORITY)
-public class ScaleImg<T extends RealType<T>> extends
+@Plugin(type = Ops.Transform.ScaleView.class, priority = Priority.HIGH_PRIORITY)
```
ctrueden: This now needs to live in package `scaleView`, not `scale`.

ctrueden:
One more question: should we consolidate `interpolateView` and `scaleView`? The `interpolateView` is more naive, whereas the `scaleView` does some prep work to make sure the interpolator works better around the edges of the image.

stelfrich:
> to make sure the interpolator works better around the edges of the image

Good point, @ctrueden! I have not tested this properly, but depending on the `InterpolatorFactory` the `interpolateView` will just fail when the interpolator tries to access coordinates outside of the input `Interval`..

stelfrich:
Remaining TODOs (from @dietzc):
- [ ] Replace `Views.*` calls with ops
- [ ] Check if `T extends RealType` is required
- [ ] Optional `OutOfBoundsStrategy` as parameter

src/main/java/net/imagej/ops/transform/scaleView/DefaultScaleView.java
```diff
@@ -70,6 +72,10 @@
 	@Parameter
 	private InterpolatorFactory<T, RandomAccessible<T>> interpolator;
 
+	@Parameter(required = false)
+	private OutOfBoundsFactory<T, RandomAccessibleInterval<T>> outOfBoundsFactory =
+		new OutOfBoundsMirrorFactory<>(OutOfBoundsMirrorFactory.Boundary.SINGLE);
```
ctrueden: What if the user passes in `null`? Then won't that overwrite the default here? Did you try it?

stelfrich:
> What if the user passes in `null`? Then won't that overwrite the default here? Did you try it?

I have replaced the `extendOp` with a call to `RAIs.extend` which does exactly that check. However, when I debug the additional test (i.e. `testOutOfBoundsFactoryIsNull`), executing the `ScaleView` with

```java
ops.run(DefaultScaleView.class, in, scaleFactors, nLinearInterpolatorFactory, null);
```
will not replace the default `OutOfBoundsFactory`. What kind of black magic is doing this, @ctrueden? If this is the default case, you cannot set an optional `outOfBoundsFactory` in any op to null (that implies that the user is doing in explicitly and knows what he/she is doing), e.g. when the input is already extended..

PR #435, opened 2016-08-02
Add multiple thresholds for range filtering

stelfrich:
This adds the ability to apply a `Pair` of constant thresholds where `Pair.A` denotes the lower threshold and `Pair.B` denotes an upper threshold. Outputs will be computed as `Pair.A < I && I < Pair.B` (where `I` is the element intensity).

The PR also adds an `ApplyThresholdCollection` op that applies a `Collection<ApplyThreshold>` to an element by computing `out = thresholdOp1(in) && thresholdOp2(in) && thresholdOp3(in) && ...`.

### DIFF
--- a/src/main/java/net/imagej/ops/threshold/ThresholdNamespace.java
+++ b/src/main/java/net/imagej/ops/threshold/ThresholdNamespace.java
@@ -53,6 +53,7 @@ import org.scijava.plugin.Plugin;
  * The threshold namespace contains operations related to binary thresholding.
  *
  * @author Curtis Rueden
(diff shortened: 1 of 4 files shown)


dietzc:
after a discussion with @stelfrich in person we decided to close this PR as the suggested code can be generalized and combined with the current threshold architecture. We want to have `Condition` object which then can evaluate arbitrary `Condition`s (especially no restriction on thresholds etc). This object will then be used in _all_ threshold implementations (and potentially more implementations). We also want to have a `View` on thresholded images, rather can copying the images in the threshold method. This can then be done in the "burnIn" method as needed.

PR #436, opened 2016-08-09
HoG2DResult.tif: version with new SCIFIO headers

stelfrich:
With the introduction of SCIFIO metadata in TIFF headers, the file used in `HistogramOfOrientedGradients2DTest` for comparison became invalid. That is, there is no information on the third axis which makes SCIFIO return a 2D image although the test excepts a 3D image.


ctrueden:
Thanks! However, I would rather fix SCIFIO to still treat this file as a 3D image. Otherwise, all TIFFs ever which were not written by SCIFIO will be read incorrectly, from the perspective of the overwhelming majority of users. In other words: I think this is a bug in SCIFIO that needs to be fixed.

ctrueden:
Closed in favor of scifio/scifio#321

Issue #440, opened 2016-09-05
FillHoles is not working as expected


tibuch:
See discussion: http://forum.imagej.net/t/fillholes-imagej-ops/2591

PR #441, opened 2016-09-05
Fix fill holes

tibuch:
@dietzc, @ctrueden this should fix https://github.com/imagej/imagej-ops/issues/440.

All tests are passing.


dietzc:
wonderful, thanks!

PR #443, opened 2016-09-08
Clean up generics in TransformNamespace

stelfrich:
As discussed with @dietzc, this PR removes unnecessary type restrictions.


ctrueden:
Great!

PR #446, opened 2016-09-12
Watershed Ops added

SimonSchmid:


dietzc:
@SimonSchmid can you rebase your local branch such that your branch is in sync with the current master? thanks!

src/main/java/net/imagej/ops/image/watershed/Watershed.java
```diff
@@ -0,0 +1,393 @@
...
+
+	@Parameter(required = true)
+	boolean eightConnectivity;
+
+	@Parameter(required = false)
+	RandomAccessibleInterval<B> mask;
+
+	@SuppressWarnings("null")
+	@Override
+	public void compute1(final RandomAccessibleInterval<T> in, final ImgLabeling<Integer, IntType> out) {
```
dietzc: We shouldn't ask for a `ImgLabeling`, rather we want some a https://github.com/imglib/imglib2-roi/blob/master/src/main/java/net/imglib2/roi/geometric/PointCollection.java I guess.

src/main/java/net/imagej/ops/image/watershed/Watershed.java
```diff
@@ -0,0 +1,393 @@
...
+@Plugin(type = Ops.Image.Watershed.class)
+public class Watershed<B extends BooleanType<B>, T extends RealType<T>>
+		extends AbstractUnaryHybridCF<RandomAccessibleInterval<T>, ImgLabeling<Integer, IntType>>
+		implements Ops.Image.Watershed, Contingent {
+
+	@SuppressWarnings("rawtypes")
+	private UnaryFunctionOp<FinalInterval, ImgLabeling> createOp;
+
+	@Parameter(required = true)
+	boolean eightConnectivity;
```
dietzc: private missing

src/main/java/net/imagej/ops/image/watershed/Watershed.java
SimonSchmid: So you mean a list of PointCollections? For every label a collection?

haesleinhuepf: Why PointCollection? ImgLabelling appears a valid result type of a Watershed, from my point of view....

dietzc: @haesleinhuepf absolutely. `PointCollection` just for the seeds. Output should be `ImgLabeling` of course!

SimonSchmid: Ah ok, so you mean in the WatershedSeeded class. Advantage of using ImgLabeling for seeds is that the labels of the segments can be determined, because every segment will be labeled accoring to its seed.

dietzc: I understand. Ok. then leave it as it is  `ImgLabeling`.

ctrueden:
Superseded by #510.

PR #447, opened 2016-09-13
improvement of DT and calibrated DT added

SimonSchmid:


dietzc:
please rebase your branch with the current master such that we can easily see the changes and do the review. Thanks for your work!!

Issue #452, opened 2016-09-30
Fitting Ellipse


tibuch:
I would like to verify major- and minor-axis features (geometric 2d) https://github.com/imagej/imagej-ops/issues/449.

My understanding of major- and minor-axis is, that we fit an ellipse to the given shape and take the major- and minor-axis of the ellipse.

Since I am trying to verify the numbers I am looking for implementations in Matlab/python. 

The problem is, that I found a lot of different results. All are more or less similar (+/- 5).

What I found out about the current ops-implementation:
- [Based on ImageJ1 (EllipseFitter -> getEllipseParam()) ](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/geom/geom2d/DefaultMinorMajorAxis.java#L60) which is probably based on [Fitting an ellipse to an arbitrary shape: implications for strain analysis](http://www.sciencedirect.com/science/article/pii/S0191814103000932) 
- Adapted to polygons, based on this [paper](https://www.google.de/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&ved=0ahUKEwixramIq7fPAhWhNpoKHZEoAdYQFggeMAA&url=https%3A%2F%2Fiuks.informatik.tu-muenchen.de%2F_media%2Fmembers%2Fsteger%2Fpublications%2F1996%2Ffgbv-96-05-steger.pdf&usg=AFQjCNEwLW6iznp_ngg_6C9oZE7Mo0MF1g&sig2=XZ1GpM0-6G9GHf79zY12uA)

What I found about minor- and major-axis:
- [Use the minimum bounding box (python).](http://stackoverflow.com/questions/13542855/python-help-to-implement-an-algorithm-to-find-the-minimum-area-rectangle-for-gi)
- [Use a LS estimation to fit an ellipse (Matlab).](https://de.mathworks.com/matlabcentral/fileexchange/3215-fit-ellipse)
- [Use a fast and non-iterative ellipse fit (Matlab).](https://de.mathworks.com/matlabcentral/fileexchange/22684-ellipse-fit--direct-method-)

Questions:
1. Keep our implementation and assume correctness?
2. Replace our implementation?

dietzc:
@haesleinhuepf @bnorthan @ctrueden @stelfrich do you guys have any opinion on this?

dietzc:
[...]

Issue #449, opened 2016-09-23
Improve FeatureSet Tests


tibuch:
Test and improve FeatureSets.
- [x] 2D Geometric
- [ ] 3D Geometric
- [x] Statistic
- [ ] Haralick
- [ ] Tamura 2D
- [ ] Zernike
- [ ] Histogram

tibuch:
@ctrueden and @dietzc why do we have expensive geom tests and normal ones?

dietzc:
I think historical reasons. If you find "normal" ones which cover all the cases, it's absolutely OK.

ctrueden:
@tibuch I thought you were the one who authored those expensive tests! No?

I would much rather have only fast-running tests but which still exercise the algorithms sufficiently. But my understanding was that someone (you? someone else?) thought the small ones might not be rich/complex enough to fully exercise the algorithms?

tibuch:
@ctrueden I guess the tests were added by @seebacherd. 

I will replace the images by smaller but complexer (curves, holes and so on) ones. 

I would say the accuracy of the tests depends a lot more on small epsilons.

ctrueden:
@tibuch Thanks, sounds great!

PR #453, opened 2016-10-05
Fitting ellipse to point set

tibuch:
I implemented and tested a fitting ellipse implementation https://github.com/imagej/imagej-ops/issues/452.

I verified the results with the matlab-implementation from http://research.microsoft.com/en-us/um/people/awf/ellipse/ellipse-pami.pdf.

@ctrueden or @dietzc  could you check the copyright notice? https://github.com/imagej/imagej-ops/blob/fittingEllipseToPointSet/src/main/java/net/imagej/ops/geom/geom2d/DefaultFittingEllipse.java#L24

PS: All tests are passing :)


tpietzsch:
> I verified the results with the matlab-implementation from http://research.microsoft.com/en-us/um/people/awf/ellipse/.
That link doesn't lead to ellipse fitting description, just to awf's homepage. Can you please add a link to the paper/description?

@dietzc  @tibuch I have somewhere a `HyperEllipsoid` and fitting that to noisy point clouds. It has 
 * a method of fitting a hyper-ellipsoid to points I adapted from BoneJ (who adapted it from a Matlab script). It uses algebraic distance instead of geometric, so it doesn't handle noise very well, but with enough points it seems rather accurate.
 * computing the shortest (i.e., perpendicular) distance from a point to a (hyper-)ellipsoid (adapted from a C++ implementation). This is a surprisingly difficult problem, and an iterative solution.
 * RANSAC-like random sampling using the above things to robustly fit an a (hyper-)ellipsoid to a noisy point cloud with outliers.

See this forum discussion http://forum.imagej.net/t/licensing-question-for-imglib-algorithm-addition/1255 and this imglib-algorithm branch https://github.com/imglib/imglib2-algorithm/compare/ellipsoid-fitting

It would be good if we could put this into imglib2 somewhere. I wouldn't put it into `roi` probably.
A `HyperEllipsoid` is not a ROI, just the mathematical object. (You can make a ROI from it very easily though, it's a borderline case...)

tibuch:
@tpietzsch changed the link: http://research.microsoft.com/en-us/um/people/awf/ellipse/ellipse-pami.pdf

tpietzsch:
> @tpietzsch changed the link: http://research.microsoft.com/en-us/um/people/awf/ellipse/ellipse-pami.pdf

That still doesn't work for me!?

tibuch:
@tpietzsch strange. this one: https://pdfs.semanticscholar.org/090d/25f94cb021bdd3400a2f547f989a6a5e07ec.pdf

tpietzsch:
@tibuch thank you

tibuch:
@tpietzsch I am trying to verify all features of ops https://github.com/imagej/imagej-ops/issues/449 and I couldn't find reference numbers for the Major-Minor-Axis features. That's why I ended up implementing fitting ellipse. 

But I am happy to use your `HyperEllipsoid`.

tpietzsch:
Hmm, what are you actually trying to do? The geom2d feature stuff reads as if you are trying to find the smallest (area?) ellipse containing all the points? That is different from fitting an ellipse to a set of points, for example in the sense of: find the ellipse that minimizes mean squared distance from points to the ellipse.

tibuch:
> Hmm, what are you actually trying to do?
I try to find a more generic description of a shape in form of an ellipse. 

> The geom2d feature stuff reads as if you are trying to find the smallest (area?) ellipse containing all the points?
As far as I understand the current major-minor-axis implementation:
The best fitting ellipse is the ellipse with the "same" area as the polygon. In the code this paper is referenced: http://www.sciencedirect.com/science/article/pii/S0191814103000932

> That is different from fitting an ellipse to a set of points
It is different, but I think as long as it describes the shape (which could be a Polygon, IterableInterval or PointCollection) it works.

The reason why I am looking for a different implementation is to decouple the fitting ellipse implementation from the major-minor-axis implementation and to verify the numbers against other known implementations.

tibuch:
I guess this PR can be closed. The fitting ellipse implementation used in minor-major-axis computations is correct as far as I know and I was able to verify all geometric 2D features. 

> The geom2d feature stuff reads as if you are trying to find the smallest (area?) ellipse containing all the points?

This is true and this implementation does not do that.

PR #459, opened 2016-10-23
Scale Fixes

dietzc:
Minor fixes on `Scale` implementations. Can you review @stelfrich?

### DIFF
--- a/src/main/java/net/imagej/ops/transform/TransformNamespace.java
+++ b/src/main/java/net/imagej/ops/transform/TransformNamespace.java
@@ -87,7 +87,6 @@ import net.imglib2.RandomAccess;
 import net.imglib2.RandomAccessible;
 import net.imglib2.RandomAccessibleInterval;
 import net.imglib2.RealRandomAccessible;
-import net.imglib2.img.Img;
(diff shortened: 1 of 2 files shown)


stelfrich:
In general it looks good to me.

The call to `RAIs.extend()` was never returning the original input anyway, since the `outOfBoundsFactory` cannot be set to `null` by the user. That is, if the user requests a `ScaleView` with `null` as `outOfBoundsFactory`, Ops will replace the `outOfBoundsFactory` with the default one (as tested in [ScaleViewTest](https://github.com/imagej/imagej-ops/blob/master/src/test/java/net/imagej/ops/transform/scaleView/ScaleViewTest.java#L79-L80)). All in all, the ops overwrite the OOBF even if `input instanceof ExtendedRandomAccessibleInterval`.

dietzc:
Did you test the matching of the `DefaultScaleView`? Is it matching a `ExtendOp`?

stelfrich:
> Did you test the matching of the `DefaultScaleView`? Is it matching a `ExtendOp`?

Yes, matching works as expected.

Issue #463, opened 2016-11-14
Improve ConverterNamespace


dietzc:
* [] If used as a `Function`, all `Converter`s in the `ConvertNamespace`use `Map` to create the output. However, we should use `MapView' whenever we can (lazy).

* [] If used as a `Computer` the expected output has to be an `Img`. However, it just needs to be `IterableInterval`.

* [] Provide methods for customized conversion on `IterableInterval` or `RAI` level (e.g. more than just copy the values over), e.g. scale input values to destination type, normalize values and scale values to destination type, etc.

stelfrich:
> Provide methods for customized conversion on `IterableInterval` or `RAI` level (e.g. more than just copy the values over), e.g. scale input values to destination type, normalize values and scale values to destination type, etc. 

We already have some of that functionality in place. However, it feels awkward to use at the moment.

PR #466, opened 2016-11-18
Fix NormalizeScaleRealTypes

stelfrich:
Fixes a bug in `NormalizeScaleRealTypes` and addresses (to some extent) the last point of #463 by enabling:

```groovy
ops.op(Ops.Convert.NormalizeScale.class, out.firstElement(), in.firstElement()));
ops.run(Ops.Convert.ImageType.class, out, in, scaleOp);
```

Previously, matching of `Ops.Convert.NormalizeScale` was prohibited by a misleading contingency assumption.

@hornm, you are listed as the original author of these converters. Could you take a look if you find some time?


panovr:
I also test this fix and it works for me.

PR #468, opened 2016-11-21
Replace DefaultFloodFill

tibuch:
DefaultFloodFill implementation is replaced by the n-dimensional FloodFill implementation from imglib2. 

https://github.com/imagej/imagej-ops/issues/391
https://github.com/imagej/imagej-ops/pull/394

All tests are still passing.


dietzc:
thanks @tibuch. Why do you use different `Shape`s for test and as default (`Rectangle` and `Diamond`)? Rest looks good. Thanks!

tibuch:
Rectangle corresponds to 8-connected and diamond corresponds to 4-connected. 

As default 8-connected is used but in the tests 4-connected was used.

PR #469, opened 2016-11-22
Haralick tests

tibuch:
Tests added for the haralick features.

I had to change some formulas, but all references are included.

There was a bug in the scaling function (https://github.com/imagej/imagej-ops/commit/4e8d092e3d603fdfe3bc8f2776a66d49d91d5d08). 

If we take a look at the binning (8 buckets, min = 1, max = 254): 
binning

The new function (orange) creates 8 equal bins (last bin contains "one" element more, since the 254-elements are put into this one) and the old function (blue) creates 7 equal bins and a separate 8th bin for the 254-elements.  

Also the EPSILON which is needed to avoid log(0) is set to the Double.MIN_NORMAL value.


dietzc:
awesome. Thanks!!!

Issue #474, opened 2017-01-02
Incorrect results from integral image-based local thresholding methods


stelfrich:
There is a bug in the computation of squared integral images, which are required for methods that are based on variance/standard deviation.

This should have been spotted by unit tests that are comparing the default implementation of such methods with the integral image-based implementations. While such tests are available, the combination of parameters and out of bounds strategies lets those test pass.

PR #475, opened 2017-01-02
Fix LocalThresholdTest

stelfrich:
This PR fixes tests to fail in order to reproduce the issues described in #474.

### DIFF
--- a/src/test/java/net/imagej/ops/threshold/apply/LocalThresholdTest.java
+++ b/src/test/java/net/imagej/ops/threshold/apply/LocalThresholdTest.java
@@ -287,7 +287,7 @@ public class LocalThresholdTest extends AbstractOpTest {
 			new OutOfBoundsMirrorFactory<ByteType, Img<ByteType>>(Boundary.SINGLE),
 			1.0, Double.MAX_VALUE * 0.5);
 
-		assertEquals(out.firstElement().get(), true);
+		assertEquals(true, out.firstElement().get());
 	}
 
 	/**
@@ -298,7 +298,7 @@ public class LocalThresholdTest extends AbstractOpTest {
 		ops.run(LocalContrastThreshold.class, out, in, new RectangleShape(3, false),
 			new OutOfBoundsMirrorFactory<ByteType, Img<ByteType>>(Boundary.SINGLE));
 
-		assertEquals(out.firstElement().get(), false);
+		assertEquals(false, out.firstElement().get());
 	}
 
 	/**
@@ -309,7 +309,7 @@ public class LocalThresholdTest extends AbstractOpTest {
 		ops.run(Huang.class, out, in, new RectangleShape(1, false),
 			new OutOfBoundsMirrorFactory<ByteType, Img<ByteType>>(Boundary.SINGLE));
 
-		assertEquals(out.firstElement().get(), true);
+		assertEquals(true, out.firstElement().get());
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


src/test/java/net/imagej/ops/threshold/apply/LocalThresholdTest.java
```diff
@@ -769,4 +790,22 @@ public void testContingencyOfNormalImplementation() {
 		}
 	}
 
+	/**
+	 * Normalizes input image to [0.0, 1.0] range.
+	 */
+	private Img<DoubleType> normalizeImg(final Img<ByteType> input) {
```
dietzc: can't you use the `normalize` `Ops` here? if not, we should fix this?!

stelfrich: Good catch @dietzc! I have simplified the method in a recent commit.

The problem with `Ops.Image.Normalize` ops, however, is their restriction on the same type of input and output pixel types. Therefore, you still have to convert the input to the desired output type before applying those ops or am I missing something?

stelfrich: I'll investigate the bug and look into that idea afterwards (+ open an issue).

stelfrich:
> hm, couldn't we make the `normalize` less restrictive, for example by adding a super-class which consumes an output type?

@dietzc It is possible to use `OpService.run(Ops.Image.Normalize.class, ...)` which takes care of the conversion. So this restriction only applies to the methods in the `ImageNamespace`, e.g.:

```java
public <T extends RealType<T>> IterableInterval<T> normalize(
    final IterableInterval<T> out,
    final IterableInterval<T> in) {...}
```

dietzc:
I guess it's because we have a `Converter` from `T` to `V`. We've to check if this converter goes simply from `RealType` to `RealType` or if it only performs loss-less conversions (e.g. `Byte` to `Float` in contrast to `Float` to `Byte`).

ctrueden:
Note that the SciJava Converter framework (i.e.: not the ImgLib2 `Converter` interface) is slated for a redesign that will explicitly differentiate between lossy and lossless conversions. This will be highly useful for ops matching, among other things.

stelfrich:
Thanks for the input @dietzc and @ctrueden! I have opened an extra issue for the flexibility of the `ImageNamespace` (see #477), to keep track of the progress.

Do you have any major concerns regarding this PR?

src/test/java/net/imagej/ops/threshold/apply/LocalThresholdTest.java
```diff
@@ -287,7 +287,7 @@ public void testLocalBernsenThreshold() {
 			new OutOfBoundsMirrorFactory<ByteType, Img<ByteType>>(Boundary.SINGLE),
 			1.0, Double.MAX_VALUE * 0.5);
 
-		assertEquals(out.firstElement().get(), true);
+		assertEquals(true, out.firstElement().get());
```
ctrueden: Why not use `assertTrue`?

src/test/java/net/imagej/ops/threshold/apply/LocalThresholdTest.java
```diff
@@ -298,7 +298,7 @@ public void testLocalContrastThreshold() {
 		ops.run(LocalContrastThreshold.class, out, in, new RectangleShape(3, false),
 			new OutOfBoundsMirrorFactory<ByteType, Img<ByteType>>(Boundary.SINGLE));
 
-		assertEquals(out.firstElement().get(), false);
+		assertEquals(false, out.firstElement().get());
```
ctrueden: Why not use `assertFalse`?

ctrueden:
No concerns on my end. I am fine with merging as-is. I did ask one question about the asserts, but it's a nitpick. Not important enough to worry about before merging.

ctrueden:
Thanks for your efforts, @stelfrich!

src/test/java/net/imagej/ops/threshold/apply/LocalThresholdTest.java
stelfrich: That's a good point! I'll address that after the merge 👍

PR #476, opened 2017-01-02
Fix SquareIntegralImg

### LINKED ISSUE #474: Incorrect results from integral image-based local thresholding methods

stelfrich:
This PR fixes the computation of squared integral images: squaring of input values only needs to be done in one dimension; computations along other axes are just summations.

This makes the tests that are fixed in #475 pass again. Hence, this PR fixes #474 in combination with #475.

### DIFF
--- a/src/main/java/net/imagej/ops/image/integral/AbstractIntegralImg.java
+++ b/src/main/java/net/imagej/ops/image/integral/AbstractIntegralImg.java
@@ -61,22 +61,19 @@ public abstract class AbstractIntegralImg<I extends RealType<I>> extends
 	implements Contingent
 {
 
-	private AbstractUnaryHybridCI<IterableInterval<RealType<?>>, IterableInterval<RealType<?>>> integralAdd;
 	private UnaryComputerOp[] slicewiseOps;
 	private UnaryFunctionOp<Dimensions, RandomAccessibleInterval> createLongRAI;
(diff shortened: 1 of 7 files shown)


stelfrich:
We'll have to merge this one in addition to #475 to not have a failing build, @ctrueden. Don't know if this is on your radar as well..

ctrueden:
Drat. I did not know that #475 would break the build. This should have been done as a single PR, with all passing commits compiling with passing tests. This is always the SciJava policy: no commit should ever break compilation or break any tests.

ctrueden:
One way you could have achieved the desired stability would be to do commits in a sequence like:

1. Disable the tests which will break by the desired bug-fix.
2. The bug-fix itself.
3. Enable the updated tests which reflect the desired behavior.

Disabling the tests can be as simple as commenting out the `@Test` annotations.

stelfrich:
> This should have been done as a single PR, with all passing commits compiling with passing tests.

Sorry, my bad :-1: My reasoning was that the tests were too far downstream of the bugfix to be entangled in one PR. I'll keep the passing tests in mind for future PRs!

PR #480, opened 2017-01-18
Zernike tests

tibuch:
This PR is part of https://github.com/imagej/imagej-ops/issues/449.

The Zernike implementation puts the input-region on a disk with the smallest possible radius. Beforehand the radius was to big in the rectangular case. Also the number of pixel in the disk, which is used to normalize the moments, was wrong. 

Code cleanup removes "_" in front of variables and reduces computation steps.


dietzc:
thanks!

Issue #483, opened 2017-02-06
Convexity feature values should be between 0 and 1


imagejan:
As discussed on [gitter.im](https://gitter.im/imagej/imagej-ops?at=58989049567b0fc8138c7497), there are cases where measured convexity values are high above 1.0, where the theoretical range should be ` {0,1}` .

Here's a KNIME workflow that illustrates the issue with some segmented 3D objects:

[Convexity_Tests.zip](https://github.com/imagej/imagej-ops/files/755181/Convexity_Tests.zip)

tibuch:
@imagejan thank you for the test-workflow! I was able to identify the issue. The current implementation is based on surfaceArea(ConvexHull(Sample))/surfArea(Sample). Unfortunately the surface area of the convex hull of an object can be smaller or bigger than the surface area of the sample. 

This happens for example if the object is banana-shaped:
![surfacearea](https://cloud.githubusercontent.com/assets/5450869/22691072/faf87c72-ed38-11e6-9f91-5b5eae3e771a.png)

The correct convexity-measurement for 3D should be volume(convexHull(sample))/volume(sample). I found this paper about 3D convexity (https://www.researchgate.net/publication/236018239_A_New_Convexity_Measurement_for_3D_Meshes). They developed another 3D convexity measurement, because for certain objects the above measurement creates to high numbers (but still < 1). 

For now I will replace the convexity measurement with the volume-based implementation.

imagejan:
Isn't this the same as *Solidity* then?
See here:
https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/geom/AbstractSolidity.java#L61

In theory, the convex hull _surface_ (or _boundary_) should always be smaller than the object's _surface_, so the ratio is correct, but the calculation of the surfaces isn't. Do you agree?

tibuch:
Right, it is the same as _Solidity_.

> the convex hull surface (or boundary) should always be smaller than the object's surface

This is not true. If we have banana-shape with height 1 and compute the convex hull we get a half-circle with height 1. Now the surface area of a "half-circle" is bigger than the surface area of the banana. 
![banana](https://cloud.githubusercontent.com/assets/5450869/22694171/6d8bfea0-ed46-11e6-8f98-4977b008cdd6.png)

stelfrich:
> Now the surface area of a "half-circle" is bigger than the surface area of the banana.

I have never thought about that case, thanks for pointing it out @tibuch!

PR #484, opened 2017-02-07
Fix 3D Convexity measurement.

tibuch:
Fix for this issue https://github.com/imagej/imagej-ops/issues/483.

New:
3D convexity = volume/volumeConvexHull

Old:
3D convexity = surfaceArea(ConvexHull)/surfaceArea


src/test/java/net/imagej/ops/geom/MeshFeatureTests.java
```diff
@@ -98,7 +101,7 @@ public void convexHull3D() {
 	@Test
 	public void convexityMesh() {
 		// formula verified and ground truth computed with matlab
-		assertEquals(Ops.Geometric.Convexity.NAME, 0.983930494866521,
+		assertEquals(Ops.Geometric.Convexity.NAME, 0.84564860426,
```
imagejan: This is now exactly the same as ` Ops.Geometric.Solidity`.

tibuch:
Since the new _Convexity_ is equal to _Solidity_ this seems not to be the way to go.

As far as I can tell we have three options:
1. Keep _Convexity_ as it is. (Not mapping to [0, 1])
2. _Convexity_ == _Solidity_ (This PR)
3. Remove _Convexity_.

stelfrich:
IMHO, changing the 3D convexity implementation to emulate the 3D solidity implementation isn't a viable alternative since the transfer from 2D to 3D convexity seems obvious (boundary to surface). Isn't it a completely different story (interpretation in contrast to implementation), that this doesn't map to `[0, 1]`?

tibuch:
I agree with @stelfrich. What is your opinion @imagejan?

imagejan:
I agree, my comment above intended to say: it's wrong to change this calculation to something that's not backed by the theoretical definition.
But then: isn't the convexity still supposed to be between 0 and 1, from the mathematical term? Anything else is hard to interpret, IMHO.

I guess it all comes down to: I'd like to see a correct mathematical _definition_ of 3D Convexity, and this (as well as its correct _interpretation_) should be clearly documented - in the javadoc, and (ideally automatically) as well in downstream applications, e.g. in the node descriptions of the Feature Calculator in KNIME.

ctrueden:
@tibuch Was a consensus reached on what to do about this?

imagejan:
I think we can stick with option 1:

> 1. Keep Convexity as it is. (Not mapping to [0, 1])

and close this PR.

As mentioned in #483, I just found it counter-intuitive that convexity values can be higher than 1, so I still think this should be well documented somewhere.

ctrueden:
Thanks, everyone.

PR #494, opened 2017-03-30
Add BoxCount Op

rimadoma:
Add a BoxCounting Op that can be used to find out fractal dimension of an image. The implementation is n-dimensional. See https://en.wikipedia.org/wiki/Box_counting.


ctrueden:
Thank you!

PR #496, opened 2017-04-03
Add Outline op

rimadoma:
Add an op that copies the outlines of objects into output. It creates "hollow" versions of objects with only the surface voxels. It works in n-dimensions.


imagejan:
How about naming this `outline`, in line with ImageJ 1.x's [*Process > Binary > Outline*](https://github.com/imagej/imagej1/blob/89f010bed96875904bc06cbc5e641815adbdad97/ij/plugin/filter/Binary.java#L152-L156) or e.g. the current [KNIME implementation](https://github.com/knime-ip/knip/blob/8aa1488e09dafcbdfca903fd9ed4ed67d24e5efd/org.knime.knip.core/src/org/knime/knip/core/features/seg/ExtractOutlineImg.java)?

rimadoma:
Op renamed to `Outline`.

ctrueden:
Thanks @rimadoma!!

PR #497, opened 2017-04-12
Fix BoxCount Op output

rimadoma:
Invert point values to follow the normal (x,y) convention. Sorry for the extra bother.


ctrueden:
:+1:

PR #499, opened 2017-05-03
Don't ignore OutOfBorderFactory when calculating difference of gaussian

maarzt:
Previously the given OutOfBorderFactory was not used for the calculation of the gaussian filters.


ctrueden:
Thanks!

ctrueden:
Best practice is to add a regression test which fails without this bug-fix, but passes with it in place.

PR #504, opened 2017-06-01
Gauss implementation for non-interval RA input

axtimwalde:
* added Gauss implementation for non-interval RA input
* reduced type specifity of Gauss Ops to NumericType

### DIFF
--- a/src/main/java/net/imagej/ops/create/CreateNamespace.java
+++ b/src/main/java/net/imagej/ops/create/CreateNamespace.java
@@ -109,7 +109,7 @@ public class CreateNamespace extends AbstractNamespace {
 	{
 		@SuppressWarnings("unchecked")
 		final Img<T> result = (Img<T>) ops().run(
-			net.imagej.ops.Ops.Create.Img.class, in1, in2);
+			Ops.Create.Img.class, in1, in2);
 		return result;
 	}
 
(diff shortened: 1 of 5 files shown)


ctrueden:
@axtimwalde Are you still working on this? Or are you looking for assistance on getting the build to pass? What is the status?

axtimwalde:
@ctrueden the test fails because of a, to me, mysterious coverage error.  I asked some questions but did not understand yet what I am expected to do.  The test asks me to implement a method that does not make sense: one that creates an output although the implemented Op is a C, not a CF.

ctrueden:
Extending `AbstractUnaryHybridC` class directly makes no sense—that abstract class exists only as a common layer for `AbstractUnaryHybridCI` and `AbstractUnaryHybridCF`. If your goal was for `DefaultGaussRA` to simply be a computer, then change it to extend `AbstractUnaryComputer` instead, and the tests will pass.

axtimwalde:
@ctrueden the tests are finally passing, if you have no further objections, it would be great if this could be merged (or better rebased and squashed I guess).

ctrueden:
I cleaned up the commits, but haven't had time to 100% finalize it yet. Will do so as soon as I can. In the meantime, let me know if you don't like what you see.

axtimwalde:
@ctrueden thanks!  I do not have preferences regarding style.  Most of what you changed was there before this pull request and I tried to comply.

ctrueden:
It is done!

axtimwalde:
Thanks @ctrueden!

PR #511, opened 2017-06-23
Fix Map Op Bugs with Cursors

awalter17:
Hello!

These changes fix a couple of issues I noticed with the Map Ops.
* Fixes `Maps` methods to always call the cursor first and then compute the result. This avoids bounds issues since different cursors have different bound behaviors ([see](https://github.com/imglib/imglib2/issues/170) for more details)
* Fixes `MapIIandIIInplace` `mutate2(...)` to mutate the second operand instead of the first

I've also added tests for these fixes. The tests are using `CellImg`s instead of `ArrayImg`s, which have different cursor bound behaviors. In general, the only affected methods were regarding parallel ops so I only added tests with `CellImgs` for those specific operations. Additionally, I added one test to `ConvertImagesTest` since I initially encountered this problem trying to convert between `CellImg` types. Only one test was needed since they all call the same `map(...)` method.

All commits compile with passing tests. Please feel free to let me know if you have any questions or if any changes are needed.

Thanks, 

Alison

### DIFF
--- a/src/main/java/net/imagej/ops/map/MapIIAndIIInplace.java
+++ b/src/main/java/net/imagej/ops/map/MapIIAndIIInplace.java
@@ -65,6 +65,6 @@ public class MapIIAndIIInplace<EA> extends
 	public void mutate2(final IterableInterval<EA> in,
 		final IterableInterval<EA> arg)
 	{
-		Maps.inplace(arg, in, getOp());
+		Maps.inplace(in, arg, getOp());
 	}
 }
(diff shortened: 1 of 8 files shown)


awalter17:
@ctrueden Per our discussion I separated out the clean ups into separate commits, added the `numSteps <= 0` checks, and replaced the `while` loops with `for` loops. 

I also checked if there was any significant time difference between calling the cursors outside the loop once or having a check inside the loop, there was no difference. So 5543821 adds the check inside the loop. If you like this change you can squash it into 1f2c4b9, otherwise it can be deleted.

Please feel free to let me know if any additional changes are needed.

ctrueden:
Thanks very much @awalter17! I like 5543821185900400c38f5ade63877536388ec315. And I want to leave it separated, to document why we are doing it that way.

PR #512, opened 2017-06-28
UnaryComputerOp: make compute arguments final

frauzufall:
I wrote an `AbstractUnaryHybridCF` class that always returned zero as output because I could do something like this: 
```
@Override
public void compute(RandomAccessibleInterval<T> input, T output) {
	// ..
	output = value;
}
```
.. instead of this:
```
@Override
public void compute(final RandomAccessibleInterval<T> input, final T output) {
	// ..
	output.set(value);
}
```
The non-final version was automatically created based on `UnaryComputerOp` so I think this should be a solution? `mvn test` did not return any errors.

### DIFF
--- a/src/main/java/net/imagej/ops/special/computer/UnaryComputerOp.java
+++ b/src/main/java/net/imagej/ops/special/computer/UnaryComputerOp.java
@@ -62,7 +62,7 @@ public interface UnaryComputerOp<I, O> extends UnaryOp<I, O>,
 	 * @param output Object where the computation's result will be stored, which
 	 * <em>must be non-null and a different object than {@code input}</em>
 	 */
-	void compute(I input, O output);
+	void compute(final I input, final O output);
 
 	// -- UnaryOp methods --
 


rimadoma:
Hmm, making `output` final in the `UnaryComputerOp` does make sense, but in an HybridCF it shouldn't be `final` since it can generate the `output` when acting as a function.

ctrueden:
In Java, putting `final` on an interface method has no effect. The compiler does not enforce this on downstream implementors. So by convention, we always leave it out.

Here is an example, which compiles successfully:
```java
public class FinalOverride {
	
	public static interface Foo {
		void foo(final int x);
	}
	
	public static class Fubar implements Foo {
		@Override
		public void foo(int x) {
			x *= 2;
			System.out.println("x = " + x);
		}
	}

	public static void main(final String... args) throws Exception {
		new Fubar().foo(10);
	}
}
```

ctrueden:
As a general rule, I strive for all method parameters of all methods I write to be final. I actually wish that Java variables were final by default, and you had to write `var` for them not to be. This would be shorter 99% of the time. But, that's not how it is, so... just gotta get used to it. One thing that helps a lot is the "Clean Up" function of Eclipse, which you can configure to always add final keywords everywhere it would still compile to do so.

frauzufall:
Thanks for explaining! I get your point and I also recognized that the compiler did not enforce it, but using the Eclipse feature to automatically create the Override functions would then have included `final` function arguments (I guess?) and would have saved me a few hours of searching for the reason of the zero output value. Now that I know the error it is quite obvious and will not happen to me again, I was just wondering if I am perhaps not the only one doing that mistake.

ctrueden:
> but using the Eclipse feature to automatically create the Override functions would then have included final function arguments (I guess?)

Nope, I tested it. :wink:

ctrueden:
It literally does not make it into the bytecode of the interface:
```
$ cat X.java
    1 public interface X {
    2   public void go(final int x);
    3 }
$ javap -cp . X
Compiled from "X.java"
public interface X {
  public abstract void go(int);
}
$ cat Y.java
    1 public class Y {
    2   public void go(final int x) {
    3   }
    4 }
$ javap -cp . Y
Compiled from "Y.java"
public class Y {
  public Y();
  public void go(int);
}
```

ctrueden:
> I was just wondering if I am perhaps not the only one doing that mistake.

I just want to clarify that I really appreciate your effort, even if it unfortunately does not help people in this specific case. Other ideas to make things easier for people are of course extremely welcome! :-)

frauzufall:
There is also a discussion on [stackoverflow](https://stackoverflow.com/questions/5380177/final-arguments-in-interface-methods-whats-the-point).. No worries, I totally get your point!

PR #513, opened 2017-06-29
Fix Bug in CachedOpEnvironment Hash Code Calculation

maarzt:
I couldn't use `CachedOpEnvironment` to cache the gauss filter op, because it has an optional parameter, which I left unset. This resulted in a `NullPointerException` when calculating the hash for the optional parameter.


ctrueden:
Thanks @maarzt!

PR #515, opened 2017-07-07
Interval functionality for various Transform Ops

### LINKED ISSUE #486: Wrap Views signatures that return `Interval`
For both `rotate` and `hyperSlice`, there are two sigs: one for RA and one for RAI. We only wrap the RA one right now, and hence when you call e.g. `ops.run("rotateView", image)` the result is not an `Interval` anymore. We should wrap the RAI versions and mark them higher priority, to avoid this.

gselzer:
This branch solves issue #486 by adding Interval functionality for the Rotate and Hyperslice ops. The branch also adds Interval functionality for the Permute, InvertAxis, Subsample and Translate ops.

This branch also fixes a Transform Namespace bug. Previously, the Transform Namespace was not being tested correctly because it was looking for ops in the "View" namespace, which does not exist.

Finally this branch removes the DefaultScaleView and renames WrappedScaleView as the new DefaultScaleView. This was done due to the redundancy between the two ops and due to the fact that DefaultScaleView would never be called.

Closes #486.


src/test/java/net/imagej/ops/AbstractOpTest.java
```diff
@@ -235,6 +238,10 @@ private int pseudoRandom() {
 		}
 		assertFalse("More elements than expected", a.hasNext());
 	}
+	
+	public static <T> RandomAccessible<T> deinterval(RandomAccessibleInterval<T> input){
+		return Views.extendBorder(input);
```
stelfrich: Looking at where this method is used, replacing the original casts with calls to it seems legit. In general, however, this is not the case: with L243 you are replacing an already existing `OutOfBoundsFactory` of input with an `OutOfBoundsBorderFactory`.

ctrueden: @stelfrich How would you suggest it be changed?

stelfrich: @ctrueden No changes, since it seems to be the only alternative to replacing the casts.

src/test/java/net/imagej/ops/transform/rotateView/RotateViewTest.java
```diff
@@ -50,21 +53,60 @@
...
-	public void defaultRotateTest() {
-		Img<DoubleType> img = new ArrayImgFactory<DoubleType>().create(new int[] { 20, 10 }, new DoubleType());
-		
-		MixedTransformView<DoubleType> il2 = Views.rotate((RandomAccessible<DoubleType>) img, 1, 0);
-		MixedTransformView<DoubleType> opr = ops.transform().rotateView((RandomAccessible<DoubleType>) img, 1, 0);
-		
+	public void testDefaultRotate() {
+		final Img<DoubleType> img = new ArrayImgFactory<DoubleType>().create(new int[] { 20, 10 }, new DoubleType());
+
+		final MixedTransformView<DoubleType> il2 = Views.rotate((RandomAccessible<DoubleType>) img, 1, 0);
```
stelfrich: Also use `deinterval`? (applies to other tests as well...)

PR #517, opened 2017-07-18
Improve invert ii

gselzer:
This PR improves invert II to work for all supported `RealTypes` within ImageJ, as well as heterogeneous type conversion. The op has been tested with all data types and imaginable data values to ensure that it can work with whatever is needed. The fix contains two ops, one for smaller types (i.e. `DoubleType` and smaller) and the other for the larger types. The ops can be called with either a minimum and maximum, and the op will invert the data around the mean of the two numbers, or it can be called with no given minimum and maximum, where the data will be inverted around the center of the input image data type range.

This commit is ready for review but cannot be merged because it depends on a snapshot of ImgLib2. The dependency was changed due to a miscast `ShortType` operation and as such it cannot be merged to master until a new version of ImgLib2 is released.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>org.scijava</groupId>
 		<artifactId>pom-scijava</artifactId>
-		<version>18.1.0</version>
+		<version>19.1.1</version>
 		<relativePath />
 	</parent>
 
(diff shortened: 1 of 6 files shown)


gselzer:
@ctrueden all commits build and all tests pass on each commit

ctrueden:
Thanks @gselzer! :+1:

Issue #500, opened 2017-05-10
Add ConcatenateView op


hanslovsky:
I implemented a `ConcatenateView` op that allows to concatenate a list of `RandomAccessibleInterval` along the specified dimension, following what is in `imglib2` `Views`:
https://github.com/imagej/imagej-ops/compare/imagej:eada2b8...hanslovsky:b4df842

The pull request will have to wait until `imglib2-4.1.0` is released. I created this issue to ensure I won't forget about it and so others know the code exists and avoid code duplication.

ctrueden:
@hanslovsky If you don't mind, I'll have @gselzer take care of filing this as a PR.

hanslovsky:
@ctrueden @gselzer I do not mind so feel free to go ahead.

ctrueden:
Resolved with the merge of #518.

PR #518, opened 2017-07-18
Add Views.concatenate

gselzer:
This PR adds the ConcatenateView op (see issue #500) and tests for its functionality.

This PR also bumps the parent pom to version 16.2.0 to use necessary concatenate operations from ImgLib2.


ctrueden:
Rock and roll! :beers:

Issue #488, opened 2017-02-15
Bilateral filter


ctrueden:
As suggested by @imagejan, let's add a [bilateral filter](https://en.wikipedia.org/wiki/Bilateral_filter) by porting VIB-lib's [BilateralFilter implementation](https://github.com/fiji/VIB-lib/blob/master/src/main/java/vib/BilateralFilter.java).

ctrueden:
See also the [KNIME Image Processing `BilateralFilter` implementation](https://github.com/knime-ip/knip/blob/master/org.knime.knip.core/src/org/knime/knip/core/ops/filters/BilateralFilter.java) which is closer to what we need, in that it uses ImgLib2. But it uses the old `net.imglib2.ops` framework, which has been replaced by `net.imagej.ops`. We want to:

1. Port the linked code from VIB-lib and/or KNIME-IP to ImageJ Ops.
2. Ensure performance matches that of ImageJ—i.e., the VIB-lib implementation, available in ImageJ as the _Plugins>Process>Bilateral Filter_ command.
3. Ensure the results are correct. Double check the math as described by the Wikipedia article linked above.

PR #519, opened 2017-07-25
Bilateral Filter op

gselzer:
This PR adds a `BilateralFilter` op. The op takes a source input image, a source output image, a spatial smoothing parameter `m_SigmaS`, a range (intensity) parameter `m_SigmaR`, and a radius parameter `m_radius` used by the op to determine the range of pixels used for each individual filter operation. These last three parameters could potentially be set in the code and not alterable by the user, however now they are passed through with the op call. This decision was made as depending on the size of the image and the reasons the user is using the filter different radii and sigmas might be useful. As of now the filter does have its own Gaussian operation, taken from KNIME's implementation of the filter. I did look into using ImageJ's Gauss op for the filter however with the variant version of the Gaussian operation used for the Bilateral filter I do not think ImageJ's Gauss op will work. I would be happy to implement it in the Bilateral Filter if it is determined that it will work.

Closes issue #488.

### DIFF
--- a/src/main/java/net/imagej/ops/filter/FilterNamespace.java
+++ b/src/main/java/net/imagej/ops/filter/FilterNamespace.java
@@ -133,6 +133,28 @@ public class FilterNamespace extends AbstractNamespace {
 		return result;
 	}
 
+	/**
+	 * Executes a bilateral filter on the given arguments.
+	 * 
+	 * @param in
+	 * @param out
+	 * @param m_SigmaR
+	 * @param m_SigmaS
+	 * @param m_radius
+	 * @return
+	 */
+	@OpMethod(op = net.imagej.ops.filter.bilateral.DefaultBilateral.class)
+	public <I extends RealType<I>, O extends RealType<O>> RandomAccessibleInterval<O>
+		bilateral(final RandomAccessibleInterval<O> out,
+			final RandomAccessibleInterval<I> in, final double m_SigmaR,
+			final double m_SigmaS, final int m_radius)
+	{
+		@SuppressWarnings("unchecked")
+		final RandomAccessibleInterval<O> result = (RandomAccessibleInterval<O>) ops().run(
+			Ops.Filter.Bilateral.class, out, in, m_SigmaR, m_SigmaS, m_radius);
+		return result;
+	}
+
 	// -- convolve --
 	/** Executes the "convolve" operation on the given arguments. */
 	@OpMethod(ops = { net.imagej.ops.filter.convolve.ConvolveFFTF.class,
--- a/src/main/java/net/imagej/ops/filter/FilterNamespace.java
+++ b/src/main/java/net/imagej/ops/filter/FilterNamespace.java
@@ -138,20 +138,20 @@ public class FilterNamespace extends AbstractNamespace {
 	 * 
 	 * @param in
 	 * @param out
-	 * @param m_SigmaR
-	 * @param m_SigmaS
-	 * @param m_radius
+	 * @param sigmaR
+	 * @param sigmaS
+	 * @param radius
 	 * @return
 	 */
 	@OpMethod(op = net.imagej.ops.filter.bilateral.DefaultBilateral.class)
 	public <I extends RealType<I>, O extends RealType<O>> RandomAccessibleInterval<O>
 		bilateral(final RandomAccessibleInterval<O> out,
-			final RandomAccessibleInterval<I> in, final double m_SigmaR,
-			final double m_SigmaS, final int m_radius)
+			final RandomAccessibleInterval<I> in, final double sigmaR,
+			final double sigmaS, final int radius)
 	{
 		@SuppressWarnings("unchecked")
 		final RandomAccessibleInterval<O> result = (RandomAccessibleInterval<O>) ops().run(
-			Ops.Filter.Bilateral.class, out, in, m_SigmaR, m_SigmaS, m_radius);
+			Ops.Filter.Bilateral.class, out, in, sigmaR, sigmaS, radius);
 		return result;
 	}
 
--- a/src/main/java/net/imagej/ops/filter/bilateral/DefaultBilateral.java
+++ b/src/main/java/net/imagej/ops/filter/bilateral/DefaultBilateral.java
@@ -42,6 +42,7 @@ import net.imglib2.algorithm.neighborhood.Neighborhood;
 import net.imglib2.algorithm.neighborhood.RectangleNeighborhood;
 import net.imglib2.algorithm.neighborhood.RectangleNeighborhoodFactory;
 import net.imglib2.type.numeric.RealType;
+import net.imglib2.util.Intervals;
 import net.imglib2.view.Views;
 
 import java.util.Arrays;
@@ -162,11 +163,6 @@ public class DefaultBilateral<I extends RealType<I>, O extends RealType<O>>
 
 	@Override
 	public boolean conforms() {
-		long dimsIn[] = new long[in().numDimensions()];
-		in().dimensions(dimsIn);
-		long dimsOut[] = new long[out().numDimensions()];
-		out().dimensions(dimsOut);
-		
-		return (in().numDimensions() == 2 && Arrays.equals(dimsIn, dimsOut));
+		return (in().numDimensions() == 2 && Intervals.equalDimensions(in(), out()));
 	}
 }
--- a/src/main/java/net/imagej/ops/filter/bilateral/DefaultBilateral.java
+++ b/src/main/java/net/imagej/ops/filter/bilateral/DefaultBilateral.java
@@ -48,7 +48,12 @@ import org.scijava.plugin.Parameter;
 import org.scijava.plugin.Plugin;
 
 /**
- * Performs a bilateral filter on an image
+ * Performs a bilateral filter on an image. The parameter m_SigmaS refers to the
+ * spatial smoothing parameter; the greater the sigma, the smoother the image.
+ * The parameter m_SigmaR refers to the range smoothing parameter. The greater
+ * the sigma, the greater the effect of intensity differences. The parameter
+ * m_radius refers to the square that is considered when doing the filter on
+ * each individual picture.
  *
  * @author Gabe Selzer
  * @param <I>
--- a/src/main/java/net/imagej/ops/filter/bilateral/DefaultBilateral.java
+++ b/src/main/java/net/imagej/ops/filter/bilateral/DefaultBilateral.java
@@ -30,6 +30,7 @@
 
 package net.imagej.ops.filter.bilateral;
 
+import net.imagej.ops.Contingent;
 import net.imagej.ops.Ops;
 import net.imagej.ops.special.computer.AbstractUnaryComputerOp;
 import net.imglib2.Cursor;
@@ -59,7 +60,7 @@ import org.scijava.plugin.Plugin;
 public class DefaultBilateral<I extends RealType<I>, O extends RealType<O>>
 	extends
 	AbstractUnaryComputerOp<RandomAccessibleInterval<I>, RandomAccessibleInterval<O>>
-	implements Ops.Filter.Bilateral
+	implements Ops.Filter.Bilateral, Contingent
 {
 
 	public final static int MIN_DIMS = 2;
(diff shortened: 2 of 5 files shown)


src/main/java/net/imagej/ops/filter/FilterNamespace.java
```diff
@@ -133,6 +133,28 @@
...
 	}
 
+	/**
+	 * Executes a bilateral filter on the given arguments.
+	 * 
+	 * @param in
+	 * @param out
+	 * @param m_SigmaR
+	 * @param m_SigmaS
+	 * @param m_radius
```
imagejan: Is there a special reason you use `m_` prefix in the names? I've seen this in KNIME for `private` variables, but nowhere in the ImageJ universe, so I'd assume this doesn't conform to imagej-ops code style. Also, I'd prefer simple names such as `sigmaR`, `sigmaS` and `radius`.

src/main/java/net/imagej/ops/filter/bilateral/DefaultBilateral.java
```diff
@@ -0,0 +1,148 @@
...
+	AbstractUnaryComputerOp<RandomAccessibleInterval<I>, RandomAccessibleInterval<O>>
+	implements Ops.Filter.Bilateral
+{
+
+	public final static int MIN_DIMS = 2;
+
+	public final static int MAX_DIMS = 2;
+
+	@Parameter
+	private double m_sigmaR;
```
imagejan: Same as above, if I read `m_` I'd assume entirely `private` variable, but these here are set via reflection due to the `@Parameter` syntax.

gselzer: Whoops! That completely slipped my mind! I just went through and changed them to fit ImageJ standard.

src/main/java/net/imagej/ops/filter/bilateral/DefaultBilateral.java
```diff
@@ -48,12 +48,12 @@
...
 
 /**
- * Performs a bilateral filter on an image. The parameter m_SigmaS refers to the
+ * Performs a bilateral filter on an image. The sigmaS parameter refers to the
  * spatial smoothing parameter; the greater the sigma, the smoother the image.
- * The parameter m_SigmaR refers to the range smoothing parameter. The greater
- * the sigma, the greater the effect of intensity differences. The parameter
- * m_radius refers to the square that is considered when doing the filter on
- * each individual picture.
+ * The sigmaR parameter refers to the range smoothing parameter. The greater the
```
ctrueden: Instead of documenting parameters in the main section like this, you can put javadoc on the actual parameters below. This keeps the descriptions associated directly with the params, such that if a param is changed or removed later, the documentation will be less likely to be overlooked.

src/main/java/net/imagej/ops/filter/bilateral/DefaultBilateral.java
ctrueden: In general (not just in ImageJ), using `m_` in Java is wrong/bad for several reasons. Joel Spolsky has a [great article discussing Hungarian notation](https://www.joelonsoftware.com/2005/05/11/making-wrong-code-look-wrong/), including both App Hungarian and Systems Hungarian.

src/main/java/net/imagej/ops/filter/bilateral/DefaultBilateral.java
```diff
@@ -0,0 +1,148 @@
...
+		final double mu = 0.0;
+		return (1 / (sigma * Math.sqrt(2 * Math.PI))) * Math.exp((-0.5 * (x - mu) *
+			(x - mu)) / (sigma * sigma));
+	}
+
+	@Override
+	public void compute(final RandomAccessibleInterval<I> input,
+		final RandomAccessibleInterval<O> output)
+	{
+		if (input.numDimensions() != 2) {
```
tibuch: Use the `Contingent` interface to check input dims. The `Op` will fail during matching and not during execution.

src/main/java/net/imagej/ops/filter/bilateral/DefaultBilateral.java
```diff
@@ -0,0 +1,148 @@
...
+			cp.localize(p);
+			double d;
+			cp.localize(mi);
+			cp.localize(ma);
+			mi[0] = Math.max(0, mi[0] - radius);
+			mi[1] = Math.max(0, mi[1] - radius);
+			ma[0] = Math.min(mma1, ma[0] + radius);
+			ma[1] = Math.min(mma2, ma[1] + radius);
+			final Interval in = new FinalInterval(mi, ma);
+			final RectangleNeighborhoodFactory<I> fac = RectangleNeighborhood
```
tibuch: This factory can be reused in every iteration.

src/main/java/net/imagej/ops/filter/bilateral/DefaultBilateral.java
```diff
@@ -0,0 +1,148 @@
...
+			cp.fwd();
+			cp.localize(p);
+			double d;
+			cp.localize(mi);
+			cp.localize(ma);
+			mi[0] = Math.max(0, mi[0] - radius);
+			mi[1] = Math.max(0, mi[1] - radius);
+			ma[0] = Math.min(mma1, ma[0] + radius);
+			ma[1] = Math.min(mma2, ma[1] + radius);
+			final Interval in = new FinalInterval(mi, ma);
```
stelfrich: You could use available functionality from [`AbstractCenterAwareNeighborhoodBasedFilter`](https://github.com/imagej/imagej-ops/blob/ae691c5f38542cf3e94b88f33a1ad235b754777c/src/main/java/net/imagej/ops/filter/AbstractCenterAwareNeighborhoodBasedFilter.java) which maps a [`CenterAwareComputerOp`](https://github.com/imagej/imagej-ops/blob/ae691c5f38542cf3e94b88f33a1ad235b754777c/src/main/java/net/imagej/ops/map/neighborhood/CenterAwareComputerOp.java) onto each neighborhood of an image. That way, you would not have to take care of the neighborhood creation yourself but would get that for free. The [`DefaultSigmaFilter`](https://github.com/imagej/imagej-ops/blob/ae691c5f38542cf3e94b88f33a1ad235b754777c/src/main/java/net/imagej/ops/filter/sigma/DefaultSigmaFilter.java) implements that scheme. Although I am not entirely sure if it will work in your case, because the [CenterAwareComputerOp](https://github.com/imagej/imagej-ops/blob/ae691c5f38542cf3e94b88f33a1ad235b754777c/src/main/java/net/imagej/ops/map/neighborhood/CenterAwareComputerOp.java) implements `BinaryComputerOp<Iterable<I>, I, O>`, s.t. you will not be able to get the position in the neighborhood...

src/main/java/net/imagej/ops/filter/bilateral/DefaultBilateral.java
```diff
@@ -0,0 +1,148 @@
...
+			final RectangleNeighborhoodFactory<I> fac = RectangleNeighborhood
+				.factory();
+			rn = fac.create(p, mi, ma, in, input.randomAccess());
+			cq = rn.localizingCursor();
+			double s, v = 0.0;
+			double w = 0.0;
+			do {
+				cq.fwd();
+				cq.localize(q);
+				d = ((p[0] - q[0] - mi[0]) * (p[0] - q[0] - mi[0])) + ((p[1] - q[1] -
```
stelfrich: You could refactor this into distance function that also works with _n_-dimensional inputs

src/main/java/net/imagej/ops/filter/bilateral/DefaultBilateral.java
```diff
@@ -0,0 +1,148 @@
...
+			throw new IllegalArgumentException("Input must be two dimensional");
+		}
+
+		final long[] size = new long[input.numDimensions()];
+		input.dimensions(size);
+
+		final RandomAccess<O> cr = output.randomAccess();
+		final Cursor<I> cp = Views.iterable(input).localizingCursor();
+		final long[] p = new long[input.numDimensions()];
+		final int[] q = new int[input.numDimensions()];
```
stelfrich: Is there a reason for `q` to be an `int[]`?

src/main/java/net/imagej/ops/filter/bilateral/DefaultBilateral.java
```diff
@@ -0,0 +1,148 @@
...
+		final RandomAccessibleInterval<O> output)
+	{
+		if (input.numDimensions() != 2) {
+			throw new IllegalArgumentException("Input must be two dimensional");
+		}
+
+		final long[] size = new long[input.numDimensions()];
+		input.dimensions(size);
+
+		final RandomAccess<O> cr = output.randomAccess();
```
stelfrich: Variables could be a little more descriptive, especially since there are so many similar ones in the implementation.

src/test/java/net/imagej/ops/AbstractOpTest.java
```diff
@@ -236,6 +240,18 @@ private int pseudoRandom() {
 		assertFalse("More elements than expected", a.hasNext());
 	}
 
+	public <T extends RealType<T>> boolean areCongruent(final IterableInterval<T> in, final RandomAccessible<T> out){
```
stelfrich: That method would be super useful with an additional `epsilon` parameter to define the precision of the equality check.

src/test/java/net/imagej/ops/filter/DefaultBilateralTest.java
```diff
@@ -0,0 +1,136 @@
...
+import net.imagej.ops.filter.bilateral.DefaultBilateral;
+import net.imagej.ops.filter.gauss.GaussRAISingleSigma;
+
+import org.junit.Test;
+
+public class DefaultBilateralTest extends AbstractOpTest{
+	
+	@Test
+	public void testBigImage(){
+		final Img<ByteType> in = generateByteArrayTestImg(false, 6, 6);
```
stelfrich: You can replace L19-L26 (and similar instances of that pattern) with

```
final byte[] data = {...};
final Img<ByteType> in = ArrayImgs.bytes(data, 6, 6);
```

src/main/java/net/imagej/ops/filter/bilateral/DefaultBilateral.java
gselzer: Should be fixed!

src/main/java/net/imagej/ops/filter/bilateral/DefaultBilateral.java
gselzer: I didn't know this existed! Thanks! I just implemented it in the filter and added a test to make sure it works.

src/test/java/net/imagej/ops/AbstractOpTest.java
```diff
@@ -240,14 +240,14 @@ private int pseudoRandom() {
...
-	public <T extends RealType<T>> boolean areCongruent(final IterableInterval<T> in, final RandomAccessible<T> out){
+	public <T extends RealType<T>> boolean areCongruent(final IterableInterval<T> in, final RandomAccessible<T> out, final double epsilon){
 		Cursor<T> cin = in.localizingCursor();
 		RandomAccess<T> raOut = out.randomAccess();
 		
 		while(cin.hasNext()){
 			cin.fwd();
 			raOut.setPosition(cin);
-			if(cin.get().getRealDouble() != raOut.get().getRealDouble()) return false;
+			if((cin.get().getRealDouble() - raOut.get().getRealDouble()) > epsilon) return false;
```
stelfrich: You will have to check the absolute difference to support the case where `cin.get().getRealDouble()` is less than `raOut.get().getRealDouble()` assuming that epsilon will always be positive which is what I would suggest.

src/main/java/net/imagej/ops/filter/bilateral/DefaultBilateral.java
```diff
@@ -0,0 +1,148 @@
...
+	@Parameter
+	private double sigmaR;
+
+	@Parameter
+	private double sigmaS;
+
+	@Parameter
+	private int radius;
+
+	private static double gauss(final double x, final double sigma) {
```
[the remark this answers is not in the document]
ctrueden: @stelfrich I discussed this idea with @gselzer. We were wondering if you have any use cases or scenarios where being able to override these functions is useful. We can generalize the functions, yeah, but the resulting general thing would not really be a "bilateral" filter anymore per se, would it? So then: what do we name this more general op/method?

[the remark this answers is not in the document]
stelfrich: @ctrueden I was under the impression that bilateral filters define a class of filters of which most implementations use Gaussians to smooth the intensity difference and the spatial distance. Anyway, I am completely fine with how it is now, considering that we can generalize the op if there is demand for it (which I doubt) :+1:

PR #523, opened 2017-09-25
Fix the DefaulDifferenceVariance implementation (Haralick).

tibuch:
Thanks to Ilan Tal from the ImageJ-Forum for pointing it out.

See forum post: http://forum.imagej.net/t/radiomics-contrast-and-difference-variance/7137/4

### DIFF
--- a/src/main/java/net/imagej/ops/features/haralick/DefaultDifferenceVariance.java
+++ b/src/main/java/net/imagej/ops/features/haralick/DefaultDifferenceVariance.java
@@ -47,32 +47,37 @@ import org.scijava.plugin.Plugin;
  * @author Christian Dietz (University of Konstanz)
  * @author Tim-Oliver Buchholz (University of Konstanz)
  *
- * Formula based on http://murphylab.web.cmu.edu/publications/boland/boland_node26.html.
+ * Formula based on: http://haralick.org/journals/TexturalFeatures.pdf
  */
 @Plugin(type = Ops.Haralick.DifferenceVariance.class, label = "Haralick: Difference Variance")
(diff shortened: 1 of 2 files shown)


src/main/java/net/imagej/ops/features/haralick/DefaultDifferenceVariance.java
```diff
@@ -47,32 +47,37 @@
...
+
+		for (int i = 0; i < numGreyLevels; i++) {
+			m += i * pxminusy[i];
+		}
+
 		double sum = 0.0d;
-		for (int k = 0; k < pxminusy.length; k++) {
-			sum += k * k * pxminusy[k];
+		for (int k = 0; k < numGreyLevels; k++) {
+			sum += Math.pow(k - m, 2) * pxminusy[k];
```
ctrueden: FWIW, I benchmarked `Math.pow(k - m, 2)` vs `(k-m) * (k-m)` and there is no difference in performance.

src/main/java/net/imagej/ops/features/haralick/DefaultDifferenceVariance.java
```diff
@@ -47,32 +47,37 @@
...
 	}
-	
+
 	@Override
 	public void compute(final IterableInterval<T> input, final DoubleType output) {
 		final double[][] matrix = getCooccurrenceMatrix(input);
-
 		final double[] pxminusy = coocPXMinusYFunc.calculate(matrix);
+
+		double m = 0.0;
```
stelfrich: Could we call it `mu` or maybe `mean` to make the code a little more readable?

ctrueden:
Thank you!

Issue #490, opened 2017-02-24
Frangi Vesselness filter


stelfrich:
Since @mhl has asked that someone took over his Frangi Vesselness filter implementation ([fiji/Feature_Detection](https://github.com/fiji/Feature_Detection)) we could refactor it to an op.

He has also raised some concern about the correctness in comparison with a [MATLAB implementation](http://de.mathworks.com/matlabcentral/fileexchange/24409-hessian-based-frangi-vesselness-filter) (see http://imagej.net/Frangi) that would need to be tackled.

stelfrich:
@tinevez is working on a `TubenessOp` at https://github.com/tinevez/imagej-ops/commit/84ebb7d1538356d809846facee2d6ec5ca35d4b8. The `FrangiVesselnessOp` and the `TubenessOp` should share most of their boilerplate code...

danielsnider:
+1

imagejan:
See also [this discussion](http://forum.imagej.net/t/frangi-vesselness-filter-feedback/6747?u=imagejan) on the forum.

stelfrich:
Fixed by https://github.com/imagej/imagej-ops/commit/edd451315debec105fa0b121e6d40bd4defc2a1 (merging #525).

PR #525, opened 2017-10-05
Frangi Vesselness Filter

gselzer:
This PR adds an implementation for Frangi's Multiscale Vessel Enhancement Filter, a filter that is given a grayscale, white-on-black image of vessel structures and returns an image that highlights the vessel structures and removes noise.

This PR closes issue #490.


src/main/java/net/imagej/ops/commands/filter/FrangiVesselness.java
```diff
@@ -0,0 +1,128 @@
...
+import org.scijava.log.LogService;
+import org.scijava.plugin.Parameter;
+import org.scijava.plugin.Plugin;
+
+/**
+ * This command applied the Frangi Vesselness filter operation onto an image,
+ * allowing users to specify:
+ * 		1) The scales desired for the filter application
+ * 		2) The physical spacing between image data points
+ * 		3) Whether or not the pre-Gaussian filter should be performed.
```
stelfrich: I find the term "pre-Gaussian" a bit misleading. To me, this implies that a Gaussian is applied to an image before the computation. In practice, however, it is applied for each scale, isn't it?

src/main/java/net/imagej/ops/commands/filter/FrangiVesselness.java
```diff
@@ -0,0 +1,128 @@
...
+	@Parameter
+	private LogService log;
+
+	@Parameter
+	private OpService opService;
+
+	@Parameter
+	private Img<T> input;
+
+	@Parameter(label = "Gauss before:")
```
stelfrich: Maybe something like "Apply Gaussian at each scale:" would be more appropriate?

src/main/java/net/imagej/ops/commands/filter/FrangiVesselness.java
```diff
@@ -0,0 +1,128 @@
...
+	private boolean doGauss;
+
+	@Parameter(label = "Spacing",
+		description = "Physical distance between the observed data points")
```
stelfrich: Doesn't "Physical distance" imply something like 3 microns? From the implementation, I'd say that is more of a "relative distance": `1,1` implies that we have the same spacing in x and y, whereas `1,1,2` implies that the resolution in z is half of the resolution in x and y.

src/main/java/net/imagej/ops/commands/filter/FrangiVesselness.java
```diff
@@ -0,0 +1,128 @@
...
+			Arrays.fill(result, actual, result.length, result[actual - 1]);
+			log.warn("Not enough values were passed in the " + arrName +
+				" array. The remaining indices were filled with the last value given.");
+		}
+		return result;
+	}
+
+	@Override
+	public void run() {
+	//parse the spacing, and scales strings.
```
stelfrich: Intendation is wrong. Plus, you have used a space in comments before (which is also the preferred way according to the coding style).

src/main/java/net/imagej/ops/filter/FilterNamespace.java
```diff
@@ -47,6 +47,7 @@
 import net.imglib2.algorithm.neighborhood.Shape;
 import net.imglib2.img.Img;
 import net.imglib2.outofbounds.OutOfBoundsFactory;
+import net.imglib2.type.BooleanType;
```
stelfrich: Where did that come from?

src/main/java/net/imagej/ops/filter/vesselness/DefaultFrangi.java
```diff
@@ -0,0 +1,282 @@
...
+		}
+
+		return Math.sqrt(distance);
+	}
+
+	private double derivative(double val1, double val2, double distance) {
+		return ((val2 - val1) / distance);
+	}
+
+	public long numberOfPoints(RandomAccessibleInterval<T> image) {
```
stelfrich: There is a [`Intervals.numElements(Dimensions)`](https://github.com/imglib/imglib2/blob/1941a7e90b70fb7e5149f8cb757cacb64ab182de/src/main/java/net/imglib2/util/Intervals.java#L519-L526) method in imglib2 that you could re-use.

src/main/java/net/imagej/ops/filter/vesselness/DefaultFrangi.java
```diff
@@ -0,0 +1,282 @@
...
+			if (separation != 0) {
+				distance += (separation * separation);
+			}
+
+		}
+
+		return Math.sqrt(distance);
+	}
+
+	private double derivative(double val1, double val2, double distance) {
```
stelfrich: I guess it should be called `derive` (if we stick to verbs as method names).

src/main/java/net/imagej/ops/filter/vesselness/DefaultFrangi.java
```diff
@@ -0,0 +1,282 @@
...
+		if (spacing == null) {
+			spacing = new double[input.numDimensions()];
+			for (int i = 0; i < input.numDimensions(); i++)
+				spacing[i] = 1;
+		}
+		
+		frangi(input, output, scale);
+	}
+
+	private void frangi(RandomAccessibleInterval<T> in, RandomAccessibleInterval<U> out, int step) {
```
stelfrich: `final` parameters?

src/main/templates/net/imagej/ops/Ops.list
```diff
@@ -104,6 +104,7 @@ namespaces = ```
 		[name: "paddingIntervalOrigin",          iface: "PaddingIntervalOrigin"],
 		[name: "sigma",                          iface: "Sigma",               aliases: ["sigmaFilter", "filterSigma"]],
 		[name: "variance",                       iface: "Variance",            aliases: ["varianceFilter", "filterVariance", "var", "varFilter", "filterVar"]],
+		[name: "vesselness",			 iface: "Vesselness"],
```
stelfrich: `frangiVesselness`?

src/main/java/net/imagej/ops/filter/vesselness/DefaultFrangi.java
```diff
@@ -0,0 +1,282 @@
...
+		RandomAccess<T> behind = osmf.create(in);
+		RandomAccess<T> ahead = osmf.create(in);
+
+		RandomAccess<U> outputRA = out.randomAccess();
+
+		while (cursor.hasNext()) {
+
+			cursor.fwd();
+			
+
```
stelfrich: Blank line?

src/main/java/net/imagej/ops/filter/vesselness/DefaultFrangi.java
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	private void frangi(RandomAccessibleInterval<T> in, RandomAccessibleInterval<U> out, int step) {
+
+		// create denominators used for gaussians later.
+		double ad = 2 * alpha * alpha;
+		double bd = 2 * beta * beta;
+
+		// OutOfBoundsMirrorStrategy for use when the cursor reaches the edges.
+		OutOfBoundsMirrorFactory<T, RandomAccessibleInterval<T>> osmf = new OutOfBoundsMirrorFactory<T, RandomAccessibleInterval<T>>(
```
stelfrich: You could make this a `@Parameter` of the op and set the `OutOfBoundsMirrorFactory` as default.

src/test/java/net/imagej/ops/filter/vesselness/FrangiVesselnessTest.java
```diff
@@ -0,0 +1,98 @@
...
+			assertEquals(expectedRA.get().get(), actualRA.get().get(), 0);
+		}
+	}
+
+	@Override
+	protected Context createContext() {
+		return new Context(OpService.class, OpMatchingService.class, CacheService.class, ScriptService.class);
+	}
+
+}
```
stelfrich: Missing newline

src/main/java/net/imagej/ops/commands/filter/FrangiVesselness.java
gselzer: Yes, that is a better way of expressing that.

src/main/java/net/imagej/ops/commands/filter/FrangiVesselness.java
gselzer: You're right. I fixed it.

src/main/java/net/imagej/ops/commands/filter/FrangiVesselness.java
gselzer: Yeah, that sounds better.

src/main/java/net/imagej/ops/commands/filter/FrangiVesselness.java
gselzer: Fixed!

src/main/java/net/imagej/ops/filter/FilterNamespace.java
gselzer: Gone now!

src/main/java/net/imagej/ops/filter/vesselness/DefaultFrangi.java
gselzer: What would be the benefit of this?

src/main/java/net/imagej/ops/filter/vesselness/DefaultFrangi.java
gselzer: Oops.

src/main/java/net/imagej/ops/filter/vesselness/DefaultFrangi.java
gselzer: Sure, I guess we could add that.

src/main/java/net/imagej/ops/filter/vesselness/DefaultFrangi.java
gselzer: That is handy!

gselzer: Looking back, I am not actually sure why this is included in the class. I don't know if I meant to give this a purpose but I cannot find an instance in which this method is called. Should I just delete it?

src/main/templates/net/imagej/ops/Ops.list
gselzer: I suppose this makes sense to do.

src/test/java/net/imagej/ops/filter/vesselness/FrangiVesselnessTest.java
gselzer: This should be fixed in the clean up commit.

src/main/templates/net/imagej/ops/Ops.list
```diff
@@ -104,7 +104,7 @@ namespaces = ```
 		[name: "paddingIntervalOrigin",          iface: "PaddingIntervalOrigin"],
 		[name: "sigma",                          iface: "Sigma",               aliases: ["sigmaFilter", "filterSigma"]],
 		[name: "variance",                       iface: "Variance",            aliases: ["varianceFilter", "filterVariance", "var", "varFilter", "filterVar"]],
-		[name: "vesselness",			 iface: "Vesselness"],
+		[name: "frangiVesselness",			 iface: "FrangiVesselness"],
```
ctrueden: Be careful with the word [refactoring](https://en.wikipedia.org/wiki/Code_refactoring)—it means to change the internal structure of code _without_ changing its external behavior. So no API changes! In this case, the API definitely does change because the op interface name changed from `Vesselness` to `FrangiVesselness`. So this is _not_ a refactoring. A better commit message would be "Rename vesselness to frangiVesselness."

gselzer: Fixed!

src/main/templates/net/imagej/ops/Ops.list
```diff
@@ -104,6 +104,7 @@ namespaces = ```
 		[name: "paddingIntervalOrigin",          iface: "PaddingIntervalOrigin"],
 		[name: "sigma",                          iface: "Sigma",               aliases: ["sigmaFilter", "filterSigma"]],
 		[name: "variance",                       iface: "Variance",            aliases: ["varianceFilter", "filterVariance", "var", "varFilter", "filterVar"]],
+		[name: "frangiVesselness",			 iface: "FrangiVesselness"],
```
imagejan: This should now go to `f` in the alphabet (and indentation fixed), right?

gselzer: Oops. You are right. I will change it here and in the namespace.

src/main/java/net/imagej/ops/filter/vesselness/DefaultFrangi.java
stelfrich: Sorry for the delay, @gselzer!

We have made the `OutOfBoundsFactory` a `@Parameter` in other implementations, so you can feed in another one if you like. I understand that it is an additional parameter to provide, that in practice most user will not overwrite. So I am fine with leaving it as it is, for the sake of usability. However, I wanted to state that this is one of the pitfalls of ImageJ1: making implicit
 assumptions about the out-of-bounds strategy without being able to change it...

Maybe @ctrueden has an opinion on that topic as well?

ctrueden:
Rebased over latest master. All tests passing, so merging now.

PR #527, opened 2017-10-31
The Tubeness filter: enhance filamentous structures of a specified thickness.

tinevez:
The Tubeness filter: enhance filamentous structures of a specified thickness.

- [x] Need help: how to put it in a ops namespace?

Multi-threaded.

This filter works on 2D and 3D image exclusively and produces a score for how "tube-like" each point in the image is. This is useful as a preprocessing step for tracing neurons or blood vessels, for example. For 3D image stacks, the filter uses the eigenvalues of the Hessian matrix to calculate this measure of "tubeness", using one of the simpler metrics me mentioned in Sato et al 1997: if the larger two eigenvalues (λ₂ and λ₃) are both negative then value is √(λ₂λ₃), otherwise the value is 0. For 2D images, if the large eigenvalue is negative, we return its absolute value and otherwise return 0.

- Source image is filtered first by a gaussian with 𝜎 that sets its scale.
- Then the Hessian matrix is calculated for each pixel.
- We yield the eigenvalues of the Hessian matrix. The output of the tubeness filter is a combination of these eigenvalues:
  - in 2D where λ₂ is the largest eigenvalue: out = 𝜎 × 𝜎 × |λ₂| if λ₂ is negative, 0 otherwise.
  - in 3D where λ₂ and λ₃ are the largest eigenvalues:, out = 𝜎 × 𝜎 × sqrt( λ₂ * λ₃ ) if λ₂ and λ₃ are negative, 0 otherwise.

This results in enhancing filaments of roughly 𝜎 / sqrt(d) thickness.
Port of the tubeness filter of the VIB package, with original authors Mark Longair and Stephan Preibisch, to ImageJ-ops.

Example:
Before (SNR = 2):
source

After filtering with tubeness setting sigma = 3.53 (corresponding to a scale of 5 pixels):
tubeness

The code to generate this demo:
``` java
	public static void main( final String[] args )
	{
		ImageJ.main( args );

		final int width = 512;
		final ImagePlus imp = NewImage.createShortImage( "Source", width, width, 1, NewImage.FILL_BLACK );
		imp.show();
		final ImageProcessor ip = imp.getProcessor();
		ip.setColor( 1000 );

		final double scale = 5.;

		/*
		 * Draw structures.
		 */

		final int nLines = 10;
		final int[] pxs = new int[nLines - 1];
		final int[] pys1 = new int[nLines - 1];
		final int[] pys2 = new int[ nLines - 1 ];
		for ( int i = 0; i < nLines - 1; i++ )
		{
			final int ox1 = ( 2 * i + 1 ) * width / nLines / 2;
			final int ox2 = ( 2 * i + 2 ) * width / nLines / 2;
			final int oy1 = width / nLines;
			final int oy2 = 2 * width / nLines;
			ip.setLineWidth( 2 * i + 1 );
			ip.drawLine( ox1, oy1, ox2, oy2 );

			ip.fillRect( ox1, 2 * width / nLines + oy1, 5 * ( i + 1 ), 5 * ( i + 1 ) );
			
			final int px = width / nLines + ( int ) ( ( nLines - 2. ) / nLines * width * Math.sin( 2 * Math.PI * i / nLines ) );
			pxs[i] = px;
			final int py = ( int ) ( 6 * width / nLines + width / nLines * Math.cos( Math.PI * i ) );
			pys1[ i ] = py;
			pys2[ i ] = ( int ) ( py + 2 * 1.5 * width / nLines );

			ip.setLineWidth( 1 );
			ip.drawLine( width / nLines * i, ( int ) ( ( nLines - 2.5 ) * width / nLines ), width / nLines * i, width );
			ip.drawLine( ( int ) ( width / nLines * ( i + 0.5 ) ), ( int ) ( ( nLines - 2.5 ) * width / nLines ), ( int ) ( width / nLines * ( i + 0.5 ) ), width );
		}

		ip.setLineWidth( ( int ) scale );
		final PolygonRoi roi1 = new PolygonRoi( pxs, pys1, pxs.length, PolygonRoi.POLYLINE );
		roi1.fitSpline();
		ip.draw( roi1 );
		final PolygonRoi roi2 = new PolygonRoi( pxs, pys2, pxs.length, PolygonRoi.POLYLINE );
		roi2.fitSpline();
		ip.draw( roi2 );

		/*
		 * Corrupt by noise.
		 */

		ip.noise( 500 );
		imp.updateAndDraw();

		/*
		 * Apply tubeness
		 */


		final double sigma = scale / Math.sqrt( 2 );
		final double[] calibration = new double[] { 1., 1. };

		final Img img = ImageJFunctions.wrap( imp );
		final Context context = ( Context ) IJ.runPlugIn( "org.scijava.Context", "" );
		final OpService ops = context.getService( OpService.class );

		final UnaryHybridCF< Img, Img > tubeness = Hybrids.unaryCF( ops, TubenessOp.class, Img.class, img,
				sigma, calibration );
		final Img out = tubeness.calculate( img );
		ImageJFunctions.show( out, "Tubeness at sigma = " + sigma );

	}
```


tinevez:
Also efficient to remove slowly moving background. With a heavy intensity ramp:

Before:
rampsource

After:
ramptubeness

hadim:
Nice work. This is going to be a super useful Op !

imagejan:
@tinevez wrote:
> Need help: how to put it in a ops namespace?

You could model this after https://github.com/imagej/imagej-ops/pull/525/files.

* Add the `Tubeness` interface to the filter namespace in [`Ops.list`](https://github.com/imagej/imagej-ops/blob/ba217ba473773c38e8187c0a26a3ef4929246d91/src/main/templates/net/imagej/ops/Ops.list#L92-L116)
* Rename your op to `DefaultTubeness` (there might be different implementations in the future) and annotate with `@Plugin(type = Ops.Filter.Tubeness.class)`
* Add a `public tubeness()` method to [`FilterNamespace.java`](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/filter/FilterNamespace.java)

(I might have missed something more..)

tinevez:
Done! Thanks @imagejan .

Here is now the updated example code with the ramp:

```java
	public static void main( final String[] args )
	{
		ImageJ.main( args );

		final int width = 512;
		final ImagePlus imp = NewImage.createShortImage( "Source", width, width, 1, NewImage.FILL_BLACK );
		final ImageProcessor ip = imp.getProcessor();
		// Add a ramp


		ip.setColor( 1000 );

		final double scale = 5.;

		/*
		 * Draw structures.
		 */

		final int nLines = 10;
		final int[] pxs = new int[nLines - 1];
		final int[] pys1 = new int[nLines - 1];
		final int[] pys2 = new int[ nLines - 1 ];
		for ( int i = 0; i < nLines - 1; i++ )
		{
			final int ox1 = ( 2 * i + 1 ) * width / nLines / 2;
			final int ox2 = ( 2 * i + 2 ) * width / nLines / 2;
			final int oy1 = width / nLines;
			final int oy2 = 2 * width / nLines;
			ip.setLineWidth( 2 * i + 1 );
			ip.drawLine( ox1, oy1, ox2, oy2 );

			ip.fillRect( ox1, 2 * width / nLines + oy1, 5 * ( i + 1 ), 5 * ( i + 1 ) );
			
			final int px = width / nLines + ( int ) ( ( nLines - 2. ) / nLines * width * Math.sin( 2 * Math.PI * i / nLines ) );
			pxs[i] = px;
			final int py = ( int ) ( 6 * width / nLines + width / nLines * Math.cos( Math.PI * i ) );
			pys1[ i ] = py;
			pys2[ i ] = ( int ) ( py + 2 * 1.5 * width / nLines );

			ip.setLineWidth( 1 );
			ip.drawLine( width / nLines * i, ( int ) ( ( nLines - 2.5 ) * width / nLines ), width / nLines * i, width );
			ip.drawLine( ( int ) ( width / nLines * ( i + 0.5 ) ), ( int ) ( ( nLines - 2.5 ) * width / nLines ), ( int ) ( width / nLines * ( i + 0.5 ) ), width );
		}

		ip.setLineWidth( ( int ) scale );
		final PolygonRoi roi1 = new PolygonRoi( pxs, pys1, pxs.length, PolygonRoi.POLYLINE );
		roi1.fitSpline();
		ip.draw( roi1 );
		final PolygonRoi roi2 = new PolygonRoi( pxs, pys2, pxs.length, PolygonRoi.POLYLINE );
		roi2.fitSpline();
		ip.draw( roi2 );

		/*
		 * Add a ramp.
		 */

		final ImagePlus ramp = NewImage.createShortImage( "Source", width, width, 1, NewImage.FILL_RAMP );
		ramp.getProcessor().multiply( 1. / 5. );
		final ImagePlus imp0 = new ImageCalculator().run( "add create", imp, ramp );

		imp0.show();

		/*
		 * Corrupt by noise.
		 */

		imp0.getProcessor().noise( 500 );
		imp0.updateAndDraw();

		/*
		 * Apply tubeness
		 */

		final double sigma = scale / Math.sqrt( 2 );
		final Img< UnsignedShortType > img = ImageJFunctions.wrap( imp0 );
		final Context context = ( Context ) IJ.runPlugIn( "org.scijava.Context", "" );
		final OpService ops = context.getService( OpService.class );
		final Img< DoubleType > out = ops.create().img( img, new DoubleType() );
		ops.filter().tubeness( out, img, sigma );
		ImageJFunctions.show( out, String.format( "Tubeness at sigma = %.1f pixels.", sigma ) );
	}
```

src/main/java/net/imagej/ops/filter/FilterNamespace.java
```diff
@@ -30,8 +30,11 @@
 
 package net.imagej.ops.filter;
 
+import java.util.Arrays;
 import java.util.List;
 
+import org.scijava.plugin.Plugin;
```
imagejan: According to [ImageJ code style](https://imagej.net/Coding_style#Eclipse_code_style_profiles), the imports should be ordered strictly alphabetical.

src/main/java/net/imagej/ops/filter/tubeness/TubenessOp.java
```diff
@@ -0,0 +1,14 @@
+package net.imagej.ops.filter.tubeness;
+
+import net.imagej.ops.Ops;
+import net.imagej.ops.special.hybrid.UnaryHybridCF;
+import net.imglib2.IterableInterval;
+import net.imglib2.RandomAccessibleInterval;
+import net.imglib2.type.numeric.RealType;
+import net.imglib2.type.numeric.real.DoubleType;
+
+public interface TubenessOp< T extends RealType< T > > extends Ops.Filter.Tubeness,
```
imagejan: If I got the idea of `Ops.list` and `Ops.vm` correctly, this interface should not be necessary, as an interface is automatically created at `Ops.Filter.Tubeness` using template parsing. So if you change your `DefaultTubeness` annotation to `@Plugin( type = Ops.Filter.Tubeness.class )` it should work.

src/main/java/net/imagej/ops/filter/tubeness/DefaultTubenessOp.java
```diff
@@ -0,0 +1,263 @@
...
+				return;
+
+			final AbstractUnaryComputerOp< Iterable< DoubleType >, DoubleType > method;
+			switch ( numDimensions )
+			{
+			case 2:
+				method = new Tubeness2D( sigma );
+				break;
+			case 3:
+				method = new Tubeness3D( sigma );
```
imagejan: Codacy pointed out that there should be a `break;` here, no?

src/main/java/net/imagej/ops/filter/tubeness/DefaultTubenessOp.java
```diff
@@ -0,0 +1,263 @@
...
+			else
+				output.set( sigma * sigma * Math.sqrt( val1 * val2 ) );
+
+		}
+	}
+
+	// -- Cancelable methods --
+
+	/** Reason for cancelation, or null if not canceled. */
+	private String cancelReason;
```
imagejan: How about moving this field definition to the top of the class? If you (and/or others) disagree, we can also disable the codacy rule to avoid these annoying commit quality checks :)

src/main/java/net/imagej/ops/filter/tubeness/TubenessOp.java
tinevez: Other ops in the filter namespace have the same hierarchy. They have the `Tubeness` interface in `Ops.list` and the mother `TubenessOp` that extends it AND specifies the type of ops (hybridy, unary, etc) in the code package.

src/main/java/net/imagej/ops/filter/FilterNamespace.java
tinevez: I am using imglib2 formatter. Is there an ImageJ formatter?

imagejan: Yes: https://github.com/imagej/imagej/tree/master/config

src/main/java/net/imagej/ops/filter/tubeness/TubenessOp.java
imagejan: Oh, I see. I didn't fully understand the hierarchy yet, sorry for the buzz then.

src/main/java/net/imagej/ops/filter/tubeness/DefaultTubenessOp.java
ctrueden: Personally I prefer field definitions at the top of classes, not in the middle. https://imagej.net/Coding_style#Ordering_of_code_blocks

src/main/java/net/imagej/ops/filter/tubeness/TubenessOp.java
tinevez: Ok. Let's simplify then.

src/main/java/net/imagej/ops/filter/FilterNamespace.java
tinevez: Noted.

imagejan:
@tinevez I had to add a method signature (for `Computer` in addition to `HybridCF`) to the `FilterNamespace`, and remove method that runs with default calibration of 1, to satisfy `FilterNamespaceTest.testCompleteness` and make the build succeed (see https://github.com/imagej/imagej-ops/pull/527/commits/24d17a022600d66917e4c6e667dcda4d9ac30d4c).

I'm not sure where an additional signature `tubeness(out?, in, sigma)` (that assumes a calibration of 1) would have to go. Maybe directly in `DefaultTubenessOp` instead of the namespace?

Maybe @gab1one, @dietzc or @ctrueden can advise here.

Also, it would be good if we add a test for the op itself, no?

Otherwise, after rebasing onto current `master` and squashing a few commits, this should be good to go 👍

tinevez:
@imagejan Thanks!

- I was missing this part on signature. I noticed I had compile error but did not understand them. I guess you could spend some time teaching me requirements to pass ImageJ ops next time we meet.

- The extra signature can go you are right. 

- What kind of test would you suggest?

tferr:
@tinevez ,
this is really really nice! I've been slowly rewriting SNT so that it becomes more aware of scijava and less dependent on IJ1.  Most of the work has been to allow SNT to simultaneously parse both input image and a tubeness-filtered counterpart. So this op is incredibly useful.  

SNT uses the original [TubenessProcessor](https://github.com/fiji/VIB-lib/blob/master/src/main/java/features/TubenessProcessor.java) in multiple places. I fear I won't have the bandwidth to replace all the legacy TubenessProcessor calls. When you ported the code from Mark and Stephan, did you implement any improvements/changes to the filter itself?

BTW, I added a mention to this PR on https://imagej.net/Tubeness

tinevez:
Thanks @tferr !
No improvement, this pretty much sticks to the original code. 
We benefit from multi-threading automatically, thanks to existing ImgLib2 routines by @hanslovsky though.

ctrueden:
I have cleaned up this branch! Rebased over the latest master, squashed things together, added license headers, code formatting, etc.

There are now two commits: the first adds the `filter.tubeness` op, and the second adds the live example.

However, the second is still a WIP, because:

* The example uses ImageJ 1.x to produce an awesome test image. This code should instead be a regression test that verifies that the output matches the expected.
* We should consider whether to keep the ImageJ 1.x usage here—nothing else in the test code uses IJ1 except for a single IJ1-specific benchmark. It would be nice to eliminate the IJ1 dependency. Unfortunately, IJ2 is not up to snuff yet w.r.t. drawing these nice structures
* One way to eliminate the IJ1 usage in the short term: I converted the IJ1 code to a Groovy script, added as part of the WIP, which generates the fancy test image using IJ1. We could commit this code, together with a generated image intended as input. Then use SCIFIO to read in said input image. The script would still be there as documentation for how the image was generated, as well as offering the ability to regenerate the image with different parameters as necessary.

Anyone else have any thoughts or opinions about any of this?

tinevez:
> The example uses ImageJ 1.x to produce an awesome test image. This code should instead be a regression test that verifies that the output matches the expected.

I think it is the right think to do, but cannot be done easily: we do not have the maths to predict analytically the output of this filter. Or do we? Is there a regression test exemple for a spatial filter somewhere?

> We should consider whether to keep the ImageJ 1.x usage here—nothing else in the test code uses IJ1 except for a single IJ1-specific benchmark. It would be nice to eliminate the IJ1 dependency. Unfortunately, IJ2 is not up to snuff yet w.r.t. drawing these nice structures

Let's trash IJ1 dep, oki.

> One way to eliminate the IJ1 usage in the short term: I converted the IJ1 code to a Groovy script, added as part of the WIP, which generates the fancy test image using IJ1. We could commit this code, together with a generated image intended as input. Then use SCIFIO to read in said input image. The script would still be there as documentation for how the image was generated, as well as offering the ability to regenerate the image with different parameters as necessary.

I like this. The disk usage taken by the 2 images is acceptable?

ctrueden:
> The space taken by the 2 images is acceptable?

I'd like to create a smaller input image and compress it as much as possible, but yes. We could also reuse the input test image for multiple different filters, particularly those intended to identify tubelike structures (e.g. Frangi vesselness, ridge detection).

I tried using your image generation logic to make a 128 x 128 with line width 3 and only 5 lines instead of 10, but the algorithm didn't scale quite right. If you are pressed for time, I can ask @gselzer to take care of this—what do you think?

tinevez:
Ok go!

ctrueden:
@gselzer In case you didn't already see this: please complete this branch as discussed above. Happy to go over details on Thursday if anything is not clear.

gselzer:
@imagejan yes, they should be. I just squashed the two together.

All commits build and all tests pass on each commit.

ctrueden:
Thank you @tinevez and @gselzer for all your hard work on this!

PR #529, opened 2017-12-01
Make ifft computer really a computer

bnorthan:
IFFT computer tries to save memory by writing temporary IFFT calcuations
to the input image.  THis means the computer changes the input.  We can
avoid this by making a copy of the input and using that for the temporary
calculations.

This will make the op use more memory and perform slower.  In the future
an op type which allows using the input img for temporary calculations
should be created.


bnorthan:
@ctrueden @kkangle I submitted a potential fix to your problem.  Feel free to modify it, if needed.  

Note, I had to create the ```CreateOp``` in ```compute```, this is because there is no guarantee that the pixel type of the input/output is known when initialize is called.  If you know of a way to get around this let me know.

ctrueden:
Thanks a lot, @bnorthan. I force pushed a simplified version of your fix. Tests are passing, so I'll merge it now.

PR #396, opened 2016-05-13
Add new equation ops

bnorthan:
Add ops for calibrated equations (AbstractCalibratedEquation.java), equations passed in using an op (DefaultCalibratedEquation.java), and equations passed in using a binary lambda
(DefaultXYCalibratedEquation.java).

This allows me to implement equations fairly easily.  For example:

```
// define origin and calibration
double[] origin = new double[] { -1., -1. };
double[] calibration = new double[]{0.1,0.1};

ops.image().equation(image, (x, y) -> Math.pow(x, 2) + Math.pow(y, 2),
            origin, calibration);
```

or (since origin and calibration are optional)

```
ops.image().equation(image, (x, y) -> 10 * (Math.cos(0.3 * x) + Math.sin(
            0.3 * y)));
```

PR #531, opened 2017-12-08
Add new equation ops

bnorthan:
Add ops for coordinate equations (AbstractCoordinateEquation.java).

Including:

1. Equations passed in using an op (DefaultCoordinateEquation.java)

2. X,Y equations passed in using a binary lambda
(DefaultXYCalibratedEquation.java).

### DIFF
--- a/src/main/java/net/imagej/ops/image/ImageNamespace.java
+++ b/src/main/java/net/imagej/ops/image/ImageNamespace.java
@@ -59,7 +59,7 @@ public class ImageNamespace extends AbstractNamespace {
 	@OpMethod(op = net.imagej.ops.image.ascii.DefaultASCII.class)
 	public <T extends RealType<T>> String ascii(final IterableInterval<T> image) {
 		final String result = (String) ops().run(
-				net.imagej.ops.Ops.Image.ASCII.class, image);
+			net.imagej.ops.Ops.Image.ASCII.class, image);
 		return result;
 	}
 
@@ -69,7 +69,7 @@ public class ImageNamespace extends AbstractNamespace {
 		final T min)
 	{
 		final String result = (String) ops().run(
-				net.imagej.ops.Ops.Image.ASCII.class, image, min);
+			net.imagej.ops.Ops.Image.ASCII.class, image, min);
 		return result;
 	}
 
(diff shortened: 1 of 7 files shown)


bnorthan:
This pull request replaces #396.  I rebased it on master, cleaned it up a bit and took out the "calibration" part.  (in the future calibrated equations could be handled as a separate set of commits).  I can still implement calibrated equations easily using just a lambda with the start coordinates and spacing built.  My goal for this commit is to be able to do the following 

Implement an x,y equation using a lambda for example
```
ops.image().equation(image1, (x, y) -> 10 * (Math.cos(0.3 * x) + Math.sin(
			0.3 * y)));
```
Implement a calibrated x,y equation using a lamda 
```
// define start and end
double[] start = new double[] { -1., -1. };
double[] end = new double[] { 1., 1. };

// calculate spacing based on size and starting point
double[] spacing = new double[] { (end[0] - start[0]) / (size[0] - 1),
	(end[1] - start[1]) / (size[1] - 1) };

// implement x^2+y^2 taking into account the calibration (start and spacing)
ops.image().equation(image3, (x, y) -> Math.pow(start[0] + spacing[0] * x,
		2) + Math.pow(start[1] + spacing[1] * y, 2));
```
Implement n-dimensional equations using an op

```
// implement c[0]+10*c[1]+100*c[3]+1000*c[4]
final UnaryFunctionOp<long[], Double> op =
	new AbstractUnaryFunctionOp<long[], Double>()
{

	@Override
	public Double calculate(final long[] coords) {
		final double result = coords[0] + 10 * coords[1] + 100 * coords[2] +
			1000 * coords[3];

		return result;
	}
};

ops.image().equation(image, op);
```

src/main/java/net/imagej/ops/image/ImageNamespace.java
```diff
@@ -441,6 +444,27 @@
 		return result;
 	}
 
+	// -- coordinate equation --
```
awalter17: These should be located next to the other `equation(...)` methods under the `// -- equation --` heading.

src/main/java/net/imagej/ops/image/ImageNamespace.java
```diff
@@ -441,6 +444,27 @@
...
+	@OpMethod(op = net.imagej.ops.image.equation.DefaultCoordinatesEquation.class)
+	public <T extends RealType<T>> IterableInterval<T> equation(
+		final IterableInterval<T> out, final UnaryFunctionOp<long[], Double> in)
+	{
+		@SuppressWarnings("unchecked")
+		final IterableInterval<T> result = (IterableInterval<T>) ops().run(
+			net.imagej.ops.image.equation.DefaultCoordinatesEquation.class, out, in);
```
awalter17: This should be `ops().run(net.imagej.ops.Ops.Image.Equation.class, out, in)`. That way if someone writes another specialized `CoordinatesEquation` op, it could also be matched.

src/main/java/net/imagej/ops/image/ImageNamespace.java
```diff
@@ -441,6 +444,27 @@
...
+		return result;
+	}
+
+	@OpMethod(op = net.imagej.ops.image.equation.DefaultXYEquation.class)
+	public <T extends RealType<T>> IterableInterval<T> equation(
+		final IterableInterval<T> out, final DoubleBinaryOperator in)
+	{
+		@SuppressWarnings("unchecked")
+		final IterableInterval<T> result = (IterableInterval<T>) ops().run(
+			net.imagej.ops.image.equation.DefaultXYEquation.class, out, in);
```
awalter17: Same as above.

src/test/java/net/imagej/ops/image/equation/CoordinateEquationTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+		assertEquals(ops.stats().sum(image2).getRealDouble(), 6.567e7, 0.001);
+
+		// create a float image
+		final IterableInterval<FloatType> image3 = ops.create().img(dimensions,
+			new FloatType());
+
+		// implement x^2+y^2 taking into account the calibration (start and spacing)
+		ops.image().equation(image3, (x, y) -> Math.pow(start[0] + spacing[0] * x,
+			2) + Math.pow(start[1] + spacing[1] * y, 2));
```
awalter17: Here you should use `ops().run(net.imagej.ops.image.equation.DefaultXYEquation.class, out, in)`, to ensure that `DefaultXYEquation` is always being tested and not some other op.

This isn't an issue now, but it could become one later if someone creates another `CoordinatesEquationOp`.

src/test/java/net/imagej/ops/image/equation/CoordinateEquationTest.java
```diff
@@ -0,0 +1,176 @@
...
+				@Override
+				public Double calculate(final long[] coords) {
+					final double result = Math.pow(start[0] + coords[0] * spacing[0], 2) +
+						Math.pow(start[1] + coords[1] * spacing[1], 2);
+
+					return result;
+				}
+			};
+
+		ops.image().equation(image, op);
```
awalter17: Same as above, please use `ops().run(...)`.

src/test/java/net/imagej/ops/image/equation/CoordinateEquationTest.java
```diff
@@ -0,0 +1,176 @@
...
+				@Override
+				public Double calculate(final long[] coords) {
+					final double result = coords[0] + 10 * coords[1] + 100 * coords[2] +
+						1000 * coords[3];
+
+					return result;
+				}
+			};
+
+		ops.image().equation(image, op);
```
awalter17: Please use`ops().run(...)`, see my comment on line 102.

awalter17:
@bnorthan I split your commit so that the clean up changes are in their own commit. This makes the "Add new equation ops" (dc9c94a) commit easier to read. I also rebased this branch over the current master, and updated the license headers for your new classes.

There's a couple of minor changes that need to be made (see my comments), but after that I think this is good to merge!

Thank you for all your work on this, it'll be great to have this merged! 😺

bnorthan:
Thanks for reviewing.  I'll take a look at this tomorrow, and try to get the revisions you suggested done.

bnorthan:
Hi @awalter17 

I just updated this pull request based on your feedback.  Please let me know if there is anything else I should do.

awalter17:
@bnorthan Thanks for making those changes! I've merged this branch!

PR #533, opened 2017-12-12
Real transform

bnorthan:
This commit applies a realtransform to an input image.  

I am very open to feedback and suggestions .  The way I've implemented this, may be anti-intuitive.  

For example 

transform.scale(2) 
implements...
f(x) -> f(2x)

So compresses the image, while transform.scale(.5) expands the image

and

transform.translate(-2)  
implements...
f(x) -> f(x-2) 

So a translation of -2 moves the image forward, and "2" moves it backward

Also

1.  Images are rotated and scaled around origin, so to rotate around the center of the image, you have to translate to the origin first.

2.  It seems transforms are applied in "first in-> last out" order 

Here is a complete example

```
AffineTransform2D transform = new AffineTransform2D();

// create a series of transforms that are applied first in -> last out

// transform so image is centered at w/2, h/2
transform.translate(-image.dimension(0) / 2, -image.dimension(0) / 2);
// scale
transform.scale(1.5);
// rotate
transform.rotate(1);
// translate so image is centered on origin
transform.translate(image.dimension(0) / 2, image.dimension(0) / 2);

Img<UnsignedByteType> actualOutput = ops.create().img(image);

ops.transform().realTransform(actualOutput, image, transform);
```
Input
image
Output
image


src/main/java/net/imagej/ops/transform/TransformNamespace.java
```diff
@@ -1207,4 +1207,59 @@ public String getName() {
...
+	@OpMethod(
+		op = net.imagej.ops.transform.realTransform.DefaultTransformView.class)
+	public <T extends RealType<T>> RandomAccessibleInterval<T> realTransform(
+		final RandomAccessibleInterval<T> in,
+		final InvertibleRealTransform transform)
+	{
+		final RandomAccessibleInterval<T> result =
+			(RandomAccessibleInterval<T>) ops().run(
+				net.imagej.ops.transform.realTransform.DefaultTransformView.class, in,
+				transform);
```
awalter17: This should be `ops().run(Ops.Transform.RealTransform.class, in, transform)`. That way, in the future, someone could write an op which is a specialized version of this and it could also be matched.

Please change this for all your other `realTransform(...)` namespace methods as well.

src/main/java/net/imagej/ops/transform/realTransform/DefaultTransformView.java
```diff
@@ -0,0 +1,103 @@
...
+{
+
+	/**
+	 * The transform to apply
+	 */
+	@Parameter
+	InvertibleRealTransform transform;
+
+	@Parameter(required = false)
+	private Interval outputInterval = null;
```
awalter17: This doesn't need to be explicitly set to `null`. You can just have `private Interval outputInterval`.

src/main/java/net/imagej/ops/transform/realTransform/DefaultTransformView.java
```diff
@@ -0,0 +1,103 @@
...
+
+	@Parameter(required = false)
+	private InterpolatorFactory<T, RandomAccessible<T>> interpolator = null;
```
awalter17: This can just be `private InterpolatorFactory<T, RandomAccessible<T>> interpolator;`. There's no need for the `= null`.

src/main/java/net/imagej/ops/transform/realTransform/DefaultTransformView.java
```diff
@@ -0,0 +1,103 @@
...
+
+	@Parameter(required = false)
+	private OutOfBoundsFactory<T, RandomAccessibleInterval<T>> outOfBoundsFactory =
+		new OutOfBoundsMirrorFactory<>(OutOfBoundsMirrorFactory.Boundary.SINGLE);
```
awalter17: This isn't used (unless I'm missing something?). On line 97 you call `Views.extendZero(input)`, so zero padding is always the out of bounds strategy. Do you want the user to be able to specify their own out of bounds strategy?

src/test/java/net/imagej/ops/transform/realTransform/RealTransformTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+		final AffineTransform2D transform = new AffineTransform2D();
+
+		transform.translate(-image.dimension(0) / 2, -image.dimension(0) / 2);
+		transform.rotate(1);
+		transform.scale(0.5);
+		transform.translate(image.dimension(0) / 2, image.dimension(0) / 2);
+
+		final RandomAccessibleInterval<UnsignedByteType> actualOutput = ops
+			.transform().realTransform(image, transform);
```
awalter17: Here you should use `ops().run(net.imagej.ops.transform.realTransform.DefaultTransformView.class, image, transform);`, to ensure that `DefaultTransformView` is always tested.

awalter17:
@bnorthan I've rebased this branch over master, updated the license header, and fixed a few whitespace issues. I've also noted a few minor changes that need to be fixed, but after that this should be good to merge.

Thank you for all your work on this! 😺

bnorthan:
@awalter17 I just did all the changes you requested.  Please let me know if there is anything else.  Thanks.

imagejan:
@ctrueden @bnorthan are there any plans to merge this one soon? Or is it on hold until `scijava-ops` are out?
I'm asking because I'd like to use some of this functionality, but I can also fall back to using ImgLib2 directly.

bnorthan:
Hi @imagejan 

It's been a while... I made some changes to this over a year ago but it looks like it slipped through the cracks after that.  I'm not sure what the best approach is for changes to imagej-ops/scijava-ops right now.   I've only made a couple of small changes in the last few months, and just flagged @gselzer so he can make the same change in scijava-ops.

ctrueden:
@gselzer How much effort do you think it would be to port this to scijava-ops soon? I'd rather not add more new features to imagej-ops at this juncture—but I really want this function in the new ops. 😄

gselzer:
@ctrueden I only have time to quickly glance over this, but we can probably add all this functionality with a couple of `OpField`s. I will add it to the TODO list

imagejan:
@gselzer @ctrueden is anything like this usable in the new `scijava-ops` (https://github.com/scijava/incubator) yet?

It would be a good argument for me to get started exploring your recent work of overhauling the ops framework 🙂

gselzer:
@imagejan https://github.com/scijava/incubator/pull/36 😉

PR #536, opened 2018-01-29
Add (P)SNR, RMSE, and MAE

### LINKED ISSUE #521: Add ops for (P)SNR
Add ops for computing the signal-to-noise ratio (SNR) between a reference and noisy image:

```java
float snr = ops().image().snr(reference, noisy);
```

There is an [implementation from the BIG at EPFL](http://bigwww.epfl.ch/sage/soft/snr/) that could be used as a starting point.

stelfrich:
Adds implementations for (peak) signal-to-noise ratio, root mean square error, and mean absolute error of two images (a reference and a test image).

Closes #521.


stelfrich:
Thanks for taking some time to look at his @imagejan!

> Would it make sense to make the out parameter optional, or to add Function signatures in addition to these Computer ones? Or is it wrong to provide a default output type like e.g. DoubleType?

Totally makes sense, especially since the type parameter for the output is obsolete anyway (that was a relic from when I was trying to do all computations directly on the `RealType` objects).

> From the op signature (e.g. in the Op Finder) it wasn't clear to me which of in1 and in2 is the reference and which the test image. Maybe we can rename them to reference and test?

This would be amazing, though I fear that the information used in the Op Finder is extracted from the inherited instance variables `in1` and `in2`. I did at least change the names in the namespace methods.

> Currently, these ops live in the image namespace together with such diverse things as ascii, distancetransform, and histogram. I think this was discussed already, but would it make sense to create nested namespaces inside image, or to split them into other namespaces instead?

Good point as well, but definitely something we can discuss afterward.

stelfrich:
> This would be amazing, though I fear that the information used in the Op Finder is extracted from the inherited instance variables `in1` and `in2`. I did at least change the names in the namespace methods.

I had a quick chat with @ctrueden about this and `scijava-ops` will allow for class-based `@Parameter` annotations for `in1` and `in2`. See [`ParameterTest.java#L197-L198`](https://github.com/scijava/scijava-ops/blob/dev/param/src/test/java/org/scijava/param/ParameterTest.java#L197-L198) for an example of how it could look like.

src/main/java/net/imagej/ops/image/ImageNamespace.java
```diff
@@ -456,6 +457,80 @@
...
 	}
 
+	// -- quality --
+
+	@OpMethod(op = net.imagej.ops.image.quality.DefaultMAE.class)
+	public <T extends RealType<T>> DoubleType mae(final DoubleType out,
+		final IterableInterval<T> reference, final IterableInterval<T> test)
+	{
+		final DoubleType result = (DoubleType) ops().run(
+			net.imagej.ops.image.quality.DefaultMAE.class, out, reference, test);
```
awalter17: This should be `ops().run(Ops.Image.MAE.class, out, reference, test);`. That way if someone creates a more specific/specialized MAE op, it could be matched. Otherwise, `DefaultMAE` would always be used.

src/main/java/net/imagej/ops/image/ImageNamespace.java
```diff
@@ -456,6 +457,80 @@
...
+		return result;
+	}
+
+	@OpMethod(op = net.imagej.ops.image.quality.DefaultMAE.class)
+	public <T extends RealType<T>> DoubleType mae(
+		final IterableInterval<T> reference, final IterableInterval<T> test)
+	{
+		final DoubleType result = (DoubleType) ops().run(
+			net.imagej.ops.image.quality.DefaultMAE.class, reference, test);
```
awalter17: Same as above

src/main/java/net/imagej/ops/image/ImageNamespace.java
```diff
@@ -456,6 +457,80 @@
...
+		return result;
+	}
+
+	@OpMethod(op = net.imagej.ops.image.quality.DefaultPSNR.class)
+	public <T extends RealType<T>> DoubleType psnr(final DoubleType out,
+		final IterableInterval<T> reference, final IterableInterval<T> test)
+	{
+		final DoubleType result = (DoubleType) ops().run(
+			net.imagej.ops.image.quality.DefaultPSNR.class, out, reference, test);
```
awalter17: Should be `ops().run(Ops.Image.PSNR.class, out, reference, test)`, and same idea for the remaining namespace methods you added. They should be `run(...)` with an interface not a concrete implementation.

src/test/java/net/imagej/ops/image/quality/QualityTests.java
```diff
@@ -0,0 +1,93 @@
...
+		final DoubleType snr = new DoubleType();
+		ops.image().snr(snr, reference, test);
+
+		assertEquals(7.0937, snr.getRealDouble(), 0.0001);
+	}
+
+	@Test
+	public void testPSNR() {
+		final DoubleType psnr = new DoubleType();
+		ops.image().psnr(psnr, reference, test);
```
awalter17: Rather than using the namespace methods in tests, you should use `ops.run(net.imagej.ops.image.quality.DefaultPSNR.class, psnr, reference, test);`. This will ensure that this unit test always tests the `DefaultPSNR` and not some other PSNR op.

Same idea for your other tests.

awalter17:
@stelfrich Thanks for your work on this! I modified the name of your unit tests, so maven runs them now. And have suggested a few minor edits.

> I had a quick chat with @ctrueden about this and `scijava-ops` will allow for class-based `@Parameter` annotations for `in1` and `in2.` See [`ParameterTest.java#L197-L198`](https://github.com/scijava/scijava-ops/blob/dev/param/src/test/java/org/scijava/param/ParameterTest.java#L197-L198) for an example of how it could look like.

I don't know if we want to wait until this is resolved before merging this, or if we're OK with merging this without? Basically it is to rename/annotate `in1` and `in2` right? I think for now renaming them in the namespace methods is sufficient. Thoughts? @ctrueden? @imagejan?

I think after the minor edits I suggested, this should be good to merge. But if you object please let me know. 😺

ctrueden:
> I don't know if we want to wait until this is resolved before merging this, or if we're OK with merging this without?

If we wait, we will be waiting for at least a few months. I'd say we should merge before that. Normally, I like the namespace method variable names to match the parameter fields, but in the case of `in`/`in1`/`in2`/`out` I am OK with naming them differently, with the idea that when Ops migrates to the new SJC framework, the names from the namespace methods will serve as a hint for what to name the new `@Parameter` declarations.

stelfrich:
Thanks for your comments, @awalter17! They have been addressed in two separate commits.

> I think after the minor edits I suggested, this should be good to merge.

> I'd say we should merge before that.

Feel free to merge! 👍

PR #537, opened 2018-02-08
Fix issue with wavelengths<545

bnorthan:
The fix was based on the following advice from Jizhou LI, original author
of the diffraction based PSF code:

"The scaling factor was meant to provide an empirical choice of the
sampling rate when approximating the function exp(iW), for a specific
wavelength range (minimum is 545nm). You are right, and it is better to
consider a larger range.

Thus, it can be directly used as "am = 3 * m + 1; “ in line 360, and line
422. The latter one was a bug in my previous java code ;-(."


bnorthan:
This is only a small, couple of line change, based on a discussion with the author of the original code.  The test values also had to be changed.  I'll leave it open for a day, just in case anyone has feedback.

PR #538, opened 2018-02-14
Create from array

Xanthorapedia:
This PR creates ops that wrap 1-D or 2-D primitive arrays around `ArrayImg` or `PlanarImg` of `RealType`s. The ops have been tested with random data to ensure the values and order of the values are as expected. A call to the ops takes the input array (1-D for  `ArrayImg` and 2-D for `PlanarImg` with each 1-D element representing a plane) and the dimension of the image. If the image pixels are of variable length type, an additional parameter is required to specify the number of bits occupied by each point.

The PR is ready for code review.


imagejan:
The diff shows a lot of noise (i.e. deletions and additions of the same stuff), mostly introduced by https://github.com/imagej/imagej-ops/pull/538/commits/85ed7b468e43e3d511bc891948c27b686a344b80.

Can you rebase the branch on master and force push to clean up the history?

awalter17:
@Xanthorapedia Thank you for your work on these changes!

If you could please clean up the git history for this PR, that would be greatly appreciated. As @imagejan noted, there are a couple of commits with a lot of noise. Additionally, there appears to be duplicate commits (i.e. 99652b4db5f7673507b5d702533c0cd96975556b and 65868d93f1e9b4ef0eee703ce53c3e33a8de65ee, etc.) and a merge commit which shouldn't be there. Cleaning up the history would make these changes much easier to review, and would help bring the PR closer to a "merge-able" state.

Also please check to make sure each commit builds with passing tests.

PR #539, opened 2018-02-15
Add op for merging two ImgLabelings

stelfrich:
This combines merges contents of two `ImgLabeling`s into one. If `mask` is provided, labels will only be merged where the mask is positive and empty if negative.

### DIFF
--- a/src/main/java/net/imagej/ops/labeling/LabelingNamespace.java
+++ b/src/main/java/net/imagej/ops/labeling/LabelingNamespace.java
@@ -38,6 +38,7 @@ import net.imagej.ops.OpMethod;
 import net.imglib2.RandomAccessibleInterval;
 import net.imglib2.algorithm.labeling.ConnectedComponents.StructuringElement;
 import net.imglib2.roi.labeling.ImgLabeling;
+import net.imglib2.type.BooleanType;
 import net.imglib2.type.numeric.IntegerType;
 
 import org.scijava.plugin.Plugin;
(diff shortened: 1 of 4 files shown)


src/main/java/net/imagej/ops/labeling/LabelingNamespace.java
```diff
@@ -88,6 +93,41 @@
...
 
+	// -- merge --
+
+	@OpMethod(op = net.imagej.ops.labeling.MergeLabeling.class)
+	public <L, I extends IntegerType<I>> ImgLabeling<L, I> merge(
+		final ImgLabeling<L, I> in1, final ImgLabeling<L, I> in2)
+	{
+		@SuppressWarnings("unchecked")
+		final ImgLabeling<L, I> result = (ImgLabeling<L, I>) ops().run(
+			net.imagej.ops.labeling.MergeLabeling.class, in1, in2);
```
awalter17: This should be `ops().run(Ops.Labeling.Merge.class, in1, in2);`. That way any op which implements the `Ops.Labeling.Merge` interface could be matched, as opposed to always using `net.imagej.ops.labeling.MergeLabeling`. 

This should also be changed in your other `merge(...)` namespace methods.

awalter17:
@stelfrich I've rebased this branch over the current master, and updated the license header. I've suggested one small edit, but after that I think this is good to merge!

Thank you for all your work on this!

stelfrich:
> I've suggested one small edit, but after that I think this is good to merge!

Thanks for looking into this, @awalter17! I have squashed your proposed fixes into one commit (they seemed non-educational). This should be good to merge!

ctrueden:
Thank you @stelfrich and @awalter17!

PR #542, opened 2018-03-01
Improve invert Namespace to prevent mismatching between invert ops

gselzer:
This PR contains a refactoring of the namespaces of the invert ops and all of the relevant changes associated. Initally both `InvertII` and `InvertIIInteger` were contained within a single ops interface, `Invert`; `InvertII` does its math using `double` and is relatively fast while `InvertIIInteger` uses `BigInteger` to accommodate types that cannot fit in `double` and is slower. `InvertII` relies on `RealType` inputs while `InvertIIInteger` relies on `IntegerType`, but otherwise the method signatures are the same for both methods. The similarity between the method signatures for the two ops confused the op matcher when looking to choose between the two, and as a result tests failed. To solve this problem another interface  `InvertBigTypes` was created to be used by `InvertIIInteger` to prevent any confusion on the op matcher's part.

There are a few naming and logistical concerns I have before merging this:

* `InvertBigTypes` and `InvertIIInteger` do not have consistent names: `InvertBigTypes` was chosen as the interface because the only reason `InvertIIInteger` should be called is when people need to use it for bigger type datasets, however I have left `InvertIIInteger` with the name it has always had because it does work for all `IntegerType`s. Despite this I wonder if the naming between the two should be more consistent.

* As of right now `Invert` and `InvertIIInteger` are tested in the same test class and are in the same source package, as they were before. I do not know if this should be changed.

@ctrueden As of now all commits build and all tests pass on each commit.

### DIFF
--- a/src/main/java/net/imagej/ops/image/invert/InvertIIInteger.java
+++ b/src/main/java/net/imagej/ops/image/invert/InvertIIInteger.java
@@ -53,82 +53,84 @@ import org.scijava.plugin.Plugin;
  * @author Gabe Selzer
  */
 @Plugin(type = Ops.Image.Invert.class, priority = Priority.HIGH)
-public class InvertIIInteger<I extends IntegerType<I>, O extends IntegerType<O>> extends
-	AbstractUnaryComputerOp<IterableInterval<I>, IterableInterval<O>> implements
-	Ops.Image.Invert
+public class InvertIIInteger<T extends RealType<T>> extends
+	AbstractUnaryComputerOp<IterableInterval<T>, IterableInterval<T>> implements
+	Contingent, Ops.Image.Invert
 {
 
 	@Parameter(required = false)
-	private I min;
+	private T min;
 
 	@Parameter(required = false)
-	private I max;
+	private T max;
 
-	private UnaryComputerOp<IterableInterval<I>, IterableInterval<O>> mapper;
+	private UnaryComputerOp<IterableInterval<T>, IterableInterval<T>> mapper;
+
 
 	@Override
-	public void compute(final IterableInterval<I> input,
-		final IterableInterval<O> output)
+	public void compute(final IterableInterval<T> input,
+		final IterableInterval<T> output)
 	{
 		if (mapper == null) {
-			final BigInteger minValue = min == null ? minValue(input.firstElement()).getBigInteger() : min.getBigInteger();
-			final BigInteger maxValue = max == null ? maxValue(input.firstElement()).getBigInteger() : max.getBigInteger();
+			final BigInteger minValue = min == null ? ((IntegerType)(minValue(input.firstElement()))).getBigInteger() : ((IntegerType)min).getBigInteger();
+			final BigInteger maxValue = max == null ? ((IntegerType)(maxValue(input.firstElement()))).getBigInteger() : ((IntegerType)max).getBigInteger();
 			final BigInteger minMax = minValue.add(maxValue);
 			mapper = Computers.unary(ops(), Ops.Map.class, output, input,
-				new AbstractUnaryComputerOp<I, O>()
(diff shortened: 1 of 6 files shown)


awalter17:
@gselzer I'm wondering if it would be worth it to just create ops for each of the `IntegerType`s which cannot be lossless-ly represented with `double`.

You could put the logic of `InvertIIInteger` into an abstract class and then have the "large" types (i.e. `LongType`, `Unsigned128BitType`, etc.) extend it. The ops implementations for each "large" type would essentially all be the same, so they could probably be made via a template.

This way each of these ops can be higher priority than the faster `double` precision invert, without forcing `IntegerType`s which can be represented with `double` to use it.

That being said, I think your solution is fine as well.

imagejan:
Please forgive my limited understanding of ops matching and type hierarchies, but isn't the original issue something that could be solved by implementing [`Contingent`](https://github.com/imagej/imagej-ops/blob/8e6d3b25194103038c0c2eb32184f151c2b2bcda/src/main/java/net/imagej/ops/Contingent.java) and using its `conforms()` method in `InvertII` to reject if the type is too big for lossless computation with `double`?

Then, if `InvertII` had a high priority but wouldn't conform for big types, so `InvertIIInteger` would be matched in that case.

ctrueden:
Creating a separate `invertBigTypes` op is not the solution here. It would defeat the purpose of ops: to select the best op for each situation automagically. We need to use priorities—and `Contingent` as needed—to bisect the parameter space.

@imagejan wrote:
> reject if the type is too big for lossless computation with double?

Sure, but there is no way for the system to know whether a particular type is "too big for lossless computation" in this scenario. This detail is not recorded anywhere in the type implementations. Perhaps it should be.

In the meantime, we could use `Contingent` with a hardcoded list of types. @gselzer Let's discuss in person the best way forward here.

src/main/templates/net/imagej/ops/Ops.list
```diff
@@ -197,6 +197,7 @@ namespaces = ```
 		[name: "histogram",                      iface: "Histogram"],
 		[name: "integral",                       iface: "Integral"],
 		[name: "invert",                         iface: "Invert"],
+		[name: "invertBigTypes",                 iface: "InvertBigTypes"],
```
ctrueden: We need to make `invert` work for both "big" and "small" types, without a separate op.

gselzer:
@ctrueden all commits build and all tests pass on each commit. My only remaining concern might be the name `InvertIIInteger`. While I suppose it only does work with integer types (since we decided to keep the BigInteger maths) it is not intended for all `IntegerType`s, only those too big for the doubles math. Can you think of a better name? Maybe `InvertIIBigInteger` (though that seems a little long).

src/main/java/net/imagej/ops/image/invert/InvertIIInteger.java
```diff
@@ -52,83 +52,85 @@
...
+		return max;
+	}
 
+	@Override
+	public boolean conforms() {
+		final T inType = in().firstElement();
+		return inType instanceof Unsigned128BitType ||
+			inType instanceof UnsignedLongType ||
+			inType instanceof UnboundedIntegerType ||
+			inType instanceof LongType;
```
awalter17: Could this be replaced with something like:
```java
final T inType = in().firstElement();
if (inType instanceof IntegerType) {
	final IntegerType<?> copy = (IntegerType<?>) inType.createVariable();
	copy.setInteger(Long.MAX_VALUE);
	return copy.getIntegerLong() == Long.MAX_VALUE;
}
return false;
```

This way we don't need all the `instanceof`s (still one though 😿 ). And it would still work if another large `IntegerType` was added.

gselzer: Great!

PR #548, opened 2018-03-21
Set design lens RI equal to experiment lens RI

bnorthan:
This is a one line change to DefaultCreateKernelGibsonLanni.java that sets the default design lens refractive index (ni0) to be equal to the actual lens refractive index (ni).  Previously the design lens refractive index was hardcoded to 1.5, which resulted in aberrations in the generated PSF if the actual lens ri was not 1.5.  

This issue became apparent, when deconvolving an image that was shared on Twitter by Eugene Katrhuka.  The generated PSF wasn't right at first, but works now https://twitter.com/truenorth_ia/status/975685798229172224.  Twitter is actually a pretty useful way to waste time.  

Assuming all checks pass and no one has any other objections I'll merge this tomorrow.

### DIFF
--- a/src/main/java/net/imagej/ops/create/kernelDiffraction/DefaultCreateKernelGibsonLanni.java
+++ b/src/main/java/net/imagej/ops/create/kernelDiffraction/DefaultCreateKernelGibsonLanni.java
@@ -111,6 +111,8 @@ public class DefaultCreateKernelGibsonLanni<T extends ComplexType<T> & NativeTyp
 	public void initialize() {
 		createOp = (BinaryFunctionOp) Functions.binary(ops(), Ops.Create.Img.class,
 			Img.class, Dimensions.class, NativeType.class);
+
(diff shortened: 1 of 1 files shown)


src/main/java/net/imagej/ops/create/kernelDiffraction/DefaultCreateKernelGibsonLanni.java
```diff
@@ -111,6 +111,8 @@
 	public void initialize() {
 		createOp = (BinaryFunctionOp) Functions.binary(ops(), Ops.Create.Img.class,
 			Img.class, Dimensions.class, NativeType.class);
+		
+		ni0=ni;
```
imagejan: ImageJ-style formatting would be `ni0 = ni;` with spaces

bnorthan:
Thanks for looking at it @imagejan.  I just reran the formatter and pushed again.

PR #551, opened 2018-04-05
Add Pearsons op

etarena:
This pearsons op is based on the ["Fast" Pearsons implementation in Coloc 2](https://github.com/fiji/Colocalisation_Analysis/blob/master/src/main/java/sc/fiji/coloc/algorithms/PearsonsCorrelation.java).


ctrueden:
Thanks @etarena! 😂

PR #552, opened 2018-04-06
Rename to "DefaultPearsons" and "DefaultPValue"

etarena:
To prevent naming conflicts, added "Default" to Op names for Pearsons
and PValue implementations.


ctrueden:
Thank you @etarena !

Issue #550, opened 2018-04-04
Gibson Lanni Diffraction Kernel Discrepancies


eric-czech:
I noticed that results from the ```DefaultCreateKernelGibsonLanni``` class seem to disagree with some other implementations of the same algorithm.  I could very well be missing something and I'm hoping this is good enough context (for @bnorthan and @hijizhou I think), but here's what I found:

I'm using inputs like this to the imagej-ops implementation:

```
Img<DoubleType> img = ops.create().kernelDiffraction(
	new FinalDimensions(64, 64, 9),
	.75, // NA 
	670E-09, // lambda 
	1.33, // Specimen RI 
	1.5,  // Immersion RI 
	377E-9, // Res Lateral 
	1500E-9, // Res Axial
	2000E-9, // Particle Position 
	new DoubleType()
);
```

To attempt to do the same thing with a different implementation I tried the [python version](http://kmdouglass.github.io/posts/implementing-a-fast-gibson-lanni-psf-solver-in-python.html) (after wrapping all that up in a function) with arguments like this:

```
img = psf.GibsonLanni(
    size_x=64, size_y=64, size_z=9,
    na=.75,
    ni0=1.5, ni=1.5, # Immersion RI
    ng0=1.5, ng=1.5, # Coverslip RI
    ns=1.33,         # Specimen RI

    # All of the following are in microns
    wavelength=.67,  # aka lambda
    res_lateral=.377,
    res_axial=1.5,
    pz=2.
)
```

I also then compared this to [PSFGenerator] using the Gibson & Lanni model with those same arguments above.  Both PSFGenerator and the Python version produced essentially identical results that looked something like this (across all 9 XY planes):

![psf-py](https://user-images.githubusercontent.com/6130352/38323162-75b8ca3a-380a-11e8-98b7-b4f8c1f1f790.gif)

While the imagej-ops version produced something like this instead:

![psf-java](https://user-images.githubusercontent.com/6130352/38323200-94e9006e-380a-11e8-8138-3abc6e4163ec.gif)

I actually tried a bunch of other combinations of parameter settings too and never found a case where the Python and PSFGenerator versions differed by much, yet I can't seem to get the imagej-ops results to be comparable.

[...]

PR #553, opened 2018-04-09
Adding 3D kernel test for Gibson & Lanni diffraction model

eric-czech:
Hey @bnorthan (cc: @hijizhou),

I added a test to compare 3D PSF results here, but let me know if you see any issues with it.  The tif stack used for comparison is 12KB, so it seems well within the range of the size of images typically used in other tests.

This is re: https://github.com/imagej/imagej-ops/issues/550


ctrueden:
Awesome, very useful! Thank you!

PR #554, opened 2018-04-19
Add uniform noise op

gselzer:
This PR adds an op to add uniformly distributed noise throughout an image, as an alternative to the non-uniform `AddGaussianNoise` op, renamed in this PR from `AddNoise`.

Questions:
* As of now this Op does not take into account the value of the input, instead filling the output with random noise completely independent of whatever is in the input. Is this something that is desired or should this Op continue to disregard the input?

### DIFF
--- a/src/main/java/net/imagej/ops/filter/addUniformNoise/AddUniformNoiseRealType.java
+++ b/src/main/java/net/imagej/ops/filter/addUniformNoise/AddUniformNoiseRealType.java
@@ -1,0 +1,97 @@
+/*
+ * #%L
+ * ImageJ software for multidimensional image processing and analysis.
+ * %%
+ * Copyright (C) 2014 - 2018 ImageJ developers.
+ * %%
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ * 
+ * 1. Redistributions of source code must retain the above copyright notice,
+ *    this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ * #L%
+ */
+
+package net.imagej.ops.filter.addUniformNoise;
+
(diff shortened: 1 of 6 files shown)


src/main/java/net/imagej/ops/filter/addUniformNoise/AddUniformNoiseRealType.java
```diff
@@ -0,0 +1,103 @@
...
+			throw new IllegalArgumentException("range not allowed by op.");
+	}
+
+	@Override
+	public void compute(I input, O output) {
+		int i = 0;
+		do {
+			final double newVal = (rangeMax - rangeMin) * rng.nextDouble(true, true) +
+				rangeMin + input.getRealDouble();
+			if (newVal <= input.getMaxValue() && newVal >= input.getMinValue()) {
```
imagejan: I wonder what would be the least surprising behavior here. When I have an `UnsignedByteType` image with all pixels at `255`, this would create negative noise only, thereby shifting the image mean to be `<255`. When adding any noise, I'd expect the image mean to be essentially unchanged. Although that's a bit special case, how about making it optional if the result should be 1) clipped, 2) overflowing, or 3) some smarter way of adding noise, such as the one here?

ctrueden:
> When working with ImgLib2 types (such as UnsignedByteType), this is the observed behavior, and it actually comes in handy when working with cyclic colormaps

I suggest caution: there is no guarantee that all ImgLib2 types will behave in this way. (@tpietzsch @axtimwalde please correct me if I am wrong. ) Behavior in this case is left up to the implementation. E.g., IIRC, there are some types that when you `setReal(value)` them with an out-of-bounds value, they get corrupted somehow. @awalter17 do you remember?

awalter17:
> there are some types that when you setReal(value) them with an out-of-bounds value, they get corrupted somehow. @awalter17 do you remember?

I don't remember a problem with `setReal` (sorry!), but there was inconsistency with the out-of-bounds behavior (some wrapped, others clamped). I think _most_ of them have masking/wrapping behavior now (https://github.com/imglib/imglib2/pull/96 - this was for constructors but I think it affected the setters as well). That being said, I don't think `RealType` guarantees this wrapping behavior for `setReal` or any of the setters.

imagejan:
> I suggest caution: there is no guarantee that all ImgLib2 types will behave in this way.

Yes, I figured it would be undefined behavior (as always in ImgLib2, it's the callers responsibility to conform to types and intervals...), that's why i said "observed" behavior :smile: Ok, I will not rely on that, but the "cyclic scale" use case came to my mind as something where this would be really useful.

> when you `setReal(value)` them with an out-of-bounds value

BTW, are there terms to distinguish between out-of-bounds for intervals (i.e. x,y,z,...) and out-of-bounds for types (i.e. values)?

Anyhow, this is now far off-topic fro this PR. I suggest to simply ignore my comments (and maybe just keep in mind that it would be useful to have consistent clipping/wrapping behavior across ops in general).

gselzer:
Alright, I gave this PR some TLC. @imagejan, after some thinking I realized that I liked your solution a bit better, as I described in 5440061. I found it difficult to treat floating point and integer types the same, however (see 04af3fe), so I created two different Ops. 

I did, however, leave the computations for the integer type Op in `double`s, however, to allow for some support for some of our larger `IntegerType`s. The only other option that I saw was to convert everything to `BigInteger`s, but I did not want to do that for performance reasons. (Note that we cannot simply do this all using Imglib math because that would leave us open to all of the issues of the edge behavior of each type) Let me know what you think.

rimadoma:
I think the way setting the seed parameter works should be flipped, i.e. _don't_ use a constant seed unless otherwise specified. At the moment the default behaviour is to always generate the same sequence.

```
@Parameter(required = false)
final Long seed;
...
rng = new Random();
if (seed != null) {
    rng.setSeed(seed);
}
```

rimadoma:
Also `Random` is not thread safe. I don't know how much it matters in practice, but now if one of these ops is run multithreaded behaviour won't be deterministic even with a constant seed.

gselzer:
> I think the way setting the seed parameter works should be flipped, i.e. don't use a constant seed unless otherwise specified.

Thanks for catching that @rimadoma! Should be updated in the uniform noise Ops. It looks like `AddGaussianNoise` and `AddPoissonNoise` do this as well, should we alter them with the risk of breaking backwards compatibility?

> Also `Random` is not thread safe.

Sure, yes. I used `MersenneTwisterFast` at @ctrueden's recommendation (which [is thread-safe](https://github.com/scijava/scijava-common/blob/24b03bc8ba17dc813153f4529a54a50be9db6cc4/src/main/java/org/scijava/util/MersenneTwisterFast.java#L51)). We can switch this in the other noise Ops as well if we decide it is okay to do so.

ctrueden:
The previous default seed parameter was correct. One rule of ops is that they are deterministic. If you want TRULY RANDOM noise in your script, then you must generate a random seed and pass that as the seed. I know it's annoying, but it's one of the guarantees of ops.

Therefore, we also need to be careful about multithreaded pseudorandomness. Just because `MersenneTwisterFast` is thread-safe doesn't mean the random order will be deterministic per thread. You need to use a separate instance of the RNG per thread. Happy to discuss more tomorrow during our meeting, @gselzer.

PR #555, opened 2018-04-27
match GaussRA correctly for all non-RAI RA inputs

axtimwalde:
Thanks to @ctrueden for teaching me about contingencies.


ctrueden:
Thanks @axtimwalde! :beers:

PR #557, opened 2018-06-11
WIP: Add DefaultSharpen and DefaultSmooth Ops

gselzer:
This PR adds two Ops adapted from the Sharpen and Smooth ImageJ core plugins.

This PR is currently a WIP because: 
1) There are edge artifacts present in the Ops and I have not yet figured out the difference between the Ops and the plugins. As far as I can tell the Ops are summing up the neighborhood pixels the same as the Plugin, does. Further investigation is needed.

2) The Plugins should either be deprecated or just a wrapping of these Ops.


gselzer:
That is a good point @bnorthan,  I had little understanding/awarenesss of the convolve Ops, so I didn't even think to use them. At this point I have `DefaultSharpen` providing the same results as before, now using those two Ops, however there is more cleaning to do. It also is correctly handling 3D images

 
image

gselzer:
@bnorthan both Ops now pass most of the work on to helper Ops. Using some `System.nanotime`s in the  I found that the Op and the plugin had comparable completion times. Let me know if there is anything else that you see!

bnorthan:
Hi @gselzer it looks to me like both smooth and sharpen are almost exactly the same.  Also it would be nice to be able to take advantage of the work you have done on slice wise processing and normalization for other kernels.  Do you think it would be a good idea to make an op called something like ```SliceWise2DConvolveAndNormalize``` and then ```DefaultSmooth``` and ```DefaultSharpen``` would just call ```SliceWise2DConvolveAndNormalize``` with the appropriate kernel ??  (In that case scale would be computed in the initialize function or something).

gselzer:
@bnorthan yes, I was also thinking about that as well, however I did not do it because I had no idea where we would house such an Op, since it does multiple things. I suppose maybe it could go in `filter.convolve`?

Now that I think about it #556 is also similar, however it is slightly different in the [calculation of the value](https://github.com/imagej/imagej-ops/blob/a2b49e2a77f37952520ea214fa21efa7838461b5/src/main/java/net/imagej/ops/filter/findEdges/DefaultFindEdges.java#L145) that goes into the output. If we were to do this I think we would want to make the Op a `SliceWise2DConvolve`, and then let each `Op` do its own normalization (because in the case of `findEdges` the scale would be 0 if it was computed like in smooth and sharpen. Would you agree or do you see a better option?

bnorthan:
Hi @gselzer 

You are right we should make the ```Op``` a ```SliceWise2DConvolve``` and then let each Op do it's own normalization.  In the case of the ```findEdges``` it looks like it is 2 convolves and then the root of sum of squares of the two convolutions.  

One other quick question.  Do you think it is necessary to convert to double for the intermediate calculation??  Is there a reason for that??  In the original [plugin](https://github.com/imagej/imagej1/blob/58fec6733d7797d87bb111a3aa2ba2ce8ed9503b/ij/process/ShortProcessor.java#L659) it looks like everything is done in integer math.

bnorthan:
Hi @gselzer 

A few more thoughts after looking more closely at the IJ1 implementations.  

1.  [Smooth](https://github.com/imagej/imagej1/blob/58fec6733d7797d87bb111a3aa2ba2ce8ed9503b/ij/process/ShortProcessor.java#L660) looks to me like it is just a mean.  Or am I missing a detail??  If smooth is just a 9 by 9 mean can we just re-use [DefaultMeanFilter](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/filter/mean/DefaultMeanFilter.java) and pass it a 9 by 9 rectangle?? 

2.  [Find Edges](https://github.com/imagej/imagej1/blob/58fec6733d7797d87bb111a3aa2ba2ce8ed9503b/ij/process/ShortProcessor.java#L672) is a bit specialized because some of the kernel coefficients would be 0, and using convolve would result in un-needed multiplications.  Maybe there is a way to implement it by extending [AbstractNeightbourhoodBasedFilter](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/filter/AbstractNeighborhoodBasedFilter.java) 

3.  [Sharpen](https://github.com/imagej/ImageJA/blob/8642f9b254f65dc3ab85d768b1f3358fce57bc3f/src/main/java/ij/process/ImageProcessor.java#L1508) is implemented with [Convolution](https://github.com/imagej/imagej1/blob/58fec6733d7797d87bb111a3aa2ba2ce8ed9503b/ij/process/ShortProcessor.java#L688).  Do you think it be worth it to add an ```Normalizer``` function to ```NaiveConvolve``` to avoid looping through pixels twice??  (ie instead of calling ```Convolved``` followed by ```Divide``` integrate the normalization into Convolve, as is done in IJ1).

Just some thoughts.  Let me know if I've misunderstood anything.

bnorthan:
Hi @gselzer 

1.  Have you tested to see if ```DefaultMeanFilter``` and ```DefaultSmooth``` produce the same result??  I think the line in question is used for rounding purposes, and the [setReal]( https://github.com/imglib/imglib2/blob/4095c1e2aaffb5dfb725ffe1cdbeb884868fa64a/src/main/java/net/imglib2/type/numeric/integer/AbstractIntegerType.java#L66) function in ```AbstractIntegerType``` may serve the same purpose.  What do you think?

2.  Is ```FindEdges``` much of a priority right now given the ongoing redesign efforts??  I'm happy to review it further, but would it make sense to hold off until after the next major iteration of ops is released?? 

3.  You are probably right.  No need to make ```Convolve``` too complicated unless there is an important reason.  When people optimize for speed they may override the op with a native or GPU implementation anyway.

Issue #559, opened 2018-06-14
IterableMax.compute( Iterable ) always returns DoubleType regardless of input generic parametr


hanslovsky:
This is a follow-up to a [bug report on the ImageJ forum](http://forum.imagej.net/t/stats-max-op-returns-wrong-result-type/11447) and mostly c&p of [my post in that thread](http://forum.imagej.net/t/stats-max-op-returns-wrong-result-type/11447/3?u=hanslovsky):

MWE
```java
import net.imagej.ImageJ;
import net.imglib2.img.array.ArrayImg;
import net.imglib2.img.array.ArrayImgs;
import net.imglib2.img.basictypeaccess.array.FloatArray;
import net.imglib2.type.numeric.real.DoubleType;
import net.imglib2.type.numeric.real.FloatType;

public class Dummy
{

	public static void main( final String[] args )
	{
		final ImageJ ij = new ImageJ();
		final ArrayImg< FloatType, FloatArray > floats = ArrayImgs.floats( 1, 2, 3 );
		final DoubleType doubleResult = ( DoubleType ) ( Object ) ij.op().stats().max( floats );
		final FloatType floatResult = ij.op().stats().max( floats );
		System.out.println( doubleResult + " " + floatResult );
		ij.context().dispose();
	}


```
throws this exception:
```
Exception in thread "main" java.lang.ClassCastException: net.imglib2.type.numeric.real.DoubleType cannot be cast to net.imglib2.type.numeric.real.FloatType
	at Dummy.main(Dummy.java:16)
```
As you can see, that weird `( DoubleType ) ( Object )` cast works in line 15, but the expected correct line 16
```java
final FloatType floatResult = ij.op().stats().max( floats );
```
throws an exception.

I am not an imagej-ops expert by any means but here are my 2 cents: The culprit is probably [`AbstractStatsOp.createOutput`](https://github.com/imagej/imagej-ops/blob/caecb88fafcb5b7d769fdaffca724a09ed943896/src/main/java/net/imagej/ops/stats/AbstractStatsOp.java#L52): It always creates a `DoubleType`.  [`IterableMax`](https://github.com/imagej/imagej-ops/blob/caecb88fafcb5b7d769fdaffca724a09ed943896/src/main/java/net/imagej/ops/stats/IterableMax.java) could override `createOutput`, e.g. something like this:
```java
@Override
public T createOutput( final Iterable< T > input )
{
	return input.iterator().next().createVariable();
}
```
[...]

PR #560, opened 2018-06-14
Change return type of IterableMax and IterableMin

bnorthan:
This pull request addresses #559 by implementing the solution proposed by @hanslovsky.

It also adds a small [test](https://github.com/imagej/imagej-ops/compare/stats_return_types?expand=1#diff-a8c314e13c6da6b4b0a2872da63a8bcaR161) to verify the return type is now the same as the input type.


ctrueden:
Thanks @bnorthan!

While we're at it: do you think the min and max ops should use `T extends Comparable<T>` rather than `T extends RealType<T>`? It looks like the only tricky part would be the current hack of using `getMinValue()`/`getMaxValue()` for the initial value.

PR #561, opened 2018-06-14
Shadows Op

gselzer:
This Op provides an `Op` version of the [ImageJ Core Plugin](https://github.com/imagej/imagej-plugins-commands/tree/master/src/main/java/net/imagej/plugins/commands/correlate) created by Barry Dezonia. 

Below is a demo of the new `Shadows` Op (middle) run on [this image](http://imagej.net/images/abe.tif) (left), taken from [imagej.net](http://imagej.net/images/), alongside the output of the same image run through the Plugin (right).

shadowscomparison

Points of Discussion:
* As of now the `Op` generates its Shadows kernel through the used of an angle `Parameter`, allowing for a single `Op` as opposed to the eight different `Plugin`s that already exist. The use of an angle parameter also allows the user to generate shadows at any desired angle instead of being limited to the eight predefined angles. The only downside to this is that in order to keep the algorithm simple the `Op` uses sine and cosine values to generate the kernel which differs from the linear kernel of the older `Plugin`s, resulting in differing outputs (though in most cases not visually distinguishable). Is it okay that the `Op` differ from the `Plugin` it adapts in order to increase its flexibility?  

TODO:
* Write a test for the `Op`. The above point should be decided on before a test is written. If we decide that the difference in output should be allowed then a simple Regression test can be created, otherwise we could test the `Op` based on the output of the `Plugin` if we deem it necessary to make the output exactly the same.


gselzer:
@ctrueden I found test commit that was for some reason not pushed (probably because it is a mess). The test passes, however it introduces a test-scope dependency on [scijava-io-http](https://github.com/scijava/scijava-io-http) that I am not sure we want. It also forces the download of abe.tif, which seems bad to me. However it does have a method `downloadOnce` that seems valuable; maybe it should go in `DownloadService`. Let me know what you think of this

PR #562, opened 2018-06-26
JOML rotate ops

rimadoma:
* Adds `LinAlg` namespace
* Add JOML dependency
* Adds rotation ops for `Vector3d` and `Vector3f`
* Adds various convenience OpMethods for the above
* Bumps SNAPSHOT version

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -266,6 +266,10 @@
 			<groupId>org.apache.commons</groupId>
 			<artifactId>commons-math3</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.joml</groupId>
+			<artifactId>joml</artifactId>
+		</dependency>
 
 		<!-- Test scope dependencies -->
 		<dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 
 	<groupId>net.imagej</groupId>
 	<artifactId>imagej-ops</artifactId>
-	<version>0.41.2-SNAPSHOT</version>
+	<version>0.42.0-SNAPSHOT</version>
 
 	<name>ImageJ Ops</name>
 	<description>ImageJ Operations: a framework for reusable algorithms.</description>
--- a/src/main/java/net/imagej/ops/OpEnvironment.java
+++ b/src/main/java/net/imagej/ops/OpEnvironment.java
@@ -51,6 +51,7 @@ import net.imagej.ops.geom.GeomNamespace;
 import net.imagej.ops.image.ImageNamespace;
 import net.imagej.ops.imagemoments.ImageMomentsNamespace;
 import net.imagej.ops.labeling.LabelingNamespace;
+import net.imagej.ops.linalg.LinAlgNamespace;
 import net.imagej.ops.logic.LogicNamespace;
 import net.imagej.ops.map.neighborhood.CenterAwareComputerOp;
 import net.imagej.ops.math.MathNamespace;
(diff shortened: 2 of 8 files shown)


rimadoma:
If this PR is accepted, maybe the `linalg` namespace would be a good place for the op in PR #420

ctrueden:
Cool, thanks @rimadoma! Do you think it would be too general to call the namespace `linear` instead of `linalg`? What do other software libraries call it?

rimadoma:
I was thinking this name space would be for linear algebra on (simple) vectors and matrices. I think `linear` could easily lead to a loss of coherence, where the name space hosts ops from vastly different domains of maths and image processing. How about just calling it `vecmath`?

Apache commons calls their package `org.apache.commons.math3.linear`, but I think their situation is different, since the package is just limited to mathematics.

In any case, I see these ops eventually migrating to `scijava-ops` at which point the naming can be reconsidered.

imagejan:
`linalg` is used by [numpy/scipy](https://docs.scipy.org/doc/numpy/reference/routines.linalg.html) as well, and in [MATLAB](https://mathworks.com/help/symbolic/mupad_ug/linear-algebra-library.html) (at least in their outdated MuPAD notebooks), it seems.

I think `linalg` isn't such a bad choice.

bnorthan:
> What do other software libraries call it?

@ctrueden in the last year or so I've mostly been wrapping [BLAS](http://www.netlib.org/blas/) (Basic Linear Algebra Specification), and using a bit of [ND4]([https://nd4j.org/) (N-Dimensionsal Arrays for Java).   I'd vote for calling the namespace ```linalg``` and use it to implement a subset of blas functionality in a user friendly way. 

@rimadoma if you haven't seen it already there is a [transform](https://github.com/imagej/imagej-ops/tree/master/src/main/java/net/imagej/ops/transform) namespace that contains a rotate function.  Although this namespace is specific to imglib2 structures.

I've also been working a [```RealTransform```](https://github.com/imagej/imagej-ops/pull/533/commits/b8825734ad35cefa9c2ab68cfc38a72dc1bb7879) to apply a affine transform to an image.  

Do you guys think all ```rotate``` ops should be in ```linalg``` or in ```transform``` ??  

I also wonder if this functionality belongs in the core imagej-ops project, or should be in a separate ```joml-ops``` repo.    Should ```imagej-ops``` be for ```imglib2``` implementations, and other implementations implemented in separate projects??

xulman:
> Do you guys think all rotate ops should be in transform or is it OK to have a rotate in linalg ??

Do we have/Can we have aliases in the ops world?
It is not always black-and-white to categorize stuff...

ctrueden:
This repository needs to be chopped into multiple different repositories before year's end. I agree that putting the JOML ops into their own repository makes sense. Calling the namespace `linalg` sounds good, since it sounds more consistent and preferred by all. Regarding aliases, yes, ops can have any number of aliases, including aliases in different namespaces. But we should choose them carefully because we will need to support them all into the future—if we want to reclaim a particular alias later for something different, it would still break existing scripts expecting that op to be something else.

rimadoma:
What's your opinion y'all: should this `linalg` repository also host _ojAlgo_ i.e. arbitrary size matrix ops as well? The consensus at least at the MPI-CBG Dresden '17 hackathon was that there's need for both JOML and ojAlgo libraries for linear algebra.

rimadoma:
@ctrueden Enjoy your vacation! I'll start mine before you come back, and won't return until August 2.

ctrueden:
This is a beautiful PR, so I'm going to merge it. We can deal with the modularization separately as part of the more general modularization of ImageJ Ops.

PR #568, opened 2018-07-27
Add n-d support for derivativeGauss

gselzer:
This PR adds support for N-Dimensional Partial Derivative/Gaussian convolutions, adding acceptable input types 

Existing `Op` functionality does not change with the update (i.e. the [existing regression test](https://github.com/imagej/imagej-ops/blob/c6953badc5ce87d6692c19c954eaa19ce4c1bf44/src/test/java/net/imagej/ops/filter/derivativeGauss/DefaultDerivativeGaussTest.java#L62) still passes without any data modification)

TODO: Are any more tests required?


ctrueden:
:+1:

PR #569, opened 2018-08-01
Fix permuteCoordinatesInverseView namespace methods

gselzer:
Before the permuteCoordinatesInverse methods in the transform namespace were calling ops.run() with `Ops.Transform.PermuteCoordinatesView.class`, which was wrong. Now they call with `Ops.Transform.PermuteCoordinatesInverseView.class`.


ctrueden:
Good catch, thanks @gselzer!

PR #570, opened 2018-08-02
Least squares fitting op for quadratic surfaces (quadrics)

rimadoma:
* Adds package `stats.regression`
* Adds package `stats.regression.leastSq`
* Adds an op that solves the equation of a quadratic surface with least squares fitting
* Adds missing dependencies to POM

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -270,6 +270,14 @@
 			<groupId>org.joml</groupId>
 			<artifactId>joml</artifactId>
 		</dependency>
+		<dependency>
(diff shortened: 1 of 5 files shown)


ctrueden:
Thanks @rimadoma!

Would you mind if I rename the op to `leastSquares` instead of `leastSq`? I appreciate the succinctness of short names, but the other stats ops use full names e.g. `stats.sumOfSquares`, and I'd like to be consistent and clear with the names.

rimadoma:
Fair enough, I prefer consistency as well.

PR #572, opened 2018-08-27
Fix DefaultFloodFill Img creation when called as a Function

gselzer:
Previously, when DefaultFloodFill was called as a function its `initialize()` method used [`CreateImgFromInterval`](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/create/img/CreateImgFromInterval.java), which could be problematic for the `Op` if the input image was not also `DoubleType` since that particular create `Op` returns a `DoubleType` `Img`. The Create Op was replaced with [`CreateImgFromRAI`](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/create/img/CreateImgFromRAI.java), which returns an `Img` that is of the same `Type`, avoiding this error.


gselzer:
@ctrueden this looks ready to merge to me. I rebased the commit over master, the singular commit builds with passing tests.

PR #574, opened 2018-08-28
Add Maximum Truncated Kendall Tau (MTKT) Op

etarena:
This colocalization Op is based on the MTKT algorithm from Wang et al. (2017).


imagejan:
How do you get the reference values for the tests? Also, `assertEquals(double, double, double)` should have a positive delta, not `0.0`, no?

ctrueden:
> How do you get the reference values for the tests?

Empirically. They are regression tests, to ensure behavior does not change.

> Also, `assertEquals(double, double, double)` should have a positive delta, not `0.0`, no?

The idea is to fail the test if the known good value is not exactly the same as expected, hence the use of a zero delta.

I ran the unit tests on my macOS system in Eclipse, as well as on a Linux system. Here are the problems I found:

- `MTKTTest.testMTKTpValue` fails with `expected:<0.24> but was:<0.0>`
- `MTKTTest.testMTKTpValueZeroCorr` fails with `expected:<0.58> but was:<0.4>`
- `MTKTTest.testMTKTpValuePosCorr` is too slow, taking 59s
- `MTKTTest.testMTKTPosCorr` is too slow, taking 15s
- `MTKTTest.testMTKTpValueZeroCorr` is too slow, taking 3.4s

To make the P-value tests faster, you can use fewer iterations (e.g. 10 instead of 100), and/or use smaller images. I did not dig in to `testMTKTPosCorr`, so do not have advice about that.

etarena:
I think the problem lies in the [DefaultPValue class](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/coloc/pValue/DefaultPValue.java) ... generating seeds for each thread.  If thread # changes - the calculated values change in these tests as well.  So it's not necessarily a Linux versus Mac issue - at least at the moment.  I'm on it.  thanks @ctrueden !!!

etarena:
I believe that this branch is ready to merge at this point... if folks don't mind taking another look - I'd appreciate it.  👍

src/main/java/net/imagej/ops/coloc/ColocUtil.java
```diff
@@ -51,4 +53,15 @@ public static boolean sameIterationOrder(Iterable<?> i1, Iterable<?> i2) {
 			IterableInterval<?> ii2 = (IterableInterval<?>) i2;
 			return Intervals.equalDimensions(ii1, ii2) && Util.equalIterationOrder(ii1, ii2);
 		}
+
+	/** Fisher-Yates shuffle. */
+	public static void shuffle(int[] array, Random rnd) {
```
imagejan: Would it make sense to put this into a utility class somewhere in SciJava? Almost the same is use in SCIFIO's `FakeFormat` [here](https://github.com/scifio/scifio/blob/9562ad0a6c4132bc3881d0483b71d1f25eb6ad1a/src/main/java/io/scif/formats/FakeFormat.java#L830-L839).

(This comment shouldn't block merging this PR, of course.)

etarena: @ctrueden ... thoughts on this one?  I defer to you ...

PR #577, opened 2018-09-12
Fix min/max String Exception bug

gselzer:
When you provide an image of a type that can be negative, the Op can attempt to access the string of valid ascii image characters at a negative index, throwing an error. This PR fixes the bug and adds a test containing the code that threw an error before, as well as making sure that the clamped ascii image is not the same as the unclamped image.

### DIFF
--- a/src/main/java/net/imagej/ops/image/ascii/DefaultASCII.java
+++ b/src/main/java/net/imagej/ops/image/ascii/DefaultASCII.java
@@ -6,13 +6,13 @@
  * %%
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
- * 
+ *
  * 1. Redistributions of source code must retain the above copyright notice,
  *    this list of conditions and the following disclaimer.
  * 2. Redistributions in binary form must reproduce the above copyright notice,
  *    this list of conditions and the following disclaimer in the documentation
  *    and/or other materials provided with the distribution.
- * 
+ *
  * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
  * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
  * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
@@ -46,7 +46,7 @@ import org.scijava.plugin.Plugin;
  * <p>
  * Only the first two dimensions of the image are considered.
  * </p>
- * 
+ *
  * @author Curtis Rueden
  */
 @Plugin(type = Ops.Image.ASCII.class)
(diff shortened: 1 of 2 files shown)


src/main/java/net/imagej/ops/image/ascii/DefaultASCII.java
```diff
@@ -115,6 +117,8 @@ public String calculate(final IterableInterval<T> input) {
 			// normalized = (value - min) / (max - min)
 			tmp.set(cursor.get());
 			tmp.sub(min);
+			// NB: if a value is below min we set tmp to zero.
+			if (tmp.compareTo(zero) < 0) tmp.setZero();
 			final double normalized = tmp.getRealDouble() / span.getRealDouble();
```
ctrueden: As long as we are bending over backwards to use the ImgLib2 math APIs... why don't we use `tmp.div(span)`, and only afterward call `getRealDouble` on the result?

src/main/java/net/imagej/ops/image/ascii/DefaultASCII.java
```diff
@@ -115,6 +117,8 @@ public String calculate(final IterableInterval<T> input) {
 			// normalized = (value - min) / (max - min)
 			tmp.set(cursor.get());
 			tmp.sub(min);
+			// NB: if a value is below min we set tmp to zero.
+			if (tmp.compareTo(zero) < 0) tmp.setZero();
```
ctrueden: It is inconsistent to protect against <0 normalized values here, but then protect against >=1 normalized values inside the `charAt` call below. I would prefer to do both checks the same way: by clamping `charIndex` to `[0, charLen)`. Then you do not need any contortions with a special `zero` variable etc.

gselzer:
@ctrueden I reorganized the math within `DefaultASCII.java` and replaced the `assertTrue` statements with `assertEquals` (or `assertNotEquals`). Let me know if there is anything else that you see. Thanks!

src/main/java/net/imagej/ops/image/ascii/DefaultASCII.java
gselzer: Actually, this will not work if the type does not support decimal values, since we are normalizing the data. I will fix this Wednesday morning.

gselzer:
@ctrueden after pushing the changes that you suggested I got test failures. This is due to the fact that we cannot simply use `tmp.div(span)` get the normalized value to pull from the ASCII string, since this will cause a rounding error if the `RealType` of the image is not `FloatType` or `DoubleType`. I left `double normalization` in the code because I did not see a better way of doing it at the time. Let me know if you have any concerns or ideas as to improving this section.

ctrueden:
I have a thought. This op is nothing else than a convert/scale/dither to the right destination type, followed by a direct mapping of that type's values to character symbols. We should lean on another op foe the first part, and solve the scaling conversion problem really well across a wide variety of types in that helper op.

gselzer:
@ctrueden I took your advice and leaned on some helper `Op`s to simplify the code. Note that we convert to `DoubleType` to prevent rounding errors (correct me if I am wrong, but I think this is necessary) and used `image.normalize` to scale the data down. From there all that the `Op` has to do is cast each data value to an int, then use that int to convert to ASCII. The regression tests that we made still pass.

gselzer:
@ctrueden I rebased this over `master`. It seems to me like it is ready to merge. Thoughts?

PR #583, opened 2018-11-23
FrangiVesselnessCommand: set headless = true

gab1one:
I just noticed this was not set to headless = true. Therefore it could not be used in KNIME via our ImageJ2 integration.


ctrueden:
Thanks @gab1one!

PR #589, opened 2019-01-03
Make Chunk indexes longs

bnorthan:
This pull request changes ```Chunk``` to use long indexes instead of int to potentially address [this](https://forum.image.sc/t/big-images-cause-an-array-index-out-of-bounds-exception/216920) forum issue.

The code seems to build successfully and all tests seem to pass, however since many files were changed there could be potential unforeseen consequences.  Let me know if there is any further due diligence I should do for this pull request.

### DIFF
--- a/src/main/java/net/imagej/ops/features/hog/HistogramOfOrientedGradients2D.java
+++ b/src/main/java/net/imagej/ops/features/hog/HistogramOfOrientedGradients2D.java
@@ -191,7 +191,7 @@ public class HistogramOfOrientedGradients2D<T extends RealType<T>>
 		final CursorBasedChunk chunkable = new CursorBasedChunk() {
 
 			@Override
-			public void execute(int startIndex, int stepSize, int numSteps) {
+			public void execute(long startIndex, long stepSize, long numSteps) {
 				final Cursor<FloatType> cursorAngles = Views.flatIterable(angles).localizingCursor();
 				final Cursor<FloatType> cursorMagnitudes = Views.flatIterable(magnitudes).localizingCursor();
 				final Cursor<FloatType> cursorDerivative0 = Views.flatIterable(finalderivative0).localizingCursor();
@@ -202,7 +202,7 @@ public class HistogramOfOrientedGradients2D<T extends RealType<T>>
 				setToStart(cursorDerivative0, startIndex);
 				setToStart(cursorDerivative1, startIndex);
 
-				for (int i = 0; i < numSteps; i++) {
+				for (long i = 0; i < numSteps; i++) {
 					final float x = cursorDerivative0.get().getRealFloat();
 					final float y = cursorDerivative1.get().getRealFloat();
 					cursorAngles.get().setReal(getAngle(x, y));
--- a/src/main/templates/net/imagej/ops/math/ConstantToArrayImageP.vm
+++ b/src/main/templates/net/imagej/ops/math/ConstantToArrayImageP.vm
@@ -56,6 +56,7 @@ import net.imglib2.type.numeric.real.FloatType;
 import org.scijava.Priority;
 import org.scijava.plugin.Parameter;
 import org.scijava.plugin.Plugin;
+import org.scijava.util.ArrayUtils;
 
 /**
  * Wrapper class for multi-threaded version of binary math operations between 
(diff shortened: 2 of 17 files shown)


src/test/java/net/imagej/ops/thread/RunInterleavedChunkerArray.java
```diff
@@ -49,9 +49,9 @@ public void compute(final A[] input, final A[] output) {
 
 			@Override
 			public void
-				execute(int startIndex, final int stepSize, final int numSteps)
+				execute(long startIndex, final int stepSize, final long numSteps)
 			{
-				int i = startIndex;
+				int i = (int)startIndex;
```
ctrueden: There is a method `org.scijava.util.ArrayUtils.safeMultiply32(long...)` that you can use to downcast a long (or the product of multiple longs) to an int safely. It fails fast if the downcast would overflow.

src/main/templates/net/imagej/ops/math/ConstantToArrayImageP.vm
```diff
@@ -98,15 +98,15 @@ public final class ConstantToArrayImageP {
 			ops().run(ChunkerOp.class, new Chunk() {
 
 				@Override
-				public void execute(final int startIndex, final int stepSize,
-						final int numSteps) {
+				public void execute(final long startIndex, final int stepSize,
+						final long numSteps) {
 					if (stepSize != 1) {
-						for (int i = startIndex, j = 0; j < numSteps; i = i
+						for (int i = (int)startIndex, j = 0; j < numSteps; i = i
```
ctrueden: Should we use `org.scijava.util.ArrayUtils.safeMultiply32(long...)` here?

src/main/templates/net/imagej/ops/math/ConstantToArrayImageP.vm
```diff
@@ -98,15 +98,15 @@ public final class ConstantToArrayImageP {
...
 								+ stepSize, j++) {
 							data[i] ${op.operator}= value;
 						}
 					} else {
-						for (int i = startIndex; i < startIndex + numSteps; i++) {
+						for (int i = (int)startIndex; i < startIndex + numSteps; i++) {
```
ctrueden: Should we use `org.scijava.util.ArrayUtils.safeMultiply32(long...)` here?

src/main/java/net/imagej/ops/features/hog/HistogramOfOrientedGradients2D.java
```diff
@@ -191,7 +191,7 @@ public void compute(RandomAccessibleInterval<T> in, RandomAccessibleInterval<T>
 		final CursorBasedChunk chunkable = new CursorBasedChunk() {
 
 			@Override
-			public void execute(int startIndex, int stepSize, int numSteps) {
+			public void execute(long startIndex, int stepSize, long numSteps) {
```
ctrueden: Is there any value to `long stepSize`? Or do you think `int stepSize` is sufficient always?

ctrueden: If changing it is easy, we might as well, since we are breaking the API anyway right now. Better to break once now, then need to break again in future, no?

If, in the process of changing it, you find the change becoming more involved than anticipated, then please abandon the effort and we can stick with `int`.

ctrueden:
Thank you again, @bnorthan! Is this ready for merge?

src/main/java/net/imagej/ops/features/hog/HistogramOfOrientedGradients2D.java
bnorthan: Changing the interface was straight forward.  I just made the changes and rebased the commits.  Before merging I need to run some final due dillegence tests with > 4G images.  I'll get to that later today or first thing Monday.  In the mean time let me know if there is anything else I should look at.

bnorthan:
Yes.  I just ran a couple of due diligence tests on big (>4G) images and it looks good.

ctrueden:
Thanks @bnorthan! You're the best! 🏆 🍻

mdoube:
Confirmed, fixed related BoneJ issue https://github.com/bonej-org/BoneJ2/issues/163

PR #593, opened 2019-02-01
Correctly order params of distance transform

### LINKED ISSUE #590: distancetransform uses wrong order for in and out
in the following code, which comes from `ImageNamespace.class`, the arguments `in` and `out` are swapped in the signature of `distancetransform`, as the convention seems to be that `out` comes before `in`. More seriously, the arguments are also swapped in the call to `ops().run()`, which **expects** the opposite order (first `out` then `in`), as specified by the `CommandInfo` associated with the `Op` in question.

```java
	/** Executes the "distancetransform" operation on the given arguments. */
	@OpMethod(ops = {
		net.imagej.ops.image.distancetransform.DefaultDistanceTransform.class,
		net.imagej.ops.image.distancetransform.DistanceTransform2D.class,
		net.imagej.ops.image.distancetransform.DistanceTransform3D.class })
	public <B extends BooleanType<B>, T extends RealType<T>>
		RandomAccessibleInterval<T> distancetransform(
			final RandomAccessibleInterval<B> in,
			final RandomAccessibleInterval<T> out)
	{
		@SuppressWarnings("unchecked")
		final RandomAccessibleInterval<T> result =
			(RandomAccessibleInterval<T>) ops().run(Ops.Image.DistanceTransform.class,
				in, out);
		return result;
	}
```

gselzer:
@dariomel @ctrueden I fixed the param order and added a test that called the `Op` as a `Computer` to make sure that this all works as intended. Let me know if there is anything else that you see!

Closes #590

### DIFF
--- a/src/main/java/net/imagej/ops/image/ImageNamespace.java
+++ b/src/main/java/net/imagej/ops/image/ImageNamespace.java
@@ -104,19 +104,14 @@ public class ImageNamespace extends AbstractNamespace {
 	// -- distance transform --
 
 	/** Executes the "distancetransform" operation on the given arguments. */
-	@OpMethod(ops = {
-		net.imagej.ops.image.distancetransform.DefaultDistanceTransform.class,
-		net.imagej.ops.image.distancetransform.DistanceTransform2D.class,
-		net.imagej.ops.image.distancetransform.DistanceTransform3D.class })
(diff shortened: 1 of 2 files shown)


ctrueden:
Thanks @gselzer !

PR #594, opened 2019-02-07
Make filters computers

bnorthan:
This commit makes the FFT Filter functions (like ConvolveFFTF and RichardsonLucyF) into computers and renames them (ie PadAndConvolveFFT and PadAndRichardsonLucy) to better convey their purpose. 

Hopefully this will make these ops work better with the new version of ops.  (@ctrueden and @gab1one mentioned missing Functions or Computers can be automatically generated but it is better to go from Computer to Function, instead of the other way around.)

Note that assuming this commit is compatible with the other new changes,  it would make sense to merge this commit with those changes, as merging it alone would break pre-existing examples (because there are no long filter functions).  

Also note there still exists a separate ```ConvolveFFTC``` this is needed for the case where the inputs are already padded and the FFT memory is setup (for example iterative convolutions called during the Richardson Lucy algorithm).


ctrueden:
Thank you @bnorthan !

PR #598, opened 2019-03-06
Update tests to scifio Location-based API

gab1one:


imagejan:
The `<scifio.version>` property needs to be overridden in `pom.xml` for the builds to pass, no?

gab1one:
Yes and https://github.com/scifio/scifio/pull/391 needs to be merged first

ctrueden:
Even though this was never merged, I think we can get away with throwing away these changes now, because the `LocationResolver` API and restoration of the `String`-based `IO` methods enables sufficient backwards compatibility that the tests still work with newer SCIFIOs. :+1:

PR #602, opened 2019-04-24
Add Fill Holes command

stelfrich:
This PR adds a `Command` that wraps the `Ops.Morphology.FillHoles` op.


ctrueden:
Thanks, @stelfrich!

PR #604, opened 2019-04-26
Remove unnecessary loop, function call from DefaultCoordinatesEquation

gselzer:
Within `DefaultCoordinatesEquation` there existed both an obsolete for loop (on line 68) and an obsolete function call (on line 70). Both of these were removed in order to speed up the Op. All affected tests pass with this change.


gselzer:
@ctrueden I think this can be merged, no?

Issue #579, opened 2018-10-11
watershed crashes when seeds are at edge of mask


tischi:
@imagejan 
please see here: https://forum.image.sc/t/watershed-ops-strange-behaviour/20189

ctrueden:
Thanks for the report, @tischi.

@SimonSchmid Any chance you have time and interest to investigate, since you coded this op?

SimonSchmid:
Hi,
Interest, yes. But I will be very busy for at least the next 6 months and won't find any time I think. Sorry!

ctrueden:
@SimonSchmid Totally understandable. Thanks for responding.

frauzufall:
Hey, I made a [PR](https://github.com/imagej/imagej-ops/pull/615) to fix this, if anyone has time to review it, that'd be great :)

ctrueden:
Fixed by #615.

PR #615, opened 2019-12-30
WatershedSeeded: Fix NoSuchElementException #579

frauzufall:
- in case a seed was on the edge of the mask, a `NoSuchElementException` was thrown
- this commit fixes the issue by doing out of bound checks before accessing the labeling value

I am not very familiar with the details of watershed, but this commit should not introduce any change to the algorithm itself. @tischi can you try if that would fix your issue?


ctrueden:
Thanks @frauzufall!

PR #616, opened 2020-02-06
Parallize BoxCount with more than 1 'translation'

rimadoma:
Following issue [196](https://github.com/bonej-org/BoneJ2/issues/196) on the BoneJ2 hub, speeds up the BoxCounting op, when user has requested more than one "translation". Translations move the boxes around in hopes of finding the best fit, but add work in O(n^d) fashion.

I tried the obvious strategy of parallizing the boxes first, but that didn't run much faster.

### DIFF
--- a/src/main/java/net/imagej/ops/topology/BoxCount.java
+++ b/src/main/java/net/imagej/ops/topology/BoxCount.java
@@ -30,7 +30,6 @@
 package net.imagej.ops.topology;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Spliterator;
 import java.util.stream.IntStream;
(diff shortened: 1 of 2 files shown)


ctrueden:
Thanks @rimadoma.

@gselzer We'll need to port this work to scijava-ops, right?

gselzer:
@ctrueden yeah, this will have to be ported. I'll add it to my list

gselzer:
@ctrueden changes are now ported to scijava-ops work

rimadoma:
@gselzer Thank you! Should I commit future changes there?

gselzer:
@rimadoma so we currently have all of the ops within imagej-ops copied into scijava-ops (see [this branch](https://github.com/scijava/scijava-ops/tree/imagej-rebase/src/main/java/net/imagej/ops)) so that they could be updated to use the new Op framework. Eventually all of the ops, as well as this work, will all make its way back into the imagej-ops repository, once all of the work on scijava-ops is finalized. So long story short, probably not :grin:

Issue #620, opened 2020-03-19
Median not returning mean of two middle values for even number of values


frauzufall:
I wanted to migrate code from Python to ImageJ2 and noticed the difference in calculating the median. In Ops, it does not return the mean of the two middle values for an even number of values.
```
ImageJ ij = new ImageJ();
List<DoubleType> values  = new ArrayList<>();
values.add(new DoubleType(0.));
values.add(new DoubleType(1.));
System.out.println(ij.op().stats().median(values));
```
.. prints `1.0` whereas python would return `0.5`. 

The [test](https://github.com/imagej/imagej-ops/blob/master/src/test/java/net/imagej/ops/stats/StatisticsTest.java#L176-L179) for median would probably not catch this since the unsigned byte values in the middle of a 100x100px image could both be `128`.

ctrueden:
Wow, that's an embarrassing bug!

ctrueden:
@gselzer Please look into the best way of addressing this, in both the current imagej-ops and the new scijava-ops port. Thank you!

rimadoma:
I'm on it for `imagej-ops`. `DefaultMedian` calls `DefaultQuantile` with quantile `0.5`, which (as I understand it) accomplishes a slightly different task: it returns the number in the array with the [nearest rank](https://en.wikipedia.org/wiki/Percentile#The_nearest-rank_method). I'm thinking of rewriting `DefaultMedian` with [this algorithm](https://rcoh.me/posts/linear-time-median-finding/) that guarantees _O(n)_ time.

imagejan:
This bug *might* have been introduced by the "merger" of the `Median`, `Percentile` and `Quantile` ops in https://github.com/imagej/imagej-ops/commit/cd382ca303132d10af1bf97105b86b3795318a10, I'm not sure.

@rimadoma wrote:
> accomplishes a slightly different task: it returns the number in the array with the nearest rank.

According to the [*Quantile*](https://en.wikipedia.org/wiki/Quantile#Even-sized_population) definition, it still should return the average of two values for even-sized distributions. Is this a different case for *Percentile*?

rimadoma:
> According to the Quantile definition, it still should return the average of two values for even-sized distributions.

My interpretation of the current implementation is that it returns the nearest rank. It's always `return array.get(k)`.

> This bug might have been introduced by the "merger" of the Median, Percentile and Quantile ops in #318

[...]

PR #621, opened 2020-03-21
Fix issue #620

rimadoma:
Fixes issue #620 and adds tests for `DefaultMedian`.

### DIFF
--- a/src/main/java/net/imagej/ops/stats/DefaultQuantile.java
+++ b/src/main/java/net/imagej/ops/stats/DefaultQuantile.java
@@ -74,7 +74,7 @@ public class DefaultQuantile<I extends RealType<I>, O extends RealType<O>>
 	 * Returns the value of the kth lowest element. Do note that for nth lowest
 	 * element, k = n - 1.
 	 */
-	private double select(final ArrayList<Double> array, final int inLeft,
+	static double select(final ArrayList<Double> array, final int inLeft,
 		final int inRight, final int k)
 	{
 
@@ -140,7 +140,7 @@ public class DefaultQuantile<I extends RealType<I>, O extends RealType<O>>
 	}
 
 	/** Helper method for swapping array entries */
-	private void swap(final List<Double> array, final int a, final int b) {
+	private static void swap(final List<Double> array, final int a, final int b) {
 		final double temp = array.get(a);
 		array.set(a, array.get(b));
 		array.set(b, temp);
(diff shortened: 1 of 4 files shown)


src/main/java/net/imagej/ops/stats/DefaultQuantile.java
```diff
@@ -61,25 +62,30 @@
...
 		final ArrayList<Double> statistics = new ArrayList<>();
 
-		final Iterator<I> it = input.iterator();
-		while (it.hasNext()) {
-			statistics.add(it.next().getRealDouble());
+		input.forEach(i -> statistics.add(i.getRealDouble()));
+
+		final int k = (int) (statistics.size() * quantile);
+		double result = select(statistics, k);
+		if (statistics.size() % 2 == 0) {
```
imagejan: How about quantiles other than the median? Is the convention to *always* take the mean of two values when the sample is even-sized?

The explanation [on wikipedia](https://en.wikipedia.org/wiki/Quantile#Even-sized_population) suggests that averaging is done for the *median* only, but not for any other quantiles (e.g. quartiles etc.).

The [R Documentation](https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/quantile) lists 9 possible algorithms (see `type` argument) to compute the quantile.

rimadoma: Good question. According to [apache.commons version](https://commons.apache.org/proper/commons-math/javadocs/api-3.6/org/apache/commons/math3/stat/descriptive/rank/Percentile.html) of quantile: _"There are several commonly used methods for estimating percentiles (a.k.a. quantiles) based on sample data. For large samples, the different methods agree closely, but when sample sizes are small, different methods will give significantly different results."_ 

As I'm not familiar with quantiles myself, maybe the wisest thing is to leave `DefaultQuantile` as is, and add this implementation to `DefaultMedian`. What do you think @imagejan?

rimadoma:
@imagejan So I changed the PR so that only `DefaultMedian` changes, not `DefaultQuantile`. Since there seems to be several accepted methods of calculating quantile, I reckon `DefaultQuantile` works the way it does for a reason, and should be left alone.

ctrueden:
Thank you very much, @rimadoma!

@gselzer Please forward port this work to imagej-ops2.

gselzer:
@ctrueden work is ported on the imagej/imagej-ops2/init branch. Thanks @rimadoma!

PR #623, opened 2020-04-25
Add calibration to Marching cubes mesh

rimadoma:
Adds an optional parameter to add calibration to the output mesh of `DefaultMarchingCubes`, as discussed on the [forum](https://forum.image.sc/t/computing-3d-surface-area-of-binary-object-using-imagej-ops/36807/8)


rimadoma:
Is it possible to have a `@Parameter CalibratedAxis[] axes`? I'm getting "last argument should be vararg" errors from namespace tests.

kephale:
This should be accounted for in imagej-mesh: https://github.com/imagej/imagej-mesh/issues/22

frauzufall:
@rimadoma just a note, @tpietzsch and I have been working on significantly speeding up `DefaultMarchingCubes` on [this branch](https://github.com/imagej/imagej-ops/tree/speedup-marchingcubes). I want to add add the new version to imagej-mesh as static methods and call this from ops. I will ping you once there is a PR so that we can also address this issue.

PR #618, opened 2020-03-07
Optimise BoxCount further

rimadoma:
Further optimises ops related to issue https://github.com/bonej-org/BoneJ2/issues/196

Changes execution times for a 256 x 256 x 256 binary image:
* BoxCount _25 s -> 1.1 s_ (Paremeters: maxSize 256, minSize 1, scaling 2.0) 
* Outline _2 s -> 500 ms_

PR #624, opened 2020-05-12
Fix work per thread in Outline

rimadoma:
In PR #618 @ctrueden reported `Outline` tests failing on a Linux box. Since neither of us could reproduce the issue in other environments and Travis CI was happy, I eventually shrugged it off as an isolated incident. Now @mdoube has reported failing tests in BoneJ, which depend on `Outline`. I think it's the same bug. This PR has been tested to fix the issue for him.

~~I don't have the resources to actually solve the bug right now (still can't reproduce). Rather than leave the issue lurking, I'm reverting the op to single threaded execution.~~

@ctrueden Could you please check whether this fixes the issue on the Linux box you mentioned?

### DIFF
--- a/src/main/java/net/imagej/ops/morphology/outline/Outline.java
+++ b/src/main/java/net/imagej/ops/morphology/outline/Outline.java
@@ -46,12 +46,12 @@ import net.imglib2.view.Views;
 import org.scijava.log.LogService;
 import org.scijava.plugin.Parameter;
 import org.scijava.plugin.Plugin;
+import org.scijava.thread.ThreadService;
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
-import org.scijava.thread.ThreadService;
 
 /**
  * The Op creates an output interval where the objects are hollow versions from
--- a/src/main/java/net/imagej/ops/morphology/outline/Outline.java
+++ b/src/main/java/net/imagej/ops/morphology/outline/Outline.java
@@ -39,7 +39,6 @@ import net.imglib2.RandomAccessibleInterval;
 import net.imglib2.outofbounds.OutOfBounds;
 import net.imglib2.type.BooleanType;
 import net.imglib2.type.logic.BitType;
-import net.imglib2.util.Util;
 import net.imglib2.view.ExtendedRandomAccessibleInterval;
 import net.imglib2.view.Views;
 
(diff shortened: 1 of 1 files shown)


frauzufall:
@rimadoma just FYI I can't reproduce the issue on my Linux machine either.

rimadoma:
> @rimadoma just FYI I can't reproduce the issue on my Linux machine either.

Thanks! At least it's not failing on all Linux computers 😄

mdoube:
> I'm reverting the op to single threaded execution.

If I understand `Outline` correctly, it is just checking all pixels to see if they are a) foreground and b) have a background neighbour and if a) and b) are true keeping that pixel, otherwise setting it to background. Is that correct? If so it should be straightforward and robust to parallelise. It would however be a bit trickier to do 'in place' for both single and multi-threaded implementations rather than using a new output image for the result.

rimadoma:
Sorry for spamming again, but I just had an idea what the bug might be. I'm gwant to try it before this gets merged

rimadoma:
The bug was that for very small images (`voxels / nThreads < 2`) the work per thread was calculated incorrectly. The bug only started to break tests when `nThreads >= 15`.

mdoube:
This version of `Outline` is still super-buggy throwing a large number of exceptions. Closing for now.

```
[ERROR] java.util.concurrent.ExecutionException: java.lang.ArrayIndexOutOfBoundsException
	at java.util.concurrent.FutureTask.report(FutureTask.java:122)
	at java.util.concurrent.FutureTask.get(FutureTask.java:192)
	at net.imagej.ops.morphology.outline.Outline.compute(Outline.java:121)
	at net.imagej.ops.morphology.outline.Outline.compute(Outline.java:61)
	at net.imagej.ops.special.hybrid.BinaryHybridCF.calculate(BinaryHybridCF.java:63)
	at net.imagej.ops.special.hybrid.BinaryHybridCF.calculate(BinaryHybridCF.java:85)
	at org.bonej.wrapperPlugins.FractalDimensionWrapper.lambda$run$0(FractalDimensionWrapper.java:178)
	at java.util.ArrayList.forEach(ArrayList.java:1257)
	at org.bonej.wrapperPlugins.FractalDimensionWrapper.run(FractalDimensionWrapper.java:174)
	at org.scijava.command.CommandModule.run(CommandModule.java:196)
	at org.scijava.module.ModuleRunner.run(ModuleRunner.java:163)
	at org.scijava.module.ModuleRunner.call(ModuleRunner.java:124)
	at org.scijava.module.ModuleRunner.call(ModuleRunner.java:63)
	at org.scijava.thread.DefaultThreadService.lambda$wrap$2(DefaultThreadService.java:225)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.ArrayIndexOutOfBoundsException
```

PR #632, opened 2020-08-11
Bugfix: Grey-value co-occurrence matrices should be symmetric

haesleinhuepf:
Hi all,

as discussed with a user and @imagejan on the[forum](https://forum.image.sc/t/clij-optimisation-on-glcm/41240/21) we assume that the grey level co-occurrence matrices should be symmetric. Otherwise, the direction of the read-out would play a role in further post-processing of the matrix.

Maybe @dietzc has an opinion here.

Thanks!

Cheers,
Robert


imagejan:
I think we might want to have both a non-symmetric and a symmetric version of the co-occurrence matrix op. It seems for texture feature calculations really the non-symmetric matrix is used (although I find it surprising, because that means that mirroring an image results in different texture features...).

For reference, here's how it's done in `scikit-image` (where they default to `symmetric=False` and `normed=False`):

https://github.com/scikit-image/scikit-image/blob/2bf1a351a0c2e19fcd37168460a18954d4aabdd7/skimage/feature/texture.py#L15-L105

PR #639, opened 2021-08-16
Outline: use only 1 thread until the multithreading model is fixed.

mdoube:


mdoube:
This PR (and https://github.com/imagej/imagej-ops/commit/d00da72880cb9e8817cf5202cb98a48d39e05110) is redundant and #624 should be merged instead.

PR #640, opened 2021-11-19
DefaultOpService: Fix setMaxThreads to check range of local variable

karlduderstadt:
The range of the global variable maxThreads is being checked in the
setMaxThreads method, but the local variable should be checked. The
global one is zero by default before being set so calling the function
always triggers the IllegalArgumentException no matter what value is
provided.


ctrueden:
Doh! Thanks, @karlduderstadt!

PR #644, opened 2022-06-08
Rewrite OpSearchResults as signatures

gselzer:
This PR rewrites `OpSearchResults`, with the goal to reduce the number of *near-duplicate* search results from the `OpSercher`.

## The `OpSignature`

To reduce the number of duplicate search results, we'd ideally differentiate **only** by algorithm, folding all implementations of that algorithm into one search result. Our task, then, is to group Ops into sets, where each set corresponds to a different algorithm.

In ImageJ Ops, we can write implementations with overloading signatures; we then let the Op matcher choose the best implementation for our inputs. Thus the most consistent folding behavior would align with the folding behavior of the matcher. In other words, groups should share:
* Op name
* Op functional type
* Simplified parameter types for *all* inputs and outputs (such as `image` or `number`)

Ideally, we would also be able to fold secondary parameters into a single search result; this PR does not introduce that functionality into this PR as of writing. I avoided this functionality knowing that SciJava Ops does not differentiate between primary and secondary parameters. Implementing secondary parameter folding would likely result in inconsistent results between ImageJ Ops and ImageJ Ops2. 

To make these conditions concrete, we introduce `OpSignature`, a class designed to encapsulate the information common to a set of implementations. Each `OpInfo` should be able to provide an `OpSignature`, and Ops that are different implementations of the same algorithm should provide the same `OpSignature`.

We also introduce the `OpSignatureModule`, which can execute *any* known Ops described by `OpSignature` given suitable inputs, and the `OpSignatureInfo` that can create the `OpSignatureModule`. These classes allow `OpSignatures` to be runnable in Fiji just like Ops and Modules, along with access to the pre- and postprocessor chain of the Context.

## Searching for Ops

The `OpSearcher` currently returns one result for each *implementation* available containing the searched text in its name. With this PR, the results can easily be grouped by `OpSignature`. Simplified types are provided to the `OpSearcher`, allowing the folding to occur in an extensible way, with easy support for simplified type addition. For each group, only one `OpInfo`'s `OpSignature` is chosen; this `OpSignature` is responsible for delegating to all Ops in the set.

This change will make searching for Ops in a graphical setting (e.g. Fiji, napari) *much* easier, and should be well reviewed.

TODO:
* [ ] Make `OpSearcher`'s simplified types extensible for e.g. napari, provide them externally (with a sensible default).
* [ ] Key the `OpSearcher`'s Op map on `OpSignature`.


src/main/java/net/imagej/ops/OpSignature.java
```diff
@@ -0,0 +1,188 @@
...
+	private final String functionalType;
+	private final List<ModuleItem<?>> inputs;
+	private final List<ModuleItem<?>> outputs;
+
+	public OpSignature(final OpInfo info) {
+		this(info.getName(), info.getType(), info.inputs(), info.outputs());
+	}
+
+	public OpSignature(final String name, final Class<?> functionalType,
+		final List<ModuleItem<?>> inputs, final List<ModuleItem<?>> outputs)
```
gselzer, line 39: I dislike the `List<ModuleItem<?>>` inputs; it doesn't seem like the `OpSignature` should know about them (or about `Module`s in general). But they are a container for so much data we need, and it's so convenient to store the data this way (so that we can call `inputsFor`, for example.

src/main/java/net/imagej/ops/OpSignature.java
```diff
@@ -0,0 +1,188 @@
...
+
+	public List<String> outputNames() {
+		return outputs.stream() //
+			.map(ModuleItem::getName) //
+			.collect(Collectors.toList());
+	}
+
+	@Override
+	public boolean equals(Object that) {
+		// ensure that is a OpSignatureInfo
```
gselzer, line 93: OpSignature, not OpSignatureInfo

src/main/java/net/imagej/ops/OpSignature.java
```diff
@@ -0,0 +1,188 @@
...
+		if (!(that instanceof OpSignature)) return false;
+		OpSignature thatSig = (OpSignature) that;
+		// check name
+		return buildShortSig().equals(thatSig.buildShortSig());
```
gselzer, line 97: I feel like this is not enough?

We should check name, functional type, input type and name, and output type and name.

src/main/java/net/imagej/ops/OpSignature.java
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	@Override
+	public int hashCode() {
+		return Objects.hash(buildShortSig());
```
gselzer, line 102: This should hash on those objects.

src/main/java/net/imagej/ops/OpSignature.java
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	@Override
+	public String toString() {
+		return buildShortSig();
```
gselzer, line 107: `buildShortSig` can just go in here.

src/main/java/net/imagej/ops/OpSignature.java
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	// -- Helper methods --
+
+	public String buildShortSig() {
```
gselzer, line 112: Can we add documentation for these methods? Really, all methods in this class should have documentation.

src/main/java/net/imagej/ops/OpSignature.java
```diff
@@ -0,0 +1,188 @@
...
+	private String paramList(Iterable<ModuleItem<?>> items, String beforeFirst) {
+		boolean first = true;
+		StringBuilder sb = new StringBuilder();
+		for (ModuleItem<?> item : items) {
+			if (first) {
+				sb.append(beforeFirst);
+				first = false;
+			}
+			else sb.append(", ");
+			sb.append(shortParam(item, true));
```
gselzer, line 135: Remove `true`, just hardcode that variable in `shortParam`.

We can probably simplify `shortParam` if we do this.

src/main/java/net/imagej/ops/OpSignature.java
```diff
@@ -0,0 +1,188 @@
...
+		}
+		// FooBar -> fooBar
+		return s.substring(0, 1).toLowerCase() + s.substring(1);
+	}
+
+	private boolean isUpperCase(char c) {
+		return c >= 'A' && c <= 'Z';
+	}
+
+	private static final Map<Class<?>, String> SIG_SIMPLIFICATIONS =
```
gselzer, line 175: This should be extensible.

src/main/java/net/imagej/ops/OpSignatureInfo.java
```diff
@@ -0,0 +1,72 @@
...
+ * @author Gabriel Selzer
+ */
+public class OpSignatureInfo extends AbstractModuleInfo {
+
+	private final OpService opService;
+
+	private final OpSignature signature;
+
+	/**
+	 * Defines an OpSignatureInfo with
```
gselzer, line 25: Finish this.

src/main/java/net/imagej/ops/OpSignature.java
```diff
@@ -0,0 +1,188 @@
...
+		this.functionalType = functionalType.getName();
+		this.inputs = inputs;
+		this.outputs = outputs;
+	}
+
+	public String getName() {
+		return name;
+	}
+
+	public String getFunctionalType() {
```
gselzer, line 51: This seems to always be `Op.class`, not e.g. `UnaryFunctionOp` :frowning:

src/main/java/net/imagej/ops/OpSignatureModule.java
```diff
@@ -0,0 +1,62 @@
...
+ * {@link OpSignatureInfo}.
+ * 
+ * @author Gabriel Selzer
+ */
+class OpSignatureModule extends AbstractModule {
+
+	private final OpService opService;
+	private final OpSignatureInfo info;
+
+	public OpSignatureModule(OpService opService, OpSignatureInfo info) {
```
gselzer, line 24: Maybe call `super()`?

src/main/java/net/imagej/ops/OpSignature.java
gselzer: We no longer use the `List<ModuleItem<?>>`s for input and output.

src/main/java/net/imagej/ops/OpSignature.java
gselzer: Thanks, fixed.

src/main/java/net/imagej/ops/OpSignature.java
gselzer: Fixed.

src/main/java/net/imagej/ops/OpSignature.java
gselzer: Fixed.

src/main/java/net/imagej/ops/OpSignature.java
gselzer: Yup!

Fixed.

src/main/java/net/imagej/ops/OpSignature.java
gselzer: Better documented now.

src/main/java/net/imagej/ops/OpSignature.java
gselzer: Fixed.

src/main/java/net/imagej/ops/OpSignature.java
gselzer: Now it is!

src/main/java/net/imagej/ops/OpSignatureInfo.java
gselzer: Finished!

src/main/java/net/imagej/ops/OpSignature.java
gselzer: I don't think this is an issue.

src/main/java/net/imagej/ops/OpSignatureModule.java
gselzer: Done!

gselzer:
@ctrueden do you have time to review this?

I think it's pretty clean, but would want to do more testing and get feedback before a merge.

ctrueden:
> do you have time to review this?

I put it on my list for Wednesday. Is that soon enough?

gselzer:
I was thinking about this over the weekend, and I wanted to record my thoughts:

`OpSignature` feels like the key to exposing/learning about Ops' functionality. It should be treated as a hint to the user saying "hey, I know there's an Op that looks like this!" Users should then build an `OpRef` aligning with a particular `OpSignature`, and that `OpRef` then gets matched to the `OpInfo` most tailored to the inputs.

Given `OpSignature`, though, one cannot currently assume there is an Op with the input types of the `OpSignature`. This is due to the way that inputs are currently folded, via `OpSignature.reduce()`. The goal of that method is to:
1. Combine similar signatures
2. Describe the input and output types in a way that is coherent for users.

E.g. let's say you have two Ops:
* A `foo.bar` implementation `Function, DoubleType>`
* A `foo.bar` implementation `Function, LongType>`
The user will only care about the fact that Ops has the functionality `foo.bar(image, number)`; this is what we want to display to users, and `OpSignature` allows us to do it! **But** if you pass a `Img<LongType>` then you won't get a match (until ImageJ Ops2, at least), so you might come away thinking the `OpSignature` lied to you... :disappointed: But we don't have to reduce.

# The difference between `OpInfo`, `OpRef`, and `OpSignature`:
* `OpInfo`s describe an Op implementation (i.e. a `Class`).
* `OpRef`s describe a user's request for an Op *with* concrete input objects.
* `OpSignature`s are an attempt to describe *many* `OpInfo`s via some *reduction* of the input and output types.

gselzer:
> The only thing that makes me nervous is the name `OpSignature`

I'm happy to debate names!

> The term "signature" is usually used in computing to mean something for validating identity.

Sure, that meaning exists, but I think the meaning of inputs and outputs to a function is more common. [This Wikipedia page](https://en.wikipedia.org/wiki/Type_signature) is the first result I see when searching "computer science signature".

> I kind of like `OpApproximation` for its honesty. Or maybe `OpStandardization`, since the point is to standardize each argument to some baseline agreed-upon pseudo-types, right?

I prefer the latter, since you are right about the point of this work. But approximation and standardization also have their own meanings. 

I picked @elevans' brain, and he suggested `OpProxy`, which I like.

gselzer:
Another name I like is `OpListing`, if we are viewing this thing as the thing that is presented to users

ctrueden:
Ooh yeah, `OpListing` is :+1: from me.

gselzer:
Here's another question I've been struggling with:

Should the `OpListing` contain parameter names? I can think of reasons for and against:
### For:
* It allows us to make the `OpListing` string much more informative. `filter.frangiVesselness(image, image, array "spacing", number "scale")` is much more informative than `filter.frangiVesselness(image, image, array, number)`, which might appear overwhelming.

### Against:
* The names cannot factor into matching, so they probably shouldn't factor into `OpListing` equality. The goal of `OpListing` is to reduce different implementations with the "same signature" down to one result, and different parameter names do not make a "different signature"
* If we reduce listings with *different parameter names* which `OpListing` gets its parameters chosen?

src/main/java/net/imagej/ops/OpListing.java
```diff
@@ -0,0 +1,348 @@
...
+	 *
+	 * @param name the name of the Op.
+	 * @param functionalType the basic type of the Op (e.g.
+	 *          {@link net.imagej.ops.special.function.UnaryFunctionOp})
+	 * @param inputNames the input names of the Op
+	 * @param inputTypes the input types of the Op
+	 * @param returnName the return name of the Op
+	 * @param returnType the return type of the Op
+	 */
+	public OpListing( //
```
ctrueden, line 124: The only code calling this constructor is in this class, so maybe we shouldn't expose this publicly yet.

src/main/java/net/imagej/ops/OpListing.java
```diff
@@ -0,0 +1,348 @@
...
+		// (pure) output
+		if (returnName.isPresent() && returnType.isPresent()) {
+			final String returnString = shortParam(returnName.get(), returnType
+				.get());
+			sb.append(" -> ").append(returnString);
+		}
+		return sb.toString();
+	}
+
+	public OpListing reduce(final Function<Type, Type> typeReducer) {
```
ctrueden, line 266: I don't have a clear vision for how this will be used, externally. I understand that it's an attempt to make it possible to generalize the `OpListing` logic across more types than those hardcoded by `SIG_SIMPLIFICATIONS`, but in practice, there needs to be a way for napari-imagej to tell the `OpService` that it wants to use some alternative simplifications all the time when searching, right? This seems like it should be global state to the `OpService`. And then the `reduce` guts here could stay private.

A more general way to deal with this, without custom functions, would be for the `typeReducer` to lean on the `ConvertService`. If a type can be converted to e.g. `RAI`, then it is reduced to the string `"image"`. If it can be converted to `Number`, then it is reduced to the string `"number"`. Or maybe we don't want to literally use strings as pseudo-types, we can still pick some canonical Java type for each of image and number perhaps, since I imagine this would be important for the `OpListingModule`.

hinerm:
@gselzer does this PR actually work for you in Fiji?

Here's what I get when I search for something:
image

gselzer:
> @gselzer does this PR actually work for you in Fiji?
> 
> Here's what I get when I search for something: image

Dang, I see that in Ops too now, @hinerm. I'm guessing it's actually an issue in that `Types` utility class, but I'll see what I can do to mitigate it.

gselzer:
Rewriting [this line](https://github.com/imagej/imagej-ops/blob/56a5a29a096b4c0596cd1142afc41972e6e52f52/src/main/java/net/imagej/ops/search/OpSearcher.java#L68) to be `.filter(e -> Types.isAssignable(Types.raw(type), e.getKey()))` fixes that issue. This is likely the correct fix.

But I was correct that this is a deeper issue within `Types`. It turns out that [`CaptureTypeImpl`](https://github.com/scijava/scijava-common/blob/5d8c6ca9e3d3e82faa0644622af6f3d6f502e787/src/main/java/org/scijava/util/Types.java#L3685)s are propagating into the `Types.isAssignable` checks in the link above, due to the way that the `CommandModuleItem` determines the parameterized type of the inputs. `CaptureTypeImpl`s have no support in `Types.isAssignable`, leading to the throw `IllegalStateException` [here](https://github.com/scijava/scijava-common/blob/5d8c6ca9e3d3e82faa0644622af6f3d6f502e787/src/main/java/org/scijava/util/Types.java#L1303).

Do you agree that the `Types.raw()` addition is correct, @hinerm?

src/main/java/net/imagej/ops/OpListingModule.java
```diff
@@ -0,0 +1,91 @@
...
+	public ModuleInfo getInfo() {
+		return info;
+	}
+
+	@Override
+	public void run() {
+		Object[] args = opArgs();
+		Object out = this.opService.run(info.getName(), args);
+		for (ModuleItem<?> item : info.outputs()) {
+			// TODO: What if there are multiple pure outputs?
```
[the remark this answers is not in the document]
gselzer: Ugh, things like this make me appreciate SciJava Ops.

Here's my plan:
1. Count the number of *pure* outputs; we *should* be able to get this from the `OpListingInfo`
**If there are 2+**:
2. Throw an IAE if `out` is not a `List`.
3. Assign each in order from the `List` to the `output`s.

That doesn't seem so bad...

src/main/java/net/imagej/ops/search/OpSearcher.java
```diff
@@ -52,58 +63,121 @@
 @Plugin(type = Searcher.class, priority = Priority.VERY_HIGH - 1)
 public class OpSearcher implements Searcher {
 
-	@Parameter
-	private OpService opService;
+	private Function<Type, Type> paramReducer = (type) -> {
+		Class<?> raw = Types.raw(type);
+		Optional<Type> simpleName = SIG_SIMPLIFICATIONS.entrySet() //
+				.parallelStream() //
```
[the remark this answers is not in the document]
gselzer: Thanks!

src/main/java/net/imagej/ops/search/OpSearcher.java
```diff
@@ -52,58 +63,121 @@
...
+		this.paramReducer = func;
+	}
 
-		final LinkedHashSet<OpInfo> matches = new LinkedHashSet<>();
+	// -- Helper methods --
 
-		// Get the list of all ops for consideration.
-		final List<OpInfo> ops = opService.infos().stream() //
-			.collect(Collectors.toList());
+	private int sortByName(String name1, String name2, String filterText) {
```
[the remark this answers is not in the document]
gselzer: Thanks!

src/main/java/net/imagej/ops/search/OpSearcher.java
```diff
@@ -52,58 +63,121 @@
...
 
-		final String baseDir = //
-			appService.getApp().getBaseDirectory().getAbsolutePath();
+		// Update nameSets
+		updateNameSets();
+
+		// Find all names matching text
+		String filterText = filterString(text);
+
+		return opsOfName.keySet().parallelStream() //
```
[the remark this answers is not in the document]
gselzer: Thanks!

src/main/java/net/imagej/ops/OpListing.java
```diff
@@ -0,0 +1,358 @@
...
+ * A data structure designed to capture an abstract algorithm. As far as ImageJ
+ * Ops is concerned, Ops that have the same {@link OpListing} are
+ * <em>different implementations of the same algorithm</em>, much as method
+ * overloading in Java might be considered different implementations of the same
+ * functionality.
+ * <p>
+ * If you have <b>concrete Objects</b> instead of <b>types</b>, use
+ * {@link OpRef}.
+ *
+ * @author Curtis Rueden
```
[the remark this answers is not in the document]
gselzer: Woah, I didn't realize it was *just* you! I coulda sworn I added myself...

src/main/java/net/imagej/ops/OpInfo.java
```diff
@@ -160,6 +160,10 @@ public Class<? extends Op> getType() {
 		return opType;
 	}
 
+	public OpListing listing() {
+		return new OpListing(this);
```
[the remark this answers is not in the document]
gselzer: Ehh, good question.

I could see different implementations wanting to override how they get reduced? Maybe we have an `OpListing` that would only want to reduce its pure inputs, but *not* its outputs?

But I don't have a great use case.

hinerm:
> Do you agree that the `Types.raw()` addition is correct, @hinerm?

@gselzer this seems reasonable to me!

>CaptureTypeImpls have no support in Types.isAssignable, leading to the throw IllegalStateException [here](https://github.com/scijava/scijava-common/blob/5d8c6ca9e3d3e82faa0644622af6f3d6f502e787/src/main/java/org/scijava/util/Types.java#L1303).

OK.. I'm assuming `CaptureType` is truly necessary despite the presence of [`WildcardType`](https://docs.oracle.com/javase/8/docs/api/javax/lang/model/type/WildcardType.html). If it doesn't already exist there should be an issue in SJC to add support in `Types.isAssignable`.

src/main/java/net/imagej/ops/OpInfo.java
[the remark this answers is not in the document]
ctrueden: OK, it's fine to leave as is then.

gselzer:
@hinerm I was actually just about to push them :sweat_smile:

Just making sure that the tests pass with the changes

gselzer:
@hinerm can you see if you prefer this change in Fiji?

hinerm:
@gselzer the search results are looking good!

But.. they may not actually be functional? For example, if I run `math.add(number, number)` nothing happens.
image

I expected a particular `PrimitiveMath.IntegerAdd` signature to "win" and run (e.g. manually running `add(int, int)`):
image

gselzer:
> For example, if I run `math.add(number, number)` nothing happens.

The issue is that we are getting stuck (silently :frowning) in input harvesting.

Specifically, the `AbstractInputHarvester` is trying to create a `org.scijava.widget.DefaultWidgetModel`, which fails [here](https://github.com/scijava/scijava-common/blob/5d8c6ca9e3d3e82faa0644622af6f3d6f502e787/src/main/java/org/scijava/widget/DefaultWidgetModel.java#L95) since the `DefaultModuleService` cannot convert [`"0"` into a `Number`](https://github.com/scijava/scijava-common/blob/5d8c6ca9e3d3e82faa0644622af6f3d6f502e787/src/main/java/org/scijava/module/DefaultModuleService.java#L332). Previously, we were converting `"0"` into some primitive type, which the `ConvertService` must have been able to do.

Thus we should probably make the fix in scijava-common.

Unfortunately, I'm guessing that this isn't an isolated scenario and that we will see similar errors pop up.

gselzer:
@hinerm so `math.add` *runs* now :sweat_smile: 

The issue is that scijava-common had *no* support for `Number`, only for its subclasses. See scijava/scijava-common#441 for the fix that allows us to run that result. You can build it and install it in Fiji and things "work"

Unfortunately, the matching is wonky. I entered in "a=63.5" and "b=40" and got a result of "103". After some digging I found out that this seems to be a matching bug, as the Op `IntegerAdd` is being called :unamused:

*But* all of this work really seems beyond the scope of this PR; it's all bugs in upstream libraries.

I'm going to take another look at this PR tomorrow and see if we can't figure out any other issues with this code (or if there are any I missed, as I saw @ctrueden had some suggestions above)

src/main/java/net/imagej/ops/OpListingModule.java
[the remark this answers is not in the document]
gselzer: Fixed and tested in 3a5a40645

src/main/java/net/imagej/ops/OpListing.java
```diff
@@ -0,0 +1,348 @@
...
+	/**
+	 * Convenience {@link OpListing} constructor for Ops without returns.
+	 *
+	 * @param name the name of the Op.
+	 * @param functionalType the basic type of the Op (e.g.
+	 *          {@link net.imagej.ops.special.function.UnaryFunctionOp})
+	 * @param inputNames the input names of the Op
+	 * @param inputTypes the input types of the Op
+	 */
+	public OpListing( //
```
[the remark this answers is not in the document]
gselzer: I get rid of this one in 84bfb2f52, we don't use it anymore

src/main/java/net/imagej/ops/OpListing.java
gselzer: Made private in 64aae77d1. I'd actually prefer that people use this constructor, but we can always make it public later.

src/main/java/net/imagej/ops/OpListing.java
gselzer: I think the `ConvertService` will be over-eager here. We don't have an `Object` here, only `Type`s. Thus we'll have `String` inputs/outputs be reduced to `Img`s, and we'd have to increase complexity to make that work. I'd say we leave it for now, unless you have more thoughts/ideas.

hinerm:
>But all of this work really seems beyond the scope of this PR; it's all bugs in upstream libraries.

@gselzer I agree. Things look good and I also merged https://github.com/scijava/scijava-common/pull/441

>the Op IntegerAdd is being called for priority reasons 

My first intuition is that we need to decouple parameter harvesting from actually running the Op. So if the user picks `math.add(number, number)` we need to harvest two `Number` params and then use those to match and run the Op.

gselzer:
> My first intuition is that we need to decouple parameter harvesting from actually running the Op. So if the user picks `math.add(number, number)` we need to harvest two `Number` params and then use those to match and run the Op.

Yup, that's what happens (it happens [here](https://github.com/imagej/imagej-ops/pull/644/files#diff-bafcd7fa07caa3743c4fb5fa33bff5b668cd9b52cf501ab452ac85d801809b64R65-R70))

hinerm:
>Yup, that's what happens 

Annoying! OK, then I'd say don't worry about it for now

Issue #645, opened 2022-09-01
OpListings ignore preallocated outputs


gselzer:
There are issues with the way I wrote `OpListing`s. Shocker. 💥

imagej/napari-imagej#96 is a manifestation of one flaw of the current implementation: it does not retain any information about `ItemIO.BOTH` `ModuleItem`s. This is made clear by the [state](https://github.com/imagej/imagej-ops/blob/92837a806a9d1c61f76a74b7e5c1cd472bce6fb8/src/main/java/net/imagej/ops/OpListing.java#L62) held by the `OpListing`; we know the Op's inputs, and we know it's returns, but we *don't* know *what gets mutated along the way*. This limits the capabilities of the `OpListing`, and prevents us from accessing some functionality.

What I'd propose is instead to refactor the `OpListing` state to the following information:
    * Op name
    * Op functional type
    * Parameter names
    * Parameter types
    * Parameter I/O type
    * Parameter optionality
This will allow us to retain that information, and removes the need for a separate list of output types.

We may then also need to improve [filtering](https://github.com/imagej/imagej-ops/blob/92837a806a9d1c61f76a74b7e5c1cd472bce6fb8/src/main/java/net/imagej/ops/search/OpSearcher.java#L107) step of `OpSearcher.search()` to allow for more corner cases. One case I had in mind might occur when two Ops share the same name, *reduced* pure inputs, and *reduced* outputs, but one is a `Computer` and the other is a `Function`. E.g. maybe for some number type `T` you have

`BinaryComputer<ArrayImg<T> ,T, RAI<T>> math.add`
`BinaryFunction<CellImg<T>, T, RAI<T>> math.add` 

In that case, I'd want to see something like `math.add(img, number) -> img` for both resulting *reduced* `OpListing`s. Currently, we'd get two listings, as the current logic would suggest that the `Computer` has *no* return type, while the `Function` does, and thus the `distinct()` check in `OpSearcher.search()` would leave us with two `OpListings` :frowning:

But, we could get around this by being smarter, with the refactored state, and a better filter. All we really need to check to reduce N `OpListing`s to one is to ensure that the first 4 of those new variables are the same; some idea of "union"s could allow for discrepancies in the last two variables to reduce to one `OpListing`.

I'll try writing a fix to the logic soon...

PR #647, opened 2022-09-16
Various OpListing Improvements

gselzer:
This PR was originally designed to address imagej/imagej-ops#645, but breaks from the plan addressed in that PR a bit.

This PR does include the `OpListing` refactoring described in that issue, which is part of what we need for imagej/napari-imagej#96 (we'll need a release of imagej-ops, plus a PR in that repository to actually address the issue). Specifically, I wrote the *private* class `OpListing.Parameter` to keep track of all the needed state. This class also makes it significantly easier to make improvements to the stringification of each parameter.

The other improvement made by this PR is the re-inclusion of *all* parameter names. Previously, we were leaving out any names conforming to either `"out"` or `"inX"`, because I thought they were redundant. It turns out that this got confusing, so I added them back. I actually tried making a bunch of different modifications to the `OpListing` strings (see [this zulip thread](https://imagesc.zulipchat.com/#narrow/stream/327236-ImageJ2/topic/OpListing.20indicator.20for.20Mutable.20Outputs/near/298647346)), but for now I'm going to leave them out, and add them in later if people need (thanks for the suggestion @hinerm)

I'm not sure this fully solves #645, as I did not include the additional `OpListing` filtering I talked about in that PR. *But*, if we do it, I'd rather address it in another PR as it is a separate issue.

### DIFF
--- a/src/main/java/net/imagej/ops/OpListing.java
+++ b/src/main/java/net/imagej/ops/OpListing.java
@@ -31,17 +31,16 @@ package net.imagej.ops;
 
 import java.lang.reflect.Type;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 import java.util.function.Function;
 import java.util.stream.Collectors;
-import java.util.stream.IntStream;
 
 import org.scijava.ItemIO;
 import org.scijava.module.DefaultMutableModuleItem;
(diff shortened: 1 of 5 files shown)


hinerm:
@gselzer is this currently a draft PR?

gselzer:
@hinerm nope! Ready for review!

hinerm (approved):

PR #654, opened 2024-04-05
Addition of getType() API 

tpietzsch:
Changes that will be required after imglib/imglib2#312 is merged and released

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -256,6 +256,13 @@
 		<!-- NB: Deploy releases to the SciJava Maven repository. -->
 		<releaseProfiles>sign,deploy-to-scijava</releaseProfiles>
 
+		<imglib2.version>7.0.3</imglib2.version>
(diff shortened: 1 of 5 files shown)


ctrueden:
Remaining compilation error:
```
[ERROR] ...imagej-ops/src/main/java/net/imagej/ops/labeling/MergeLabeling.java:[95,81] reference to sample is ambiguous
  both method <T>sample(net.imglib2.IterableInterval<java.lang.Void>,net.imglib2.RandomAccessible<T>) in net.imglib2.roi.Regions and method <T>sample(net.imglib2.roi.IterableRegion<?>,net.imglib2.RandomAccessible<T>) in net.imglib2.roi.Regions match
```
I will fix it after I return from vacation, if no one else has done so already!

tpietzsch:
I fixed the remaining compile error.

This is ready to go as soon as https://github.com/imagej/imagej-common/pull/112 is merged and released.

`imagej-common` is the only SNAPSHOT dependency left

PR #657, opened 2026-02-13
Overhaul of Voxelization ops with new algorithm. Also included test

andmccall:
Wanted to address a major issue with the voxelization Op not working appropriately. The old algorithm was returning every pixel of every triangle's entire bounding box as true. This new method only sets the pixels that the surface goes through to true.

In general, this method should now be more symmetric with marching cubes, so that after:
```
mesh = ops.geom().marchingCubes(image);
result = ops.geom().voxelization(mesh, image);
ops.morphology().fillHoles(result, result, new DiamondShape(1));
```
Image and result should match in the majority of cases. The notable exception would be images with holes themselves, generating a surface encompassed entirely by a surface in marching cubes. Voxelization would generate both surfaces, but I believe all encompassed holes would be filled after the fillHoles call, though I haven't tested this yet.

I did add some comments in the code for some optional changes that could be made. Originally I had an additional parameter for the wall thickness, but wasn't sure if this was really beneficial. I also wasn't sure if I should be using another Op (fill holes) during the voxelization Op testing, so I left it out and just compared the result to a static number.

Edit: was just looking through and noticed I forgot to delete the LogService parameter, which I was using for troubleshooting an issue.

### DIFF
--- a/src/main/java/net/imagej/ops/geom/GeomNamespace.java
+++ b/src/main/java/net/imagej/ops/geom/GeomNamespace.java
@@ -691,6 +691,28 @@ public class GeomNamespace extends AbstractNamespace {
 		return result;
 	}
 
+	//Two OpMethods below cause a Mismatched inputs error, though they both work fine if built with 'Skip tests'
+//	@OpMethod(op = net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class)
+//	public RandomAccessibleInterval<BitType> voxelization(final Mesh in, double wallThickness) {
+//		final RandomAccessibleInterval<BitType> result =
+//				(RandomAccessibleInterval<BitType>) ops().run(net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class, in, null, false, wallThickness);
+//		return result;
+//	}
+//
+//	@OpMethod(op = net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class)
+//	public RandomAccessibleInterval<BitType> voxelization(final Mesh in, final Interval dimensions, double wallThickness) {
+//		final RandomAccessibleInterval<BitType> result =
+//				(RandomAccessibleInterval<BitType>) ops().run(net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class, in, dimensions, false, wallThickness);
+//		return result;
+//	}
+
+	@OpMethod(op = net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class)
+	public RandomAccessibleInterval<BitType> voxelization(final Mesh in, final Interval dimensions, boolean scaleMeshToDimesions, double wallThickness) {
+		final RandomAccessibleInterval<BitType> result =
+				(RandomAccessibleInterval<BitType>) ops().run(net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class, in, dimensions, scaleMeshToDimesions, wallThickness);
+		return result;
+	}
+
 	
 	@OpMethod(op = net.imagej.ops.geom.geom2d.DefaultVerticesCountPolygon.class)
 	public DoubleType verticesCount(final Polygon2D in) {
--- a/src/main/java/net/imagej/ops/geom/GeomNamespace.java
+++ b/src/main/java/net/imagej/ops/geom/GeomNamespace.java
@@ -37,6 +37,7 @@ import net.imagej.ops.Namespace;
 import net.imagej.ops.OpMethod;
 import net.imagej.ops.Ops.Geometric.Voxelization;
 import net.imagej.ops.geom.geom3d.mesh.VertexInterpolator;
+import net.imglib2.Interval;
 import net.imglib2.IterableInterval;
 import net.imglib2.RandomAccessibleInterval;
 import net.imglib2.RealLocalizable;
@@ -278,13 +279,6 @@ public class GeomNamespace extends AbstractNamespace {
 			net.imagej.ops.Ops.Geometric.ConvexHull.class, in);
 		return result;
 	}
-	
-	@OpMethod(op = net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class)
-	public RandomAccessibleInterval<BitType> voxelization(final Mesh in, final int width, final int height, final int depth ) {
-		final RandomAccessibleInterval<BitType> result = (RandomAccessibleInterval<BitType>) ops().run(
-				Voxelization.class, in, width, height, depth );
-	 	return result;
-	}	
 
 	@OpMethod(op = net.imagej.ops.geom.geom2d.DefaultConvexityPolygon.class)
 	public DoubleType convexity(final Polygon2D in) {
@@ -678,26 +672,25 @@ public class GeomNamespace extends AbstractNamespace {
 	
 	@OpMethod(op = net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class)
 	public RandomAccessibleInterval<BitType> voxelization(final Mesh in) {
-		@SuppressWarnings("unchecked")
 		final RandomAccessibleInterval<BitType> result =
 			(RandomAccessibleInterval<BitType>) ops().run(net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class, in);
 		return result;
 	}
 
 	@OpMethod(op = net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class)
-	public RandomAccessibleInterval<BitType> voxelization(final Mesh in, final int width) {
-		@SuppressWarnings("unchecked")
+	public RandomAccessibleInterval<BitType> voxelization(final Mesh in, final Interval dimensions) {
 		final RandomAccessibleInterval<BitType> result =
-			(RandomAccessibleInterval<BitType>) ops().run(net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class, in, width);
+			(RandomAccessibleInterval<BitType>) ops().run(net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class, in, dimensions);
 		return result;
 	}
 
 	@OpMethod(op = net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class)
-	public RandomAccessibleInterval<BitType> voxelization(final Mesh in, final int width, final int height) {
+	public RandomAccessibleInterval<BitType> voxelization(final Mesh in, final Interval dimensions, boolean scaleMeshToDimesions) {
 		final RandomAccessibleInterval<BitType> result =
-			(RandomAccessibleInterval<BitType>) ops().run(net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class, in, width, height);
+				(RandomAccessibleInterval<BitType>) ops().run(net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class, in, dimensions, scaleMeshToDimesions);
 		return result;
 	}
+
 	
 	@OpMethod(op = net.imagej.ops.geom.geom2d.DefaultVerticesCountPolygon.class)
 	public DoubleType verticesCount(final Polygon2D in) {
--- a/src/main/java/net/imagej/ops/geom/geom3d/DefaultVoxelization3D.java
+++ b/src/main/java/net/imagej/ops/geom/geom3d/DefaultVoxelization3D.java
@@ -67,9 +67,6 @@ public class DefaultVoxelization3D extends AbstractUnaryFunctionOp<Mesh, RandomA
 	@Parameter
 	private OpService ops;
 
-	@Parameter
-	private LogService logService;
-
 	private final double wallThickness = 1.0; //This could be made into a parameter if needed
 	private final long[] offset = new long[] {0,0,0};
 	private double scale = 1.0;
@@ -81,7 +78,7 @@ public class DefaultVoxelization3D extends AbstractUnaryFunctionOp<Mesh, RandomA
 			float[] bounds = Meshes.boundingBox(input);
 			long[] outputInterval = new long[3];
 			for (int i = 0; i < 3; i++) {
-				outputInterval[i] = (long)Math.ceil(bounds[i+3]-bounds[i]);
+				outputInterval[i] = (long)Math.ceil(bounds[i+3]-bounds[i])+1;
 			}
 			dimensions = new FinalInterval(outputInterval);
 			setScale(input, bounds);
--- a/src/main/java/net/imagej/ops/geom/geom3d/DefaultVoxelization3D.java
+++ b/src/main/java/net/imagej/ops/geom/geom3d/DefaultVoxelization3D.java
@@ -78,7 +78,7 @@ public class DefaultVoxelization3D extends AbstractUnaryFunctionOp<Mesh, RandomA
 			float[] bounds = Meshes.boundingBox(input);
 			long[] outputInterval = new long[3];
 			for (int i = 0; i < 3; i++) {
-				outputInterval[i] = (long)Math.ceil(bounds[i+3]-bounds[i])+2;
+				outputInterval[i] = (long)Math.ceil(bounds[i+3]-bounds[i]);
 			}
 			dimensions = new FinalInterval(outputInterval);
 			setScale(input, bounds);
@@ -128,7 +128,7 @@ public class DefaultVoxelization3D extends AbstractUnaryFunctionOp<Mesh, RandomA
 		double[] axisScaling = new double[3];
 		for (int i = 0; i < 3; i++) {
 			offset[i] = (long) Math.floor(bounds[i]) - dimensions.min(i);
-			axisScaling[i] = (dimensions.max(i) - dimensions.min(i)) / (bounds[i + 3] - bounds[i]);//dimensionRange/boundingboxRange
+			axisScaling[i] = (dimensions.max(i) - dimensions.min(i)) / ((bounds[i + 3]+2) - bounds[i]);
 		}
 		scale = Math.min(axisScaling[0], Math.min(axisScaling[1], axisScaling[2]));
 	}
--- a/src/main/java/net/imagej/ops/geom/geom3d/DefaultVoxelization3D.java
+++ b/src/main/java/net/imagej/ops/geom/geom3d/DefaultVoxelization3D.java
@@ -29,21 +29,20 @@
 package net.imagej.ops.geom.geom3d;
 
 import net.imagej.mesh.Mesh;
+import net.imagej.mesh.Meshes;
 import net.imagej.mesh.Triangle;
-import net.imagej.mesh.Vertices;
 import net.imagej.ops.OpService;
 import net.imagej.ops.Ops;
 import net.imagej.ops.special.function.AbstractUnaryFunctionOp;
-import net.imglib2.FinalInterval;
-import net.imglib2.RandomAccess;
-import net.imglib2.RandomAccessibleInterval;
-import net.imglib2.RealLocalizable;
-import net.imglib2.RealPoint;
+import net.imglib2.*;
 import net.imglib2.img.Img;
+import net.imglib2.iterator.LocalizingIntervalIterator;
 import net.imglib2.type.logic.BitType;
 
+import net.imglib2.util.Intervals;
 import org.apache.commons.math3.geometry.euclidean.threed.Vector3D;
 import org.scijava.ItemIO;
+import org.scijava.log.LogService;
 import org.scijava.plugin.Parameter;
 import org.scijava.plugin.Plugin;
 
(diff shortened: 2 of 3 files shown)


andmccall:
Just discovered a bug, working on fixing it and should have a new commit soon.
Edit: done. Sorry for the changes after the pull request, didn't notice these issues in initial testing.

andmccall:
@gselzer and @ctrueden,

Think I have all the kinks worked out and it's ready for review. I did find that there were a couple @OpMethods I put in the GeomNamespace that I had to comment out for the JUnit tests to pass (Input Mismatch error), even though the function calls worked fine when I compiled them without tests. Since it sounds like everything is gonna move to SciJava Ops, I figured it wasn't worthwhile for me to figure out the issue as I'm not very familiar with JUnit testing. 

Anyway, I'll for SciJava-ops and work on converting this to the SciJava-Ops format.

src/test/java/net/imagej/ops/geom/MeshFeatureTests.java
```diff
@@ -201,6 +191,13 @@ public void verticesCountMesh() {
...
-		// https://github.com/imagej/imagej-ops/issues/422
+		/*Value of 184 here corresponds with:
+		RandomAccessibleInterval<BitType> result = (RandomAccessibleInterval<BitType>) ops.run(DefaultVoxelization3D.class, mesh, ROI);
+		ops.morphology().fillHoles(result, result, new DiamondShape(1));
+		assertEquals(Ops.Geometric.Voxelization.NAME, ROI.size(), Regions.countTrue(result));
+		 */
+		assertEquals(Ops.Geometric.Voxelization.NAME, 184,
+		Regions.countTrue((RandomAccessibleInterval<BitType>) ops.run(DefaultVoxelization3D.class, mesh, ROI)));
+
 	}
```
gselzer, line 202: This comment has absolutely nothing to do with your work, since it's clear you're just copying the precedent, but man, I'm remembering how much I disliked these tests 😆 ; if the goal is to assert proper voxelization, it is of course not sufficient to count the number of `true`s - you'll also need to know where they are. 

I would love a test like this for the eventual SciJava Ops version, but up to you whether you think we should do something like this now (probably shouldn't go here though):

```suggestion
	public void voxelization3D() {
		// Create a "square" mesh from [1, 1, 1] to [2, 2, 1]
		Mesh m = new NaiveDoubleMesh();
		m.vertices().add(1, 1, 1);
		m.vertices().add(1, 2, 1);
		m.vertices().add(2, 1, 1);
		m.vertices().add(2, 2, 1);
		m.triangles().add(0, 1, 2);
		m.triangles().add(2, 3, 1);

		// Voxelize it
		Interval interval = new FinalInterval(new long[] {0, 0, 0}, new long[] {3, 3, 3});
		RandomAccessibleInterval<BitType> voxelized = (RandomAccessibleInterval<BitType>) ops.run(DefaultVoxelization3D.class, m, interval);

		// Check the results
		Cursor<BitType> cursor = voxelized.cursor();
		while (cursor.hasNext()) {
			boolean value = cursor.next().get();
			long[] pos = cursor.positionAsLongArray();
			if (pos[0] < 1 || pos[0] > 2) {
				// x dimension outside the square
				assertFalse(value);
			}
			else if (pos[1] < 1 || pos[1] > 2) {
				// y dimension outside the square
				assertFalse(value);
			}
			else if (pos[2] != 1) {
				// z dimension outside the square
				assertFalse(value);
			}
			else {
				// "In" the square
				assertTrue(value);
			}
		}
	}

```

src/main/java/net/imagej/ops/geom/GeomNamespace.java
```diff
@@ -678,26 +672,47 @@ public double[] vertexInterpolator(final int[] p1, final int[] p2,
...
+//		return result;
+//	}
+//
+//	@OpMethod(op = net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class)
+//	public RandomAccessibleInterval<BitType> voxelization(final Mesh in, final Interval dimensions, double wallThickness) {
+//		final RandomAccessibleInterval<BitType> result =
+//				(RandomAccessibleInterval<BitType>) ops().run(net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class, in, dimensions, false, wallThickness);
+//		return result;
+//	}
+
```
gselzer, line 708: These do not work because optional Op signatures are only "valid" if parameters are left off right to left. In this case, valid signatures would be:

* `voxelization(final Mesh in, final Interval dimensions, final boolean scaleMeshToDimensions, final double wallThickness)`
* `voxelization(final Mesh in, final Interval dimensions, final boolean scaleMeshToDimensions)`
* `voxelization(final Mesh in, final Interval dimensions)`
* `voxelization(final Mesh in)`

Of course these two signatures are not included. The context for this right-to-left dropping is to remove ambiguity when you have two optional parameters that are of the same type e.g. for some
```java
voxelization(final Mesh in, final double a, final Double b)
```
you could not have both 
```java
voxelization(final Mesh in, final double a)
voxelization(final Mesh in, final double b)
```

However, your use case does not have this problem. You could consider skipping the type check (unclear if this is what you meant by skip tests) - do you think this is an acceptable use case of that parameter @ctrueden?

```suggestion
	@OpMethod(op = net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class, skipTypeCheck = true)
	public RandomAccessibleInterval<BitType> voxelization(final Mesh in, double wallThickness) {
		final RandomAccessibleInterval<BitType> result =
				(RandomAccessibleInterval<BitType>) ops().run(net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class, in, null, false, wallThickness);
		return result;
	}

	@OpMethod(op = net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class, skipTypeCheck = true)
	public RandomAccessibleInterval<BitType> voxelization(final Mesh in, final Interval dimensions, double wallThickness) {
		final RandomAccessibleInterval<BitType> result =
				(RandomAccessibleInterval<BitType>) ops().run(net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class, in, dimensions, false, wallThickness);
		return result;
	}
```

src/main/java/net/imagej/ops/geom/GeomNamespace.java
```diff
@@ -278,13 +279,6 @@ public List convexHull(final Mesh in) {
 			net.imagej.ops.Ops.Geometric.ConvexHull.class, in);
 		return result;
 	}
-	
-	@OpMethod(op = net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class)
-	public RandomAccessibleInterval<BitType> voxelization(final Mesh in, final int width, final int height, final int depth ) {
-		final RandomAccessibleInterval<BitType> result = (RandomAccessibleInterval<BitType>) ops().run(
-				Voxelization.class, in, width, height, depth );
-	 	return result;
-	}
```
gselzer, line 287: This is a breaking change 🛠️ 

Since your Op has no overlap in parameter types with the existing Op, we could consider offering the two Ops side-by-side? Could also consider some sort of deprecation for the old Op...

src/main/java/net/imagej/ops/geom/geom3d/DefaultVoxelization3D.java
```diff
@@ -29,376 +29,214 @@
...
 @Plugin(type = Ops.Geometric.Voxelization.class)
 public class DefaultVoxelization3D extends AbstractUnaryFunctionOp<Mesh, RandomAccessibleInterval<BitType>>
 		implements Ops.Geometric.Voxelization {
 
 	@Parameter(type = ItemIO.INPUT, required = false)
-	private int width = 10;
-
-	@Parameter(type = ItemIO.INPUT, required = false)
-	private int height = 10;
+	private Interval dimensions;
```
gselzer, line 61: Let's think about the tasks a user might need to accomplish using this Op. I can see the user wanting to:
* Obtain a voxelized image across the entire mesh bounding box
* Obtain a voxelized image across some custom interval (e.g. if the mesh is massive maybe they don't want to voxelize the whole thing).

Unless you can think of additional tasks, I think that this collection of goals would map nicely to a `UnaryHybridCF` Op, instead of a `UnaryFunctionOp`. This would allow users to call the Op with any preallocated output image defined on whatever interval they want, which would make the second goal easy. Additionally, if they called the Op as a Function, then you could use the logic you have for the bounding box to create a new image over the entire mesh interval.

What do you think about restructuring the Op to be a `UnaryHybridCF`? Do you think you could get rid of the `Interval` parameter that way?

(As an aside, in SciJava Ops hybrid Ops aren't a thing, we'd just write two separate Ops for that. But we'll cross that bridge later 😄)

src/main/java/net/imagej/ops/geom/geom3d/DefaultVoxelization3D.java
```diff
@@ -29,376 +29,214 @@
...
 
 	@Parameter(type = ItemIO.INPUT, required = false)
-	private int depth = 10;
+	private boolean scaleMeshToDimesions = false;
```
gselzer, line 64: You said [here](https://forum.image.sc/t/new-voxelization-algorithm-for-imagej-ops/119292/3?u=gselzer):

>Additionally, if the mesh wasn't previously an image the vertex coordinates in the mesh could be negative, which doesn't work very well with converting to images. Also, it just wasn't that hard to program.

If the goal is to describe the voxels a mesh passes through, setting this parameter to `true` would almost certainly obfuscate that information, since the scale and the offset are not reflected in the result.

What do you think about changing the type of the Op as I mentioned above? If that were the case, we would no longer have to worry about the scale and offset being necessary to transfer meaning between the mesh and the voxelized image...

src/main/java/net/imagej/ops/geom/geom3d/DefaultVoxelization3D.java
```diff
@@ -29,376 +29,214 @@
...
+
+
+	private Vector3D nearestPointInTriangle3D(Vector3D p, Vector3D[] t) {
+
+		Vector3D ab = t[1].subtract(t[0]);
+		Vector3D ac = t[2].subtract(t[0]);
+
+
+		//region Obtain projection (p) of origP onto plane of triangle
+		// Find the normal to the plane: n = (b - a) x (c - a)
```
gselzer, line 165: I really appreciate the comments here! This kind of context will ease future maintenance.

Would you be willing to add similar comments to the other functions as well?

src/main/java/net/imagej/ops/geom/geom3d/DefaultVoxelization3D.java
gselzer: I would also wonder whether you could get rid of the `scale` and `offset` state within the Op if you did this. One key aspect of SciJava Ops is that Ops are stateless, so it'd be great to remove any state if possible if we want to eventually port this work over there!

gselzer (commented):
@andmccall thanks so much for doing this work! It's great to see contributions!

I really love your code - I think it's a massive improvement over the existing functionality. Just some things I was thinking about as I was looking through.

(Also, for those who aren't aware, there was a bit of discussion [here](https://forum.image.sc/t/new-voxelization-algorithm-for-imagej-ops/119292?u=gselzer)

src/main/java/net/imagej/ops/geom/GeomNamespace.java
gselzer: Would be great to either get these working or delete them from the code

src/test/java/net/imagej/ops/geom/MeshFeatureTests.java
andmccall: Hi @gselzer,

Another option that I just discovered yesterday is that we can compare the result of `voxelization(mesh, originalBinary)` to the result of `net.imglib2.roi.boundary.Boundary(originalBinary, new DiamondShape(1))`. It turns out that this produces identical results for nearly all meshes. This would allow the continued use of the slightly more complex `ROI` image with the pre-calculated corresponding mesh that was present in ImageJ ops (if this carried over to SciJava Ops).

src/main/java/net/imagej/ops/geom/GeomNamespace.java
andmccall: I'm fine with it either way. Wasn't strongly invested in having these signatures, I just wasn't really sure what the issue was, or how to do the SkipTypeCheck.

src/main/java/net/imagej/ops/geom/GeomNamespace.java
andmccall: This would change the algorithm used, so I'm not sure you would want to do this, but you could do:

`final RandomAccessibleInterval<BitType> result = (RandomAccessibleInterval<BitType>) ops().run(
				Voxelization.class, in, new FinalInterval(width, height, depth));`

This means that the output from any old code would change, but the function call wouldn't. 

I imagine that is not the preferred option, but it doesn't really matter to me, can definitely keep the old algorithm in there in whatever manner you deem best. I simply forgot about the whole "can't make breaking changes" thing, never done any library coding before.

src/main/java/net/imagej/ops/geom/geom3d/DefaultVoxelization3D.java
andmccall: I'm not sure as to the difference between these 😅, or how it would work without the interval. I'll have to take a closer look, but will be quite busy until next week at the earliest. Just wanted to reply cause my work schedule tends to slingshot between totally free and a completely packed schedule, and I think it's swinging back to packed for a little while.

src/main/java/net/imagej/ops/geom/geom3d/DefaultVoxelization3D.java
andmccall: Absolutely. I love adding comments like this to math, as it takes me forever to wrap my head around this stuff and I get lost without the comments. 

I did have a related question about SciJava, as I just want to clarify that I'm understanding everything correctly. It seems like a lot of the SciJava Ops are intentionally kept to one `function`, and these functions are chained together into a `computer`. So would the ideal be to break apart this Voxelization class into functions: `ProjectionOntoPlane`?, `FindNearestPointInTriangle`, `GetDistanceToTriangle`, and maybe others; then compile all these individual functions into a Voxelization `computer`? Or would it be better to do this as an `Inplace`? I'm not really sure if the default is to create more functions, or to create functions only if there's reason to believe they'd be useful to others (and I have no idea if they would be).

gselzer: > I love adding comments like this to math, as it takes me forever to wrap my head around this stuff and I get lost without the comments.

Me too 😁 

> It seems like a lot of the SciJava Ops are intentionally kept to one `function`, and these functions are chained together into a `computer`. 

There's definitely some terminology here. `Function`s, `Computer`s, and `Inplace`s are all first-class algorithm types in SciJava Ops (and here in ImageJ Ops), but they all handle their outputs differently:
* `Function`s create a new object every time.
* `Computer`s fill up an output buffer given to them as a parameter.
* `Inplace`s *overwrite* one of their inputs with the output data (this is uncommon in our libraries but exists out in the wild).

So would the ideal be to break apart this Voxelization class into functions: `ProjectionOntoPlane`?, `FindNearestPointInTriangle`, `GetDistanceToTriangle`, and maybe others; then compile all these individual functions into a Voxelization `computer`? 

I don't think it's worth making these functions into Ops right now - we could always do that later if there's a use for others!

src/main/java/net/imagej/ops/geom/geom3d/DefaultVoxelization3D.java
gselzer: > Just wanted to reply cause my work schedule tends to slingshot between totally free and a completely packed schedule, and I think it's swinging back to packed for a little while.

I definitely understand, there is no rush!

src/main/java/net/imagej/ops/geom/GeomNamespace.java
gselzer: > This would change the algorithm used, so I'm not sure you would want to do this, but you could do:
> 
> `final RandomAccessibleInterval<BitType> result = (RandomAccessibleInterval<BitType>) ops().run( Voxelization.class, in, new FinalInterval(width, height, depth));`
> 
> This means that the output from any old code would change, but the function call wouldn't.

Yes, this option is certainly available, however I think I'd prefer avoiding breakages if possible. I wouldn't say that the previous implementation was a *bug*, it's just a different algorithm (although certainly a strange one)

> I simply forgot about the whole "can't make breaking changes" thing, never done any library coding before.

It happens to the best of us, don't worry 🙂

src/main/java/net/imagej/ops/geom/GeomNamespace.java
gselzer: I also do not feel strongly - I just want to make sure that you get the code you want in, and it's easier to add these things in later than to remove them now. So if you don't have a use let's just remove them?

andmccall: Sounds good to me.

src/test/java/net/imagej/ops/geom/MeshFeatureTests.java
andmccall: Correction: not `DiamondShape(1)`, but `Boundary.StructuringElement.FOUR_CONNECTED`.

andmccall: Tried to address this without having to create a mesh from scratch, using the `net.imglib2.roi.boundary.Boundary(originalBinary,Boundary.StructuringElement.FOUR_CONNECTED)` method I mentioned. Does show that the shape of the voxelized mesh after matches the expected shaped based on the boundary function.

src/main/java/net/imagej/ops/geom/GeomNamespace.java
andmccall: Separated out my new algorithm to defaultVoxelize3D class, instead of replacing defaulVoxelization3D. I did mark the old one as deprecated as well.

src/main/java/net/imagej/ops/geom/geom3d/DefaultVoxelization3D.java
andmccall: Had the time to go through and figure this out, and I agree, I think `UnaryHybridCF` is the way to go, so I switched the new algorithm over to that. I also got rid of scale and offset along with it as you suggested. Made the code much simpler.

src/main/java/net/imagej/ops/geom/geom3d/DefaultVoxelization3D.java
andmccall: Tried to add comments to everywhere that I thought it might help in understanding. Though as I'm writing this I realize I didn't add comments to `MeshFeaturesTest`, though I didn't see many explanatory comments in the tests outside of "verified with MatLab", so not sure if I should add some there?

andmccall:
@gselzer,
Tried to address all your points with the most recent commit, and I responded to each point individually above. Let me know what you think when you have the chance. Sorry it took so long, work has been quite busy. Finally getting some help here soon, so hopefully the SciJava Ops version won't take quite so long.

src/main/java/net/imagej/ops/geom/GeomNamespace.java
```diff
@@ -675,27 +677,51 @@ public double[] vertexInterpolator(final int[] p1, final int[] p2,
...
 	@OpMethod(op = net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class)
 	public RandomAccessibleInterval<BitType> voxelization(final Mesh in, final int width, final int height) {
 		final RandomAccessibleInterval<BitType> result =
-			(RandomAccessibleInterval<BitType>) ops().run(net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class, in, width, height);
+				(RandomAccessibleInterval<BitType>) ops().run(net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class, in, width, height);
+		return result;
+	}
+
+	@OpMethod(op = net.imagej.ops.geom.geom3d.DefaultVoxelize3D.class)
+	public RandomAccessibleInterval<BitType> voxelize(final Mesh in) {
```
gselzer: One thing I would like to avoid is a profileration of Op names. In other words, an Op's name is supposed to reference it's underlying purpose. If we can't specify a theoretical difference between, in this case, `voxelization` and `voxelize`, then those two things should share the same Op name. 

In a perfect world, I think we should (going off memory, no guarantee it works):
1. Make your new Op be of type `Ops.Geometric.Voxelization.class`, **but higher priority**
2. Add here any **new** signatures enabled by your new Op (for example, the `Mesh -> RAI<BitType>` version right below should stay)
3. For all `OpMethod` signatures that apply to both Ops, change to:

```diff
-	@Deprecated
-	@OpMethod(op = net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class)
-	public RandomAccessibleInterval<BitType> voxelization(final Mesh in) {
-		@SuppressWarnings("unchecked")
-		final RandomAccessibleInterval<BitType> result =
-				(RandomAccessibleInterval<BitType>) ops().run(net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class, in);
-		return result;
-	}
+	@OpMethod(ops = {
+			net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class,
+			net.imagej.ops.geom.geom3d.DefaultVoxelize3D.class
+	})
+	public RandomAccessibleInterval<BitType> voxelization(final Mesh in) {
+		@SuppressWarnings("unchecked")
+		final RandomAccessibleInterval<BitType> result =
+				(RandomAccessibleInterval<BitType>) ops().run(Ops.Geometric.Voxelization.class, in);
+		return result;
+	}

```

That should mean that `ops.geom().voxelization(myMesh)` should always match the new Op, but `ops.geom().voxelization(myMesh, 5, 5)` would match the old Op.

gselzer (commented):
Thanks for addressing the feedback. One more idea for you...

src/main/java/net/imagej/ops/geom/GeomNamespace.java
andmccall: Just wondering, would it be better to change the class name of the old Voxelization method to something like `DeprecatedVoxelization3D`, and make the new Voxelization method the `DefaultVoxelization3D`. Then update all the Op references to use the appropriate names? I know things are switching to SciJava ops, but I just don't want to leave ImageJ Ops messy.

gselzer: > would it be better to change the class name of the old Voxelization method to something like `DeprecatedVoxelization3D`, and make the new Voxelization method the `DefaultVoxelization3D`

I think that this would be another breaking change, right? For example, `ops().run(net.imagej.ops.geom.geom3d.DefaultVoxelization3D.class, my_mesh)` is public API returning a `RAI<BitType>`. If we did the name switch you're talking about, then a user who was calling this will now get a different result when they thought they were targeting a particular implementation *by class name*.

Maybe we can learn from this though - do you think your Op name can pertain to the particular algorithm you're using somehow? Such that if years later another person has a new algorithm, the naming conflict would be easier?

src/test/java/net/imagej/ops/geom/MeshFeatureTests.java
gselzer: To be clear, what you're saying here is that the **old** Op `DefaultVoxelization3D` produces nearly equivalent results to `net.imglib2.roi.boundary.Boundary` with `FOUR_CONNECTED`?

If so, it's unfortunate that we have a whole implementation here. Might be good to add in a `@see` tag describing the similarities between the two...

src/main/java/net/imagej/ops/geom/GeomNamespace.java
andmccall: Maybe just `DefaultVoxelization3D_EucledianDistance.class`?

src/test/java/net/imagej/ops/geom/MeshFeatureTests.java
andmccall: > To be clear, what you're saying here is that the old Op DefaultVoxelization3D produces nearly equivalent results to net.imglib2.roi.boundary.Boundary with FOUR_CONNECTED?

No, not at all. My new implementation does. Effectively:

Binary image -> Marching Cubes -> New Voxelization algorithm -> Binary image of Mesh

and

Binary image -> `Boundary` with `FOUR_CONNECTED` -> Binary image of boundary

The "Binary image of Mesh" and "Binary image of boundary" are usually* the same, which they should be, as the mesh is generated from the boundaries of the binary image. Thus we can use this to test a good voxelization algorithm. 

\* single pixels, or single pixel width lines are not converted into Mesh by marching cubes, so these are missing in "Binary image of Mesh", but not "Binary image of boundary". Thus noisy binary images will have a difference.

andmccall: To add to this, the old implementation is broken, even beyond the issues I pointed out originally on the image.sc post. As I've been testing it for this update, it routinely fails to produce good binarized images of meshes that I test. They often have either huge gaps or massive blocks where there should be a nice smooth wall. I don't think that implementation was ever even finished, as there's large chunks of code in `DefaultVoxelization3D.class` that are entirely unused, at least as far as I could figure out.

If you're interested, I can share the groovy script and data I've been using to test and compare the two.

src/main/java/net/imagej/ops/geom/GeomNamespace.java
gselzer: `EuclideanDistanceVoxelization3D`?

andmccall: Sure, I'll make the change, test and push.

src/test/java/net/imagej/ops/geom/MeshFeatureTests.java
gselzer: > Binary image -> Marching Cubes -> New Voxelization algorithm -> Binary image of Mesh
> 
> and
> 
> Binary image -> `Boundary` with `FOUR_CONNECTED` -> Binary image of boundary
> 
> The "Binary image of Mesh" and "Binary image of boundary" are usually* the same, which they should be, as the mesh is generated from the boundaries of the binary image. Thus we can use this to test a good voxelization algorithm.
> 
> * single pixels, or single pixel width lines are not converted into Mesh by marching cubes, so these are missing in "Binary image of Mesh", but not "Binary image of boundary". Thus noisy binary images will have a difference.

Aha, I see. Thanks, it's a great explanation 🏆 No `@see` needed in that case, although if you wanted an explanatory comment in the test might be insightful for later readers.

> To add to this, the old implementation is broken, even beyond the issues I pointed out originally on the image.sc post.

> If you're interested, I can share the groovy script and data I've been using to test and compare the two.

Do you think you could share it via that forum thread?🤔

andmccall: > Do you think you could share it via that forum thread?

Posted it, but just remembered that there was a bug in the `createOutput()` of my most recent commit, that makes it so ops.geom().voxelization(mesh) won't work right. I'll be pushing the next commit with the fix soon.

andmccall:
@gselzer,
Pushed the changes that we discussed, and fixed the `createOutput` bug. Let me know if you think of anything else.

andmccall:
I actually thought of a way to improve efficiency. Gonna try it real quick, shouldn't take long.

andmccall:
@gselzer,
Everything looks good to me and I think it's ready, unless you see something else? I'm gonna get started on the SciJava ops version.

src/main/java/net/imagej/ops/geom/geom3d/DefaultVoxelization3D.java
```diff
@@ -6,13 +6,13 @@
  * %%
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
- *
```
gselzer, line 9: Just to minimize changes, it looks like nothing meaningful changed in this file - can you undo the changes?

src/main/java/net/imagej/ops/geom/geom3d/EuclideanDistanceVoxelization3D.java
```diff
@@ -0,0 +1,242 @@
...
+import org.apache.commons.math3.geometry.euclidean.threed.Vector3D;
+import org.scijava.ItemIO;
+import org.scijava.Priority;
+import org.scijava.plugin.Parameter;
+import org.scijava.plugin.Plugin;
+
+/**
+ * <p>
+ * This is a voxelizer that produces a binary image with values set to true along
+ * the surface of the mesh.
```
gselzer: Any information that you can provide that encapsulates the Op's behavior (i.e. that it uses euclidean distance to populate the dataset) would be helpful for users!

src/main/java/net/imagej/ops/geom/geom3d/EuclideanDistanceVoxelization3D.java
```diff
@@ -0,0 +1,242 @@
...
+ * </p>
+ * 
+ * @author Andrew McCall (University at Buffalo)
```
gselzer, line 64: What do you think about adding a `@see` here referencing the old implementation, and noting the reason for the priority bump?

src/main/java/net/imagej/ops/geom/geom3d/EuclideanDistanceVoxelization3D.java
```diff
@@ -0,0 +1,242 @@
...
+ */
+@Plugin(type = Ops.Geometric.Voxelization.class, priority = Priority.HIGH)
+public class EuclideanDistanceVoxelization3D<O extends RandomAccessibleInterval<BitType>> extends AbstractUnaryHybridCF<Mesh, O>
+		implements Ops.Geometric.Voxelization {
+
+	@Parameter
+	private OpService ops;
+
+	@Parameter(type = ItemIO.INPUT, required = false)
+	private double wallThickness = 1.0;
```
gselzer: Can you add a `description` to this parameter? I first assumed this was "distance from triangle plane has to be less than `wallThickness`", but this is actually "distance from triangle plane has to be less than `wallThickness/2`"

src/test/java/net/imagej/ops/geom/MeshFeatureTests.java
```diff
@@ -200,7 +194,28 @@ public void verticesCountMesh() {
...
-		// https://github.com/imagej/imagej-ops/issues/422
+	public void voxelization3D(){
+		/**
+		An ideal voxelization algorithm should be able to convert a {@link Mesh} generated from a binary image back into
+		a surface-pixel outline of the original binary image. This surface-pixel image should match the result of
+		processing the original binary image with {@link Boundary} using
+		{@link Boundary.StructuringElement.FOUR_CONNECTED}. When working with real images these can mismatch due to
+		isolated single-pixel objects in the original binary image not being incorporated into the {@link Mesh}
+		*/
+		final Img<BitType> out = new ArrayImgFactory<>(new BitType()).create(getTestImage3D());
```
gselzer, line 205: No changes necessary, just wanted to say that I love this explanation 😍

gselzer (commented):
Loving how this is turning out! Thanks again @andmccall 

Just had a few more ideas but it should be a quick commit and then I think good to merge!

src/main/java/net/imagej/ops/geom/geom3d/DefaultVoxelization3D.java
andmccall: Sorry, I'm not really sure when/how all these weird format changes came in (probably some auto-format from IntelliJ). I'll go and fix all the formatting changes, but we should keep the "@deprecated Use {@link EuclideanDistanceVoxelization3D} instead." yes? I also thought I had the whole plugin marked with a `@Deprecated` .

gselzer: D'oh! Yes, please keep the `@deprecated` :sweat_smile:

andmccall:
@gselzer, Alright, I think I got all the changes made. Had to manually make the whitespace changes for `DefaultVoxelization3D` in GitHub's file editor, as IntelliJ really didn't like the whitespace layout in the original file. Let me know if the explanations for the new algorithm make sense.

